PR #2, opened 2016-03-09
find_slack_bus verbose passage

jdedecca:
Verbose variable not passed before to find_slack_bus()


nworbmot:
Agreed, thanks for the change.

PR #3, opened 2016-03-25
Update opf.py for spillage variables

jdedecca:
Pyomo does not recognize the spill_index dictionary tuples, but does recognize it as a list instead.
Not passing it as a list results in the following error (example of storage unit 31, snapshot 0):

ERROR: Constructing component 'storage_p_spill_index' from data=None failed:
  File "C:\Anaconda3\lib\site-packages\pypsa\opf.py", line 768, in network_lopf
	ValueError: The value=('31', 0) is a tuple for set=storage_p_spill_index, which has dimen=1

### DIFF
--- a/pypsa/opf.py
+++ b/pypsa/opf.py
@@ -162,7 +162,7 @@ def define_storage_variables_constraints(network,snapshots):
     def su_p_spill_bounds(model,su_name,snapshot):
         return spill_bounds[su_name,snapshot]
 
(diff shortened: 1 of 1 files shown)


nworbmot:
Hi, this was a Python 2 versus Python 3 issue (pyomo doesn't seem to deal with iterators gracefully) - thanks for the heads up and sorry for the Easter-related delay in pulling it. Ideally we need a unit test that uses the inflow/spillage so that we would have caught this.

jdedecca:
Glad to help! Indeed more test cases would be useful - I'll be using inflows regularly so may keep an eye out on it in future versions.

PR #12, opened 2017-01-11
Update co2_price exception

jdedecca:
Multiple exception types should be a tuple following https://docs.python.org/3/tutorial/errors.html#handling-exceptions
Otherwise a syntax error is raised in Python 3.5.2, though I wonder whether this is compatible with Python 2.


nworbmot:
Hi Joao, thanks for spotting this! I had already fixed this on another branch "components_csv" and will pull this into the master in the next few days, see https://github.com/FRESNA/PyPSA/commit/fbddc30b25bf01a37b745fae57c0e2b2c74b8a98 .

PR #14, opened 2017-01-26
Copy snapshot weightings

jdedecca:
The network.copy() method does not include snapshot_weightings, providing an incorrect network copy when weights != 1.


nworbmot:
Hi, thanks for pointing out this problem, but the solution is not consistent, since network.copy() should provide a deep copy. This solution will only copy the reference to the snapshot_weightings so that the underlying data are still linked. I'll push a correct change to the master branch now.

jdedecca:
Perfect - it will help me understand more the code then!

Issue #17, opened 2017-04-12
pf() convergence indication


nmartensen:
When running a few hundred load flows in a loop, it would be nice to catch the cases where the load flow did not converge. I couldn't find a way to do this, is there any? Unless this is a bad idea for some other reason, I'd like to request a return value to be added to the pf() function.

nworbmot:
Agreed, this is a good idea. We should catch and return at least a boolean of whether it converged, the number of iterations and the error. This would have to be indexed by snapshot and subnetwork (synchronous area - unlike e.g. PowerFactory, PyPSA calls the load flow separately on each area). It may take a month or two before we get around to implementing this, so if you need it sooner, feel free to implement it and submit a pull request.

nworbmot:
Design-wise, the sub_network_pf and network_pf can return a dictionary (or descriptors.Dict) with keys ["converged","n_iter","error"] and values of DataFrames with index snapshots and columns subnetworks (or just Series indexed by snapshots in case of sub_network_pf). I.e. a bit like network.components_t.

PR #18, opened 2017-05-08
pf.py: implement pf() return code

nmartensen:
So here's a first try to address #17. Please review!


nworbmot:
Thanks Nis! Sorry for the delay, we were a bit busy over the last two weeks.

PR #19, opened 2017-08-02
Infeasibility warnings

bourbl:
Add some warning messages that could lead to optimization infeasibility. Check for NaN values or upper constraints smaller than lower constraints that exist in the max, min attributes.

It could be useful for example when importing wind profiles that have some missing or very small negative values and could be more difficult to track unless the input data is checked.

### DIFF
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -47,7 +47,7 @@ try:
 except ValueError:
     _pd_version = LooseVersion(pd.__version__)
 
-from .descriptors import Dict
+from .descriptors import Dict, get_switchable_as_dense
 
 from .io import (export_to_csv_folder, import_from_csv_folder,
                  import_from_pypower_ppc, import_components_from_dataframe,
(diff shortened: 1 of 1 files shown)


coroa:
Nice idea and looks good to me. Will test during the next few weeks.
Thanks

ksyranid:
Apparently I have wasted some time in finding such annoying errors, so I thought a pre-check could help.

I ran the unit test which didn't show anything (at least), except for the storage but I guess this is because I used gurobi. Actually I changed the tests a bit to allow the solver as argument from command line. Would that be interesting to include too?

nworbmot:
Great, thanks a lot Kostas! I've also had such infeasibility problems (particularly from the NaNs) so this is very helpful.

nworbmot:
Regarding the tests, the problem is that in the github repo we've stored the "official" GLPK solution, so I'm not sure it would help to pass the tests a solver argument, unless we stored a "gurobi" version too. GLPK was chosen because it's freely available. I think a better solution would be to modify the tests so that they have a unique solution, then it doesn't matter what solver is used.

bourbl:
I am glad I could help! I guess you are right about the tests, a unique solution would make things easier.

PR #28, opened 2018-01-10
Faster kirchoff cycle creation (define_passive_branch_flows_with_kirchhoff)

rsmith54:
We have been using PyPSA, and we noticed that for larger numbers of snapshots and branches, the time to create the kirchhoff cycles was quite high.  This PR significantly reduces the time, primarily by moving the time-independent attribute calculations outside of the snapshots loop.

I have also added a new example ```build_model.py```, which can show the speed increase.  It simply creates a (randomized) larger network to induce cycles.  On my computer, the new build_model script runs about 50% faster on this branch than when I run it versus master.

Let me know if you would like any changes.  One easy additional change could be to use ```define_sub_network_cycle_constraints``` for the cycles formulation as well, but I haven't been using it, so I haven't been able to check for sure that that would be reasonable.

### DIFF
--- a/examples/build_model.py
+++ b/examples/build_model.py
@@ -1,0 +1,245 @@
+## Demonstrate PyPSA unit commitment with a one-bus two-generator example
+#
+#
+#To enable unit commitment on a generator, set its attribute committable = True.
+#
+#
+#Available as a Jupyter notebook at http://www.pypsa.org/examples/unit-commitment.ipynb.
+
+import pypsa
+
+from pypsa.opt import network_lopf_build_model as build_model
+
+### Minimum part load demonstration
+#
+#In final hour load goes below part-load limit of coal gen (30%), forcing gas to commit.
+
+for test_i in range(100):
+
+    nu = pypsa.Network()
+
+    nu.set_snapshots(range(4))
+
+    nu.add("Bus","bus")
+
+
+    nu.add("Generator","coal",bus="bus",
+           committable=True,
+           p_min_pu=0.3,
+           marginal_cost=20,
+           p_nom=10000)
+
+    nu.add("Generator","gas",bus="bus",
+           committable=True,
+           marginal_cost=70,
+           p_min_pu=0.1,
+           p_nom=1000)
+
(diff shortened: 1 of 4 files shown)


nworbmot:
So the main speed increase is coming from taking `attribute_value = branches.at[branch_idx,attribute]*(branches.at[branch_idx,"tap_ratio"] if branch_idx[0] == "Transformer" else 1.)*subnetwork.C[cycle_i, col_j]` out of the snapshot loop? I wonder if we should just define an effective `x/r_pu` very early in the code to incorporate the tap ratio.

rsmith54:
Yes, that is the main speed increase. 

I could look at that also; I think the correct place is ```pf.calculate_dependent_values```.  Please let me know if you disagree.

nworbmot:
Great; one can also use it in `opf.define_passive_branch_flows_with_angles()` and `pf.calculate_B_H()` and then everything's consistent. What do you think @coroa? Better to do everything in one place like this, oder?

rsmith54:
It should be replaced in both those functions now.  Please let me know if you need anything else.

coroa:
I am wondering now, whether it's possible to avoid the new columns by changing the definition of {x,r}_pu 
for transformers to already include the tap_ratio? The only place where such a change has non-trivial consequences would be in pf.calculate_Y, i think.

@nworbmot What do you think?

coroa:
P.S.: Sorry for the  broken coveralls spamming (I switched it off on the master, just now), will have a look at why coverage testing doesn't work later this month.

nworbmot:
No, changing the definition of x_pu is a bad idea and could lead to confusion. Ideally we should reorganise the data structure so that the user never sees these "internal variables", but that's a project for another day. @rsmith54 many thanks for this contribution, we can make the remaining changes from here if you have other things to do.

rsmith54:
I made the csr -> csc change.  Thanks for the suggestions and quick turnaround.

coroa:
Great, thanks. @nworbmot would you extend the csv files to include the new fields?

nworbmot:
Yup, I'll get to it these week, perhaps also time for a new release.

PR #33, opened 2018-04-19
Update plot.py

jankaeh:
If radius is 0, ratios are NaN because of division by 0.


nworbmot:
Thanks! And sorry for the delay in merging, we were all changing institution to KIT.

PR #42, opened 2018-09-18
Corrected an error in "Upgrading PyPSA"

miek770:
It says "pip install -U pandas" instead of "... pypsa" in the section explaining how to upgrade PyPSA.


nworbmot:
Thanks!

PR #47, opened 2019-01-28
Custom basemap projection in plot.py

fons-:
The `basemap` argument of `plot` can now be a dictionary with additional arguments to be passed to the `Basemap` constructor. 
For example, you could use:
```python3
d = {'projection': 'tmerc', 'lon_0':10.0,'lat_0':50.0,'resolution':'i', }
network.plot(basemap=d)
```
to produce

Whereas the original:
```python3
d = True
network.plot(basemap=d)
```
produces

------

Note that passing a `boolean` or a `str` to basemap still produces the same result as before.

### DIFF
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -397,8 +397,11 @@ try:
             Series of integer weights for buses, indexed by bus names.
         n_clusters : int
             Final number of clusters desired.
+        buses_i : None|pandas.Index
+            If not None (default), subset of buses to cluster.
         kwargs
-            Any remaining arguments to be passed to KMeans (e.g. n_init, n_jobs)
+            Any remaining arguments to be passed to KMeans (e.g. n_init,
+            n_jobs).
 
         Returns
         -------


coroa:
Thanks for taking the time to send us your additions. Greatly appreciated!

There are actually two changes in this pull request.

1. I like the first one, described above, to extend the informal interface of the basemap argument to convey more settings to basemap. I only have two change requests there: (1) Update the doc-string to make it official and (2) if it is a string (`isinstance(basemap, six.string_types`) use it as resolution for backward compatibility.
2. But, you are also changing how the network projection is used for plotting. Ie you are using a separate basemap invocation your `datamap` to convert the `bus` coordinates to latlong and then plot in latlong in a separate step. Is it possible for you to separate it into a separate PR, because it needs some further discussion?

fons-:
Thanks for your comments!
1. Will do (in a new PR) 👍
2. Hmm... see my new PR about this issue.

coroa:
Are you interested in having a go at `1.` again? Extending `geomap` to accept dictionaries to make it more configurable?

fons-:
I will have a look this week! In the meantime, see #57.

PR #48, opened 2019-01-29
plot: Projection bugfix

fons-:
`bmap( , )` projects lat/lon to SRID coordinates, not the other way around ([doc](https://matplotlib.org/basemap/users/mapcoords.html)). 

When using a dataset that does not use lat/lon coordinates, this should fix a projection issue. (But I am unable to test the code on such a dataset.)

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -231,7 +231,7 @@ def plot(network, margin=0.05, ax=None, basemap=True, bus_colors='b',
                 "The WKT-encoded geometry in the 'geometry' column must be composed of LineStrings"
             segments = np.asarray(list(linestrings.map(np.asarray)))
             if basemap and basemap_present:
-                segments = np.transpose(bmap(*np.transpose(segments, (2, 0, 1))), (1, 2, 0))
+                segments = np.transpose(bmap(*np.transpose(segments, (2, 0, 1)), inverse=True), (1, 2, 0))
 
         l_collection = LineCollection(segments,
                                       linewidths=l_widths,
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -147,7 +147,7 @@ def plot(network, margin=0.05, ax=None, basemap=True, bus_colors='b',
         bmap.drawcountries()
         bmap.drawcoastlines()
 
-        x, y = bmap(x.values, y.values)
+        x, y = bmap(x.values, y.values, inverse=True)
         x = pd.Series(x, network.buses.index)
         y = pd.Series(y, network.buses.index)
 


coroa:
Thanks, for the effort!

> `bmap( , )` projects lat/lon to SRID coordinates, not the other way around ([doc](https://matplotlib.org/basemap/users/mapcoords.html)).

Yes, I understood as much yesterday. And you are right plotting a dataset with a projection must be broken at the moment. Like you I don't have a dataset in another projection at hand.

I don't think the solution is to convert the coordinates to latlong and then plot. Instead, I would rather want to plot in the coordinate system given by network.srid.

Does this mean, we would just have to remove all the calls to basemap and plot the network.x/y coordinates as-is? I'm unsure, but it's likely. Can you think of how to test that? Otherwise I'll experiment myself, but it will have to wait until the weekend.

fons-:
Hi!
I think that you are right; plotting the 'raw' bus coordinates seems to make sense, since `bmap` is plotting coastlines and borders in that same format/projection.
I don't have a different dataset, but I will try to construct one or experiment with matplotlib outside of PyPSA this week.

coroa:
Hi fons were you able to create a small test data set with non-standard projection?

fons-:
Hi! Not yet, something else came up, but I should have some time this week!

coroa:
Great, thanks! Note also https://github.com/PyPSA/PyPSA/pull/51 .

fons-:
Hi,
I have been trying to get an example working, but stumbled onto the problem that a (Basemap) projection is not uniquely defined by its SRID or projection type, it also requires some projection parameters. A mercator example:

```python
from mpl_toolkits.basemap import Basemap

# Plot bounds
x1, x2, y1, y2 = [0,20,50,60]

# Create two basemaps, the second with a bound defined
bmapA = Basemap(projection="merc")
bmapB = Basemap(projection="merc", llcrnrlat=y1, urcrnrlat=y2, llcrnrlon=x1, urcrnrlon=x2)


# Transform the coordinate (lon: 12, lat: 52):
print(bmapA([12],[52]))
print(bmapB([12],[52]))
```

This produces two different results:

```
([21349415.862658788], [66339318.940530896])
([1334338.4914161742], [353453.49673362076])
```

It could be that the `bmapA` projection is the 'standard' mercator projection. (By trail and error, I found that the bounds `[-180,180,-89.99,89.99]` are used by default.)

Do you know any examples of projections that are commonly used in datasets? If not, it might be reasonable to only support lat/lon coordinates in PyPSA. A correct conversion from some projected dataset to lat/lon coordinates seems tricky, and might best be left to the user (in the specialist case that they are using a dataset in a different projection).

coroa:
Hi,
perhaps you are right and i think it is also used only in that way (The other being broken, anyway :)).
Best,
Jonas

coroa:
Fixed by #51 / https://github.com/PyPSA/PyPSA/commit/0b7d01c5144db28cffe97b0519492487d8bdb6fe if one uses cartopy. Perhaps we should encourage switching to `cartopy` if people want to plot in other crs.

PR #50, opened 2019-02-07
components: consistency check now only prints necessary columns

hailiangliu89:
network.consistency_check() now only prints columns containing NaNs rather than all columns with an empty index.

### DIFF
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -891,11 +891,11 @@ class Network(Basic):
 
                 # check for NaN values:
                 if max_pu.isnull().values.any():
-                    for col in max_pu.columns[max_pu.isna().any()].tolist():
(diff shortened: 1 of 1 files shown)


pypsa/components.py
```diff
@@ -891,11 +891,11 @@ def consistency_check(self):
 
                 # check for NaN values:
                 if max_pu.isnull().values.any():
-                    for col in max_pu.dropna(axis=1, how='any').columns:
+                    for col in max_pu.columns[max_pu.isna().any()].tolist():
```
coroa: please stick with `isnull`. `isna` and `isnull` are synonymous, let's keep it consistent.

you don't need the `tolist` instantiation: `for col in max_pu.columns[max_pu.isna().any()]:` works fine.

PR #51, opened 2019-02-13
Update plot.py - Introduce cartopy

FabianHofmann:
Here is the first step in order to update the plot.py a bit. Cartopy enables straightforward plotting with different projections and is much better maintained in future than Basemap. 

Summary of changes (updated on Feb 14, 9:53 am):
- set new argument cartopy_present 
- rename 'basemap' argument to 'geomap' in plot()
- add projection argument in plot() to define the cartopy projection 
- create new function draw_map() outside of plot(), transfer all map drawing code into it
- rename 'bmap' to 'gmap' in draw_map()
- create new function get_projection_from_crs() which takes in account the srid of the network
- Disable transformation of coordinates and segment with basemap object (see [#48](https://github.com/PyPSA/PyPSA/pull/48)), this is probably just a temporary workaround
- create new function projected_area_factor(), which is necessary for scaling circle plots
- update docstring for 'geomap', 'projection'

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -27,6 +27,7 @@ from six import string_types
 import pandas as pd
 import numpy as np
 
+import warnings
 import logging
 logger = logging.getLogger(__name__)
 
@@ -54,7 +55,7 @@ cartopy_present = True
 try:
     import cartopy
     import cartopy.crs as ccrs
-except:
+except ImportError:
     cartopy_present = False
 
 pltly_present = True
@@ -64,10 +65,11 @@ except ImportError:
         pltly_present = False
 
 
-def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
-         line_colors='g', bus_sizes=10, line_widths=2, title="",
-         line_cmap=None, bus_cmap=None, boundaries=None,
-         geometry=False, branch_components=['Line', 'Link'], jitter=None):
+def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
+         bus_colors='b', line_colors='g', bus_sizes=10, line_widths=2,
+         title="", line_cmap=None, bus_cmap=None, boundaries=None,
+         geometry=False, branch_components=['Line', 'Link'], jitter=None,
+         basemap=None):
     """
     Plot the network buses and lines using matplotlib and Basemap.
 
@@ -126,13 +128,22 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
         logger.error("Matplotlib is not present, so plotting won't work.")
         return
 
+    if basemap is not None:
+        warnings.warn("`basemap` is deprecated, use `geomap` instead.",
+                      DeprecationWarning)
+        geomap = basemap
+
+
     if ax is None:
-        if cartopy_present and drawmap:
-            ax = plt.gca(projection=ccrs.PlateCarree())
+        if cartopy_present and geomap:
+            if projection is None:
+               projection = get_projection_from_crs(network.srid)
+
+            ax = plt.gca(projection=projection)
         else:
             ax = plt.gca()
     else:
-        if cartopy_present and drawmap:
+        if cartopy_present and geomap:
             import cartopy.mpl.geoaxes
             assert isinstance(ax, cartopy.mpl.geoaxes.GeoAxesSubplot), (
                     'The passed axis is not a GeoAxesSubplot. You can '
@@ -140,9 +151,8 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
                     'fig, ax = plt.subplots('
                     'subplot_kw={"projection":ccrs.PlateCarree()})')
 
-    if drawmap:
-        geomap, x, y = draw_map(network, jitter, ax, boundaries,
-                                margin, drawmap)
+    if geomap:
+        x, y = draw_map(network, jitter, ax, boundaries, margin, geomap)
     else:
         x, y = network.buses['x'], network.buses['y']
 
@@ -156,7 +166,7 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
             "in the second MultiIndex level of bus_sizes"
 
         bus_sizes = bus_sizes.sort_index(level=0, sort_remaining=False)\
-                        * projected_area_factor(ax)**2
+                        * projected_area_factor(ax, network.srid)**2
 
         patches = []
         for b_i in bus_sizes.index.levels[0]:
@@ -192,7 +202,7 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
             index = network.lines.index
         return pd.Series(ser,
                          index=pd.MultiIndex(levels=(["Line"], index),
-                                             labels=(np.zeros(len(index)),
+                                             codes=(np.zeros(len(index)),
                                                      np.arange(len(index)))))
 
     line_colors = as_branch_series(line_colors)
@@ -228,9 +238,9 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
                 "The WKT-encoded geometry in the 'geometry' column must be "
                 "composed of LineStrings")
             segments = np.asarray(list(linestrings.map(np.asarray)))
-            if drawmap and basemap_present:
-                segments = np.transpose(geomap(
-                        *np.transpose(segments, (2, 0, 1))), (1, 2, 0))
+#            if geomap and basemap_present:
+#                segments = np.transpose(gmap(
+#                        *np.transpose(segments, (2, 0, 1))), (1, 2, 0))
 
         l_collection = LineCollection(segments,
                                       linewidths=l_widths,
@@ -255,12 +265,24 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
 
     if basemap_present:
         ax.axis('off')
-    if cartopy_present and drawmap:
+    if cartopy_present and geomap:
         ax.outline_patch.set_visible(False)
 
     ax.set_title(title)
 
     return (bus_collection,) + tuple(branch_collections)
+
+
+def get_projection_from_crs(crs):
+    if crs == 4326:
+        # if data is in latlon system, return default map with latlon system
+        return ccrs.PlateCarree()
+    try:
+        return ccrs.epsg(crs)
+    except ValueError:
+        logger.warning(f"'{crs}' does not define a projected coordinate system. "
+                    "Falling back to latlong.")
+        return ccrs.PlateCarree()
 
 
 def compute_bbox_with_margins(margin, x, y):
@@ -273,18 +295,18 @@ def compute_bbox_with_margins(margin, x, y):
     return tuple(xy1), tuple(xy2)
 
 
-def projected_area_factor(ax):
+def projected_area_factor(ax, original_crs):
     """
     Helper function to get the area scale of the current projection in
     reference to the default projection.
     """
-    if not 'projection' in ax.__dict__:
+    if not hasattr(ax, 'projection'):
         return 1
-    if ax.projection is ccrs.PlateCarree():
+    if isinstance(ax.projection, ccrs.PlateCarree):
         return 1
     x1, x2, y1, y2 = ax.get_extent()
     pbounds = \
-        ccrs.PlateCarree().transform_points(ax.projection,
+        get_projection_from_crs(original_crs).transform_points(ax.projection,
                     np.array([x1, x2]), np.array([y1, y2]))
 
     return np.sqrt(abs((x2 - x1) * (y2 - y1))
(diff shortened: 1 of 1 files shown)


pypsa/plot.py
```diff
@@ -254,15 +253,95 @@ def as_branch_series(ser):
...
+    #First choice should be cartopy
+    if cartopy_present:
+        resolution = '50m' if isinstance(drawmap, bool) else drawmap
+        assert resolution in ['10m', '50m', '110m'], (
+                "Resolution has to be one of '10m', '50m', '110m'")
+        geomap = ax.projection
+        transformed = pd.DataFrame(
+                    geomap.transform_points(
+                        ccrs.PlateCarree(), x.values, y.values),
+                        columns=['x', 'y', 'z'], index=network.buses.index)
```
coroa: `x` and `y` *are* already in the crs defined by `net.srid`, and don't need to be transformed.

coroa: Refer also to #48

pypsa/plot.py
```diff
@@ -254,15 +253,95 @@ def as_branch_series(ser):
...
+
+    elif basemap_present:
+        resolution = 'l' if isinstance(drawmap, bool) else drawmap
+        geomap = Basemap(resolution=resolution, epsg=network.srid,
+                       llcrnrlat=y1, urcrnrlat=y2, llcrnrlon=x1,
+                       urcrnrlon=x2, ax=ax)
+        geomap.drawcountries(linewidth=0.3, zorder=-1)
+        geomap.drawcoastlines(linewidth=0.4, zorder=-1)
+
+        x, y = geomap(x.values, y.values)
```
coroa: Same

pypsa/plot.py
```diff
@@ -254,15 +253,95 @@ def as_branch_series(ser):
...
+
+    return geomap, x, y
+
+
 #This function was borne out of a breakout group at the October 2017
 #Munich Open Energy Modelling Initiative Workshop to hack together a
 #working example of plotly for networks, see:
-#https://forum.openmod-initiative.org/t/breakout-group-on-visualising-networks-with-plotly/384/7
+#https://forum.openmod-initiative.org/t/
+#breakout-group-on-visualising-networks-with-plotly/384/7
```
coroa: I'd prefer not to break the url, so you can copy it easily. Also use spaces after the hash if possible.

pypsa/plot.py
```diff
@@ -227,11 +224,13 @@ def as_branch_series(ser):
...
             linestrings = c.df.geometry.map(loads)
-            assert all(isinstance(ls, LineString) for ls in linestrings), \
-                "The WKT-encoded geometry in the 'geometry' column must be composed of LineStrings"
+            assert all(isinstance(ls, LineString) for ls in linestrings), (
+                "The WKT-encoded geometry in the 'geometry' column must be "
+                "composed of LineStrings")
             segments = np.asarray(list(linestrings.map(np.asarray)))
-            if basemap and basemap_present:
-                segments = np.transpose(bmap(*np.transpose(segments, (2, 0, 1))), (1, 2, 0))
+            if drawmap and basemap_present:
```
coroa: If we show all coastline and so on in the coordinates of `net.srid`, we do not need to reproject the segments. Otherwise you would have to do the equivalent thing with cartopy.

pypsa/plot.py
```diff
@@ -117,49 +127,36 @@ def plot(network, margin=0.05, ax=None, basemap=True, bus_colors='b',
...
-        x = x + np.random.uniform(low=-jitter, high=jitter, size=len(x))
-        y = y + np.random.uniform(low=-jitter, high=jitter, size=len(y))
-
-    if basemap and basemap_present:
-        resolution = 'l' if isinstance(basemap, bool) else basemap
-
-        if boundaries is None:
-            (x1, y1), (x2, y2) = compute_bbox_with_margins(margin, x, y)
+        if cartopy_present and drawmap:
+            ax = plt.gca(projection=ccrs.PlateCarree())
```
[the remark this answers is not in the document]
FabianHofmann: This seems to be tricky, as the srid code (as default '4326') does not define a 'projected coordinate system' but a "geodetic coordinate system", which however does not correspond to a projection map. 
See the very last definition of the following [link](https://scitools.org.uk/cartopy/docs/latest/crs/index.html)
I am trying to figure out a way...

pypsa/plot.py
```diff
@@ -254,15 +253,95 @@ def as_branch_series(ser):
...
+    Helper function to get the area scale of the current projection in
+    reference to the default projection.
+    """
+    if not 'projection' in ax.__dict__:
+        return 1
+    if ax.projection is ccrs.PlateCarree():
+        return 1
+    x1, x2, y1, y2 = ax.get_extent()
+    pbounds = \
+        ccrs.PlateCarree().transform_points(ax.projection,
```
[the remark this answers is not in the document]
FabianHofmann: Actually this is fine, as the points should be transformed to the original projection (which was assumed to be PlateCarree, however should be given in srid). Ideally this would achieved by 

```python 
    pbounds =  ccrs.epsg(net.srid).transform_points(ax.projection, 
                                np.array([x1, x2]), np.array([y1, y2]))
```

which is  however not possible as the default net.srid is not a valid epsg code (see comment above)

pypsa/plot.py
FabianHofmann: The idea behind this is the flexibility of changing to other projections, like this the following code 

```python 
fig, ax = plt.subplots(subplot_kw={'projection': ccrs.EqualEarth()})
net.plot(ax=ax)
```
is possible (again assuming the original projection is PlateCarree)

pypsa/plot.py
FabianHofmann: This was just taken over from the old code, except renaming bmap into geomap

pypsa/plot.py
FabianHofmann: this was also just taken from the normal master branch code

pypsa/plot.py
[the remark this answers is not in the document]
coroa: I don't understand the difference, to be honest. Alternatively, we might add a crs argument to the plot function, which by default is None.

Then
```
if crs is None:
   crs = net.srid

try:
   # instantiate projection from crs
except NotPossibleSinceNotCorrect:
 logger.warning(f"'{crs}' does not define a projected coordinate system. Falling back to latlong.")
  projection = ccrs.PlateCarree()
```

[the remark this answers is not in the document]
coroa: in that case we also have to differentiate between the `srid` coordinates and the `crs` coordinates. Perhaps `crs` is not a good name.

pypsa/plot.py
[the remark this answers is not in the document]
coroa: Correct. Only see above.

pypsa/plot.py
coroa: Fine.

pypsa/plot.py
coroa: Only that it turned out to be broken, see #48.

pypsa/plot.py
coroa: but broken :)

pypsa/plot.py
coroa: note that the `projection` argument here is pretty much what i had in mind with `crs` above.

coroa: I assume that's a feature of cartopy? Seems to indicate that `projection` is a better term for it.

pypsa/plot.py
```diff
@@ -255,14 +265,26 @@ def as_branch_series(ser):
...
+        return ccrs.PlateCarree()
+    try:
+        return ccrs.epsg(crs)
+    except ValueError:
+        logger.warning(f"'{crs}' does not define a projected coordinate system. "
+                    "Falling back to latlong.")
+        return ccrs.PlateCarree()
+
+
 def compute_bbox_with_margins(margin, x, y):
```
FabianHofmann: This function should be used to assoiciate the network.srid with an proper projection. As the srid=4326, which is not an valid projection for cartopy, refers to the lat/lon system, the function takes the default lat/lon projection PlateCarree in this case, without a waring, as otherwise, the warning would appear many times for the default srid.

pypsa/plot.py
```diff
@@ -71,8 +79,12 @@ def plot(network, margin=0.05, ax=None, basemap=True, bus_colors='b',
...
     ax : matplotlib ax, defaults to plt.gca()
         Axis to which to plot the network
-    basemap : bool, default True
-        Switch to use Basemap
+    geomap: bool/str, default True
+        Switch to use Basemap or Cartopy (depends on what is installed).
+        If string is passed, it will be used as a resolution argument.
+        For Basemap users 'c' (crude), 'l' (low), 'i' (intermediate),
+        'h' (high), 'f' (full) are valid resolutions options.
+        For Cartopy users '10m', '50m', '110m' are valid resolutions options.
```
coroa: Docstring for `projection` is missing

FabianHofmann: is now included :)

pypsa/plot.py
```diff
@@ -255,14 +269,26 @@ def as_branch_series(ser):
 
     if basemap_present:
         ax.axis('off')
-    if cartopy_present and drawmap:
+    if cartopy_present and geomap:
         ax.outline_patch.set_visible(False)
```
coroa: Should probably be

```
    if cartopy_present and geomap:
        ax.outline_patch.set_visible(False)
    elif basemap_present:
        ax.axis('off')
```
we prefer cartopy and fall back to basemap

coroa: or even

```
if geomap:
    if cartopy_present:
        ax.outline_patch.set_visible(False)
    ax.axis('off')
```

pypsa/plot.py
```diff
@@ -308,40 +334,40 @@ def draw_map(network=None, jitter=None, ax=None, boundaries=None,
 
     #First choice should be cartopy
     if cartopy_present:
-        resolution = '50m' if isinstance(drawmap, bool) else drawmap
+        resolution = '50m' if isinstance(geomap, bool) else geomap
         assert resolution in ['10m', '50m', '110m'], (
                 "Resolution has to be one of '10m', '50m', '110m'")
-        geomap = ax.projection
+        gmap = ax.projection
+        orig_projection = get_projection_from_crs(network.srid)
```
coroa: i think calling it `data_projection` would be clearer

pypsa/plot.py
```diff
@@ -308,40 +334,40 @@ def draw_map(network=None, jitter=None, ax=None, boundaries=None,
...
                        urcrnrlon=x2, ax=ax)
-        geomap.drawcountries(linewidth=0.3, zorder=-1)
-        geomap.drawcoastlines(linewidth=0.4, zorder=-1)
+        gmap.drawcountries(linewidth=0.3, zorder=-1)
+        gmap.drawcoastlines(linewidth=0.4, zorder=-1)
 
-        x, y = geomap(x.values, y.values)
+        x, y = gmap(x.values, y.values)
         x = pd.Series(x, network.buses.index)
         y = pd.Series(y, network.buses.index)
```
coroa: Well, the `basemap` bit is as broken, as it has been before, but I can understand that you don't want to fix it. You can leave it if you want.

pypsa/plot.py
```diff
@@ -126,23 +132,31 @@ def plot(network, margin=0.05, ax=None, drawmap=True, bus_colors='b',
...
                     'The passed axis is not a GeoAxesSubplot. You can '
                     'create one with: \nimport cartopy.crs as ccrs \n'
                     'fig, ax = plt.subplots('
                     'subplot_kw={"projection":ccrs.PlateCarree()})')
 
-    if drawmap:
-        geomap, x, y = draw_map(network, jitter, ax, boundaries,
-                                margin, drawmap)
+    if geomap:
+        x, y = draw_map(network, jitter, ax, boundaries, margin, geomap)
```
coroa: I don't think I like that `jitter` does not work if one does not have any map. I think it would be better to pull it back into the main function and hand over `x` and `y`.

pypsa/plot.py
```diff
@@ -254,11 +267,103 @@ def as_branch_series(ser):
...
     return (bus_collection,) + tuple(branch_collections)
 
 
+def get_projection_from_crs(crs):
+    if crs == 4326:
+        # if data is in latlon system, return default map with latlon system
+        return ccrs.PlateCarree()
+    try:
+        return ccrs.epsg(crs)
+    except ValueError:
```
fons-: `ccrs.epsg` requires an internet exception ([doc](https://scitools.org.uk/cartopy/docs/latest/crs/index.html#cartopy.crs.epsg)). If the request fails, the package 'pyepgs', used by cartopy, [will raise a `requests.RequestException`](https://github.com/rhattersley/pyepsg/blob/master/pyepsg.py#L287). 
```suggestion
    except requests.RequestException:
        logger.warning("A connection to http://epsg.io/ is required for a projected coordinate reference system. "
                    "Falling back to latlong.")
    except ValueError:
```
(The `requests` package should be imported.)

pypsa/plot.py
fons-: I found that setting the bounds of `Basemap` (`llcrnrlat`, etc.) dramatically changes the projection. Since the bounds are not given by the user, but somewhat arbitrary, this basemap should not be used for conversion. (I think raising an error when basemap is used in combination with a non-standard projection might be safest.)

coroa:
Merged manually in 6eb7173bc96858ece50d47d7eb0fa76f41f678b6.

PR #52, opened 2019-02-13
networkclustering: Fix aggregategenerators to handle unweighted generators

fneum:
If generators are aggregated by weight in `aggregategenerators` and happen to have a `nan` weight, the weighting does not add up to `1.`


coroa:
Thanks for the pr. I fixed the problem in https://github.com/PyPSA/PyPSA/commit/3c49b2e861d902b1eac37f8063f8cb2c79f2c068.

coroa:
:)

PR #54, opened 2019-03-15
import non-standard trafos and lines

stianchris:
Import of the CIGRE network did not work (see included example). Now with non-standard components it is imported correctly. I did not yet compare the network results in pandapower and pypsa with each other, but the parameters all seem correct. Should also be backwards-compatible.


stianchris:
Did not want to close this request, sorry!

nworbmot:
Thanks for this contribution! Sorry for the delay in incorporating it.

PR #57, opened 2019-04-09
plot.py: Using Basemap or `geomap=False` works again

fons-:
- Basemap was broken: buses and lines were not drawn on the axis.

- `geomap=False` was broken: `transform` was never defined. Using this option now gives:

I will add custom projection options this week!

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -56,6 +56,7 @@ try:
     import cartopy
     import cartopy.crs as ccrs
     import cartopy.mpl.geoaxes
+    import requests
 except ImportError:
     cartopy_present = False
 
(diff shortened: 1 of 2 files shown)


pypsa/plot.py
```diff
@@ -56,6 +56,7 @@
     import cartopy
     import cartopy.crs as ccrs
     import cartopy.mpl.geoaxes
+    import requests
```
fons-: `import requests` to fix [this line](https://github.com/PyPSA/PyPSA/blob/0b7d01c5144db28cffe97b0519492487d8bdb6fe/pypsa/plot.py#L287).

requirements.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
   - numpy
   - pyomo
   - scipy
-  - pandas>=0.19.0
+  - pandas>=0.24.0
```
fons-: See [my earlier comment](https://github.com/PyPSA/PyPSA/commit/0b7d01c5144db28cffe97b0519492487d8bdb6fe#r33106820).

coroa:
Thanks for that catch!

Sorry for making you go down through that rabbit hole, since I didn't test w/ basemap.

PR #58, opened 2019-04-10
plot.py: Custom Basemap projection; better switching between basemap and cartopy

fons-:
Cartopy is now used by default, unless the new parameter `basemap_parameters` is a dict. In that case, a Basemap is created, with the dict as additional constructor parameters. These can be used to specify a Basemap projection.

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -60,18 +60,24 @@ try:
 except ImportError:
     cartopy_present = False
 
+# Use cartopy by default, fall back on basemap
+use_basemap = False
+use_cartopy = cartopy_present
+if not use_cartopy:
+    use_basemap = basemap_present
+
 pltly_present = True
 try:
-        import plotly.offline as pltly
+    import plotly.offline as pltly
 except ImportError:
-        pltly_present = False
+    pltly_present = False
 
 
 def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
          bus_colors='b', line_colors='g', bus_sizes=10, line_widths=2,
          title="", line_cmap=None, bus_cmap=None, boundaries=None,
          geometry=False, branch_components=['Line', 'Link'], jitter=None,
-         basemap=None):
+         basemap=None, basemap_parameters=None):
     """
     Plot the network buses and lines using matplotlib and Basemap.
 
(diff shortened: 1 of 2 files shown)


fons-:
Looks like I should have created a new fork before starting this second PR, sorry!

pypsa/plot.py
```diff
@@ -139,7 +151,30 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
+        if not (cartopy_present or basemap_present):
+            # Not suggesting Basemap since it is being deprecated
+            logger.warning("Cartopy needs to be installed to use `geomap=True`.")
+            geomap = False
+
+        # Use cartopy by default, fall back on basemap
+        use_basemap = False
+        use_cartopy = cartopy_present
+        if not use_cartopy:
+            use_basemap = basemap_present
```
coroa: What's the argument behind duplicating the logic of the above 4 lines also on the top-level? Why not remove the top-level switch in lines 63-67, as well as `global` parameter?

coroa: Or instead just use the global one.

fons-: They are global because `use_basemap` and `use_cartopy` are modified from within `plot`, and used inside `draw_map`.
I included the logic in the top level to allow a user to call `draw_map` on its own.

In hindsight, it might be better to change the `if use_cartopy:` condition in `draw_map` to `if basemap_parameters=None:`. In fact, [using `{}` as default argument was a bad idea](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). I will update my PR!

fons-: Fixed in my [latest commit](https://github.com/PyPSA/PyPSA/pull/58/commits/1088fc6247b52951983ab6ee94e0fd7d2a291af7).

fons-:
@coroa Have you had a chance to look at my latest commit?

coroa:
I did. you addressed everything correctly as far as I can tell from the code. Didn't have the chance to test that cartopy and basemap both work correctly. I am putting it higher onto my list :). Sorry for the high latency.

coroa:
Merged by hand with slight modifications.

coroa:
Thanks for the work and patience!

PR #60, opened 2019-04-25
plot.py: add switch to paint land and sea areas

fneum:
Adds a boolean switch `color_geomap` to paint land areas in `whitesmoke` and oceans in `lightblue` for both cartopy and basemap.

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -71,7 +71,7 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
          bus_colors='b', line_colors='g', bus_sizes=10, line_widths=2,
          title="", line_cmap=None, bus_cmap=None, boundaries=None,
          geometry=False, branch_components=['Line', 'Link'], jitter=None,
-         basemap=None):
+         basemap=None, color_geomap=False):
     """
     Plot the network buses and lines using matplotlib and Basemap.
 
@@ -117,6 +117,8 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
     jitter : None|float
         Amount of random noise to add to bus positions to distinguish
         overlapping buses
+    color_geomap : bool, default False
+        Switch to paint land and sea areas in light colors
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


coroa:
In principle an easy and nice addition, couple of comments:
1. `draw_map` was split in the course of #58, so your change has to be split
   accordingly
2. Would be good to make the colors configurable, ie `color_geomap` could be a 
   dictionary with the keys `ocean` and `land` for example.
3. There's a typo in the `basemap`-implementation: `drawlsmask` is the correct
   spelling. (refer to  https://matplotlib.org/basemap/api/basemap_api.html#mpl_toolkits.basemap.Basemap.drawlsmask)
   
   
Thus, please test it works in cartopy and basemap.

fneum:
I have implemented the suggested changes and tested for both `basemap` and `cartopy`.

pypsa/plot.py
```diff
@@ -360,55 +368,62 @@ def projected_area_factor(ax, original_crs):
...
+
+    resolution = 'l' if isinstance(geomap, bool) else geomap
+    gmap = Basemap(resolution=resolution,
+                    llcrnrlat=y1, urcrnrlat=y2, llcrnrlon=x1,
+                    urcrnrlon=x2, ax=ax, **basemap_parameters)
+    gmap.drawcountries(linewidth=0.3, zorder=-1)
+    gmap.drawcoastlines(linewidth=0.4, zorder=-1)
+
+    if color_geomap is None:
+        color_geomap = {'ocean': 'w', 'land': 'w'}
```
coroa: Intentional to use the same color for `land` and `ocean`? You did use `whitesmoke` and `lightblue` before.

fneum: Yes. If one doesn’t specify colours, the map remains in the “minimal colour setting” as it was before.

coroa:
Hi,

hmm, sorry that i'm so slow with my reviews. Ok, then i misunderstood your intention the last time and I saw what and why FH changed the code afterwards. It would probably be better not to call `drawlsmasks` unless `color_geomap` defines any colors or evaluates to `True`.

Imagine you want to plot your network atop an imagemesh showing the capacity factor, then the new patches are now obscuring that imagemesh or you want to put the plots on a webpage with a gray background. If you don't want to plot something, then don't.

Issue #36, opened 2018-06-19
Multi-Horizon Investment


nworbmot:
As of Version 13.1, PyPSA only considers investment in one horizon, e.g. for a single year. We'd like to implement investment over multiple horizons simultaneously, e.g. for 2020, 2025, 2030, 2035, etc., to optimise investments dynamically over several years e.g. under tightening CO2 constraints. The objective function then becomes the discounted net present value (NPV) of energy system costs. Other open models such as [OSeMOSYS](http://www.osemosys.org/), [Genesys](http://www.genesys.rwth-aachen.de/), [DIETER](http://www.diw.de/dieter) (I think) and [SWITCH](https://github.com/switch-model/switch) do this already; we can copy their implementations and avoid reinventing the wheel.

On the data side:

i) We need a second index for snapshots that gives the investment year, e.g.
snapshots = [(2020, "windy day"), (2020, "sunny day"), (2030, "windy day"), (2030, "sunny day")]
The representative weather for each investment year may change with the year (e.g. due to climate change); the number of representative snapshots for each investment year may also change.

ii) We need different CO2/renewable constraints for different investment years.

iii) For assets, we need to take account of build years and lifetimes for generation, storage and transmission assets, e.g. we need additional fields

name, build year, life time, cost
wind-in-2020, 2020, 20, 1200
wind-in-2025, 2025, 20, 1000
wind-in-2030, 2030, 25, 900

Then in the code we need to activate assets in the constraints depending on whether they are around in the investment year or not.
We also need to take account of the discount rate as a pre-factor to get the NPV for each investment year.
Perhaps this should be linked to issue #37 of separating the snapshot_weighting from the duration of each snapshot, which was discussed on the PyPSA mailing list before.
BEWARE 1): The final investment horizon has to be treated carefully, to avoid end-of-horizon effects whereby assets are only built for the final horizon and have no value for the rest of their lifetimes. Either model 20 years further (with lower intra-year resolution) or have a "salvage value" for investments made in the final investment horizon (I think OSeMOSYS had this).
BEWARE 2): p_nom_max may become linked between different generators as potentials are used up.

arjunane:
Hi @nworbmot ,
just some short thoughts concerning you suggestions:
i) The number and the lengths of representative snapshots may vary as well.
[...]

PR #77, opened 2019-07-04
[DNMY] Introduce flag stating whether transmission line is operational

fneum:
This PR introduces a new attribute `operational` to the component `Line`.

| attribute | type | unit | default | description | status | 
| --- |  ---  | --- |  --- |  --- |  --- | 
| operational | boolean | n/a | True | "Switch indicating whether a line is in operation or not, also useful to label candidate lines" | Input (optional) | 

If `operational=False` the line will be ignored for all power flow / optimal power flow formulations. That means it is not included in the graph, adjacency matrix, incidence matrix, etc.; however these lines are kept in the corresponding time-invariant and time-dependent dataframes `network.lines` and `network.lines_t.---`

One obvious (more general) use-case is to observe changes as a particular transmission line becomes unavailable. In principle this is already possible by removing/adding lines to the network.

However, the main motivation is to prepare for introducing an `lopf` formulation for mixed-integer transmission expansion. The `operational` attribute will identify integer investment candidate lines (`operational=False`) for which an alternative KVL formulation is required (Big-M reformulation).

The attribute name `operational` was chosen over a more specific attribute name e.g. `candidate`, since the former is slightly more general and encompasses the candidate line interpretation.

Attribute name `active` was considered, but avoided due to ambiguities related to active/reactive power and passive branches.

**Points of Discussion:**
- `s_nom_opt=0` if `operational=False`?
- `p0=0` or `p0=NaN` if `operational=False`? (also other time-dependent attributes)
- For now, the `operational` attribute is not time-dependent since this is sufficient for the candidate line interpretation. This could be added if useful.

**Remaining TODO:**
- [x] write unit test for `lopf`
- [x]  write unit test for `lpf`


nworbmot:
I had in mind that we introduce such a boolean for **all** assets and integrate this with the multi-horizon investment (https://github.com/PyPSA/PyPSA/issues/36), so that e.g. "Wind turbine 2020" is "active" (or "operational") only in years 2020-2044 (for a 25-year lifespan).

fneum:
superseded by #127

PR #81, opened 2019-07-17
Skip pandapower unit test for Python 2.7 if pandapower >= 2.1.0 (closes #79)

### LINKED ISSUE #79: Unit testing for Python 2.7 with pandapower 2.1.0 fails
Tests for Python 2.7 fail in 7038b76c751e211135a4a7f0aa74a3d13b69b4ee as it was the first CI run with the recent `pandapower v2.1.0` (https://travis-ci.org/PyPSA/PyPSA/jobs/557399790).

While `pandapower v2.0.0` already officially dropped Python 2.7 support, the unit tests of `pypsa` work with up to `pandapower v2.0.1` (https://github.com/e2nIEE/pandapower/releases and discussion in https://github.com/e2nIEE/pandapower/issues/224).

1. A proposal would be to fix the version in `environment_dev.yaml` to `pandapower==2.0.1` until the Python 2.7 countdown has run down (https://pythonclock.org/) on January 1, 2020. We make a final release tested for Python 2.7 then, and allow again to pull most recent `pandapower`.
2. An alternative is to create Python version specific `environment_dev.yaml` files; e.g. `environment_dev_python2.yaml` and `environment_dev_python3.yaml`.
3. Another option is to drop Python 2.7 support right now.

Other opinions/options?

fneum:
Deviation from proposal in #79 by @coroa seems necessary as already the import of `pandapower` fails. Instead check whether import is successful and skip the correpsonding test against `pandapower` if it is not.

### DIFF
--- a/test/test_pf_against_pandapower.py
+++ b/test/test_pf_against_pandapower.py
@@ -3,15 +3,22 @@ from __future__ import absolute_import
 
 import pypsa
 
-import pandapower as pp
-
 import pandas as pd
 
 import numpy as np
 
+pandapower_compatible = True
+try:
+    import pandapower as pp
+except ImportError:
+    pandapower_compatible = False
 
+import six
 
+import pytest
 
+@pytest.mark.skipif(six.PY2 and not pandapower_compatible,
+                    reason="Pandapower 2.0.0 dropped support for Python 2.")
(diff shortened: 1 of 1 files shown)


coroa:
LGTM

Issue #80, opened 2019-07-15
Adding argument to l_objective


dcradu:
We are currently looking into using some `pypsa` features in our planning tools and we came across an issue that could be interesting to tackle. More specifically, the `l_objective` function in the `opt.py` module takes as argument the `model` and the `objective`, but the `sense` is set as `minimize` within the function. Would it be possible to add an extra argument to the function so that the `sense` can be input by the user (in case we want to `maximize` certain objectives)?

fneum:
Possible yes, however you could just as well multiply your expression with `-1`. Feel free to prepare a pull request. Important would be that the default behaviour remains to minimize the objective function.

```python
def l_objective(model, objective=None, sense=1):
    ...
```

where setting `sense=-1` will maximise. You may also use `maximize` and `minimize` from `pyomo`, but bare in mind you have to import them: `from pyomo.environ import maximize, minimize`.

PR #82, opened 2019-07-18
Adding sense (minimize/maximize) to l_objective.

dcradu:
According to issue #80. Opens the possibility to explicitly state the sense of the objective function (minimize, maximize), while the default setting is, as requested, minimizing the objective.

### DIFF
--- a/pypsa/opt.py
+++ b/pypsa/opt.py
@@ -38,7 +38,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-from pyomo.environ import Constraint, Objective, Var, ComponentUID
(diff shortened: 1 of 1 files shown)


coroa:
I think it would be good if 3rd party PRs always received two positive reviews before merging, so if @fneum or @nworbmot agree, they are free to merge or ping back their approval.

dcradu:
@coroa if the follow-up is the result of my "Close/Re-open" action, I apologize for that. There was no intention of pushing for any action. I am just new to the PR procedure and it was a mistake. Otherwise, I fully agree with having at least two positive reviewers on-board before any further step. Thanks!

coroa:
@dcradu No harm done!
@fneum And thanks for merging!

PR #86, opened 2019-08-12
networkclustering: keep snapshot weighting if clustering with time

fneum:
If clustering `with_time`, snapshot weightings of the original network should be carried over to the clustered network. Related to pypsa/pypsa-eur#64.


coroa:
Good catch! Thanks.

PR #90, opened 2019-08-16
Add mapbox to network.iplot(...)

fneum:
I played a little bit with adding Mapbox to the `network.iplot(...)` function.

To explore in a jupyter notebook run

```python
import pypsa
import plotly.offline as pltly
pltly.init_notebook_mode(connected=True)

network = pypsa.Network("networks/elec_s_100.nc")

token = "XXXXXX"
network.iplot(network, mapbox=True,
                      mapbox_token=token,
                      mapbox_parameters={"style": 'dark'}, size=(1000,1000))
```

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -63,6 +63,7 @@ except ImportError:
 pltly_present = True
 try:
     import plotly.offline as pltly
+    import plotly.graph_objects as go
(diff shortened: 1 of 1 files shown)


FabianHofmann:
Looks good! I went though the code and tested it.

PR #91, opened 2019-08-16
Add flow plotting functionality

FabianHofmann:
Some changes in plot.py to plot flows. 

Status was tested using the following cases

```python


import pypsa, os
import cartopy.crs as ccrs
import pandas as pd

n = pypsa.Network(os.path.dirname(pypsa.__file__) + '/../examples/ac-dc-meshed/ac-dc-data/')
n.lopf()

#standard projetion
gen = n.generators.assign(g = n.generators_t.p.mean()).groupby(['bus', 'carrier']).g.sum()

n.plot(bus_sizes=gen/1e4,
       bus_colors={'gas':'indianred', 'wind':'midnightblue'},
       margin=.5,
       flow='mean',
       line_widths=0.1,
       line_colors={'Line':'coral', 'Link':'skyblue'})
plt.show()


#different projection
n.plot(margin=.5,
       flow=pd.Series(10, n.branches().index),
       line_widths=1,
       line_colors={'Line':'coral', 'Link':'skyblue'},
       projection=ccrs.EqualEarth(),
       color_geomap=True)
plt.show()

#cmap for flow not implemented but works for static plot

n.plot(margin=.5,
       line_widths=10,
       line_colors=pd.Series(range(1, 1+len(n.branches())), n.branches().index)/10,
       projection=ccrs.EqualEarth(),
       color_geomap=True)
plt.show()

#  plot flow of snapshot

n.plot(margin=.5,
       flow=n.snapshots[0],
       line_widths={'Line':0.1, 'Link':0.1},
       projection=ccrs.EqualEarth(),
       color_geomap=True)
plt.show()
```

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -265,11 +265,11 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
                          zip(network.iterate_components(ser.keys()), ser.values())},
                     names=['component', 'name'])
         elif isinstance(ser, pd.Series) and isinstance(ser.index, pd.MultiIndex):
-                return ser.rename_axis(index=['component', 'name'])
+            return ser.rename_axis(index=['component', 'name'])
         else:
             ser =  pd.Series(ser, network.lines.index)
-        return pd.concat([ser], axis=0, keys=['Line'],
-                         names=['component', 'name']).fillna(0)
+            return pd.concat([ser], axis=0, keys=['Line'],
+                             names=['component', 'name']).fillna(0)
 
     line_colors = as_branch_series(line_colors)
     line_widths = as_branch_series(line_widths)
@@ -484,7 +484,6 @@ def directed_flow(n, flow, x=None, y=None, ax=None, geomap=True,
     """
     Helper function to generate arrows from flow data.
     """
-    #Probably plt.quiver is much faster here!
     # this funtion is used for diplaying arrows representing the network flow
     from matplotlib.patches import FancyArrow
     if ax is None:
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -231,8 +231,9 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
             "bus_colors must be a dictionary defining a color for each element " \
             "in the second MultiIndex level of bus_sizes"
 
-        bus_sizes = bus_sizes.sort_index(level=0, sort_remaining=False)\
-                        * projected_area_factor(ax, network.srid)**2
+        bus_sizes = bus_sizes.sort_index(level=0, sort_remaining=False)
+        if geomap:
+            bus_sizes *= projected_area_factor(ax, network.srid)**2
 
         patches = []
         for b_i in bus_sizes.index.levels[0]:
@@ -259,11 +260,12 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
     def as_branch_series(ser):
         # ensure that this function always return a multiindexed series
         if isinstance(ser, dict) and set(ser).issubset(branch_components):
-            return network.branches().loc[branch_components] \
-                          .join(pd.Series(ser, name='value') \
-                          .rename_axis('component')).value
+            return pd.concat(
+                    {c.name: pd.Series(s, index=c.df.index) for c, s in
+                         zip(network.iterate_components(ser.keys()), ser.values())},
+                    names=['component', 'name'])
         elif isinstance(ser, pd.Series) and isinstance(ser.index, pd.MultiIndex):
-                return ser.rename_axis(['component', 'name'])
+                return ser.rename_axis(index=['component', 'name'])
         else:
             ser =  pd.Series(ser, network.lines.index)
         return pd.concat([ser], axis=0, keys=['Line'],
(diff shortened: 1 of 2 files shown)


pypsa/plot.py
```diff
@@ -247,27 +256,40 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
         if isinstance(ser, dict) and set(ser).issubset(branch_components):
-            return pd.Series(ser)
-        elif isinstance(ser, pd.Series):
-            if isinstance(ser.index, pd.MultiIndex):
-                return ser
-            index = ser.index
-            ser = ser.values
+            return network.branches().loc[branch_components] \
+                          .join(pd.Series(ser, name='value') \
+                          .rename_axis('component')).value
```
coroa: Can you get this also working without creating a copy of all static branch attributes just to compose the relatively branches multiindex. That is so wasteful.

Maybe something along the lines of
```python
pd.concat({c.name: pd.Series(s, index=c.df.index)
           for c, s in zip(network.iterate_components(ser.keys()), ser.values())})
```

pypsa/plot.py
```diff
@@ -247,27 +256,40 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
+        elif isinstance(ser, pd.Series) and isinstance(ser.index, pd.MultiIndex):
+                return ser.rename_axis(['component', 'name'])
```
coroa: Use `rename(index= ...)` or `rename(columns=...)`, instead of `rename_axis`. It's clearer and you can change both at the same time.

pypsa/plot.py
```diff
@@ -247,27 +256,40 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
     if not isinstance(line_cmap, dict):
         line_cmap = {'Line': line_cmap}
 
     branch_collections = []
 
+    if flow is not None:
+        flow = _flow_ds_from_arg(flow, network, branch_components)\
+                        .pipe(as_branch_series)\
+                        .div(len(network.branches()) + 100)\
+                        .mul(line_widths, fill_value=1)
```
coroa: Please prefer extra `(` to `)` to line continuations with `\`. Not all people are used to them, often the indentation is all over the place and its easy to miss one.

```python
         flow = (_flow_ds_from_arg(flow, network, branch_components)
                 .pipe(as_branch_series)
                 .div(len(network.branches()) + 100)
                 .mul(line_widths, fill_value=1))
```

```
network.branches()
```
is to be considered expensive.

```python
sum(len(t.df) for t in network.iterate_components(branch_components))
```

pypsa/plot.py
FabianHofmann: yes great, I searched for something like this

pypsa/plot.py
FabianHofmann: `rename(index=...)` is only changing values in the index, not the name of the index axis, which is intended here. So I have to stick to this one. but at least now it is `.rename_axis(index=['component', 'name'])`

pypsa/plot.py
FabianHofmann: yes, agreed! Thanks for revising

FabianHofmann:
@coroa I think this is ready to merge

pypsa/plot.py
```diff
@@ -258,11 +260,12 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
-            return network.branches().loc[branch_components] \
-                          .join(pd.Series(ser, name='value') \
-                          .rename_axis('component')).value
+            return pd.concat(
+                    {c.name: pd.Series(s, index=c.df.index) for c, s in
+                         zip(network.iterate_components(ser.keys()), ser.values())},
+                    names=['component', 'name'])
         elif isinstance(ser, pd.Series) and isinstance(ser.index, pd.MultiIndex):
-                return ser.rename_axis(['component', 'name'])
+                return ser.rename_axis(index=['component', 'name'])
```
coroa: I'm pretty sure it is
`ser.rename(index=...)`

pypsa/plot.py
```diff
@@ -473,18 +478,20 @@ def _flow_ds_from_arg(flow, n, branch_components):
...
 
 
-def directed_flow(n, flow, x=None, y=None, ax=None,
+def directed_flow(n, flow, x=None, y=None, ax=None, geomap=True,
                   branch_colors='darkgreen', branch_comps=['Line', 'Link'],
                   cmap=None):
     """
     Helper function to generate arrows from flow data.
     """
+    #Probably plt.quiver is much faster here!
```
coroa: `quiver` is for a 2-d grid of vectors, not for arbitrary graphs

pypsa/plot.py
```diff
@@ -132,11 +142,11 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
     bus_collection, branch_collection1, ... : tuple of Collections
         Collections for buses and branches.
     """
-    defaults_for_branches = {
+    defaults_for_branches = pd.Series({
         'Link': dict(color="cyan", width=2),
         'Line': dict(color="b", width=2),
         'Transformer': dict(color='green', width=2)
-    }
+    }).rename_axis('component')
```
coroa:
```suggestion
    }).rename(index='component')
```

pypsa/plot.py
```diff
@@ -247,27 +258,42 @@ def plot(network, margin=0.05, ax=None, geomap=True, projection=None,
...
+        elif isinstance(ser, pd.Series) and isinstance(ser.index, pd.MultiIndex):
+                return ser.rename_axis(index=['component', 'name'])
         else:
-            index = network.lines.index
-        return pd.Series(ser,
-                         index=pd.MultiIndex(levels=(["Line"], index),
-                                             codes=(np.zeros(len(index)),
-                                                    np.arange(len(index)))))
+            ser =  pd.Series(ser, network.lines.index)
+        return pd.concat([ser], axis=0, keys=['Line'],
```
coroa: Please indent to `ser` level, since it's only important for this branch (clarity).

pypsa/plot.py
FabianHofmann: Here is the example which does not work:
```python 
network.branches().rename(index=['component', 'name'])
```
(if this is what you mean) fails as pd.DataFrame.rename wants to have a mapper or a function. 
On the other hand when passing a mapper, as 
```python 
ser = network.branches()
ser.rename(index=dict(zip(ser.index.names, ['component', 'name'])))
```
does not change the index axis names, and would only affect **values** within the index.

coroa: Seems, I was mistaken (I did mix it up with reindex).

coroa: Sorry

PR #97, opened 2019-09-23
Update link length in networkclustering

### LINKED ISSUE #94: Link length not considered in networkclustering
When clustering the network, links will be aggregated but their length not recalculated. I am not sure whether this is intended. 
This leads to a distortion of the line extension volume limit in f.e. pypsa-eur which favours 'short' links which are in reality long.
 
@coroa can you help me here?

If it is not intended I would add some line in networkclustering to cover this.

FabianHofmann:


coroa:
@FabianHofmann If your fine with the implementation change, just go ahead and merge!

PR #107, opened 2019-11-22
distributed slack implementation

### LINKED ISSUE #10: Multiple Slack Generators / Buses
Is there a trick to introduce more than one bus with slack generator behaviour?
When I modelled in SINCAL I sometimes used more than one slack bus for e.g. modelling the connection to other countries in a very abstract and ideal way.
I am asking in context of performing the lpf() and pf(). 
When introducing a storage or store/link component the p_set would define the dispatch right up front by the input data. So is there a way to trick the components towards a slack bahaviour?

fneum:
fixes #10 

Adds an option `distribute_slack` to 
```python
def network_pf(network, snapshots=None, skip_pre=False, 
x_tol=1e-6, use_seed=False, distribute_slack=False):`
```
to distribute the total slack power to all generators proportional to their dispatch at the respective snapshot.

Docstrings already extended.

@anileinchen, could you please also test this?

Remaining TODOs:
- [x] add a test
- [x] update documentation

### DIFF
--- a/test/test_pf_distributed_slack.py
+++ b/test/test_pf_distributed_slack.py
@@ -5,7 +5,8 @@ import pypsa
 def normed(s): return s/s.sum()
 
 def test_pf_distributed_slack():
-    csv_folder_name = os.path.join(os.path.dirname(__file__), "../examples/scigrid-de/scigrid-with-load-gen-trafos/")
+    csv_folder_name = os.path.join(os.path.dirname(__file__), "..",
+                      "examples", "scigrid-de", "scigrid-with-load-gen-trafos")
 
     network = pypsa.Network(csv_folder_name)
 
--- a/test/test_pf_distributed_slack.py
+++ b/test/test_pf_distributed_slack.py
@@ -17,10 +17,7 @@ def test_pf_distributed_slack():
     network.lopf(network.snapshots, solver_name='cbc', formulation='kirchhoff')
 
     #For the PF, set the P to the optimised P
-    network.generators_t.p_set = network.generators_t.p_set.reindex(columns=network.generators.index)
     network.generators_t.p_set = network.generators_t.p
-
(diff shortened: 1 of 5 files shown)


FabianHofmann:
looks good (but not tested), I would replace all `.real`'s and `.imag`'s by the functions. There is still deprecation warnings, good opportunity to tackle this

FabianHofmann:
you mind if I push a commit which removes unnecessary imports? (btw it seems that the function `normed` is not defined)

fneum:
@FabianHofmann 
- go ahead an remove unnecessary imports; I forgot to add normed; is fixed.
- Where exactly do you still get deprecation warnings for `.real`? I only got them for pd.Series / pd.DataFrame

fneum:
Found the remaining deprecation warnings for `.real/.imag`.

FabianHofmann:
> Found the remaining deprecation warnings for `.real/.imag`.

yes, looks good. They appear when running `pytest`

FabianHofmann:
mmh, okay solving the collections deprecation warning isn't compatible with python2.7, another reason to drop the support. I'll revert it for the moment

test/test_pf_distributed_slack.py
```diff
@@ -0,0 +1,42 @@
...
+
+    #There are some infeasibilities without line extensions
+    network.lines.s_max_pu = 0.7
+    network.lines.loc[["316","527","602"],"s_nom"] = 1715
+    network.storage_units.state_of_charge_initial = 0.
+
+    network.lopf(network.snapshots, solver_name='cbc', formulation='kirchhoff')
+
+    #For the PF, set the P to the optimised P
+    network.generators_t.p_set = network.generators_t.p_set.reindex(columns=network.generators.index)
```
FabianHofmann: is there a reason for first reindexing and then overwriting? Seem's to be a leftover of a the scipy-example-code somehow

PR #108, opened 2019-11-27
pf: distributed slack with custom weights and weights by capacity

### LINKED ISSUE #39: Prevent network.pf() failing on single-bus sub-networks
If there are sub-networks with only a single bus, i.e. a single bus with no connected lines or transformers, the non-linear power flow network.pf() will fail because the admittance matrix Y for this sub-network is zero. We need to add an escape for this case, whereby the slack generator balances everything at single-node sub-networks.

fneum:
This PR extends the distributed slack implementation from #107 and additionally closes #39, a long-time bug in PyPSA that prevents pf() from solving for single-bus subnetworks.

Additional options for `network.pf()` are described in the docstring
```
distribute_slack : bool, default False
        If ``True``, distribute the slack power across generators proportional to generator dispatch by default
        or according to the distribution scheme provided in ``slack_weights``.
        If ``False`` only the slack generator takes up the slack.
    slack_weights : dict|str, default 'p_set'
        Distribution scheme describing how to determine the fraction of the total slack power
        (of each sub network individually) a bus of the subnetwork takes up.
        Default is to distribute proportional to generator dispatch ('p_set').
        Another option is to distribute proportional to (optimised) nominal capacity ('p_nom' or 'p_nom_opt'). 
        Custom weights can be specified via a dictionary that has a key for each
        subnetwork index (``network.sub_networks.index``) and a
        pandas.Series/dict with buses or generators of the
        corresponding subnetwork as index/keys.
        When specifying custom weights with buses as index/keys the slack power of a bus is distributed
        among its generators in proportion to their nominal capacity (``p_nom``) if given, otherwise evenly.
```

- [x] distribute slack by capacity
- [x] give option to provide custom **bus** weights to distribute slack for each subnetwork individually
- [x] give option to provide custom **generator** weights to distribute slack for each subnetwork individually
- [x] handle special case with only a single-bus subnetwork without passive branch components (closes #39)
- [x] update documentation
- [x] update tests

**Squash if merging!**

### DIFF
--- a/doc/power_flow.rst
+++ b/doc/power_flow.rst
@@ -77,7 +77,7 @@ to their dispatch (``distribute_slack=True``), instead of being
 allocated fully to the slack bus, the active power balance is altered to
 
 .. math::
-   \textrm{Re}\left[ V_i \left(\sum_j Y_{ij} V_j\right)^* \right] - P_i - P_{slack}\gamma_i & = 0 \hspace{.7cm}\forall\hspace{.1cm} i \in PV \cup PQ \cup slack
+   \textrm{Re}\left[ V_i \left(\sum_j Y_{ij} V_j\right)^* \right] - P_i - P_{slack}\gamma_i = 0 \hspace{.7cm}\forall\hspace{.1cm} i \in PV \cup PQ \cup slack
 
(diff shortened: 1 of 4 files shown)


nworbmot:
Looks great, thanks @fneum!

PR #115, opened 2019-12-31
Harmonise nomopyomo, pyomo LOPF status and termination_condition

nworbmot:
Make nomopyomo LOPF return values `status` and `termination_condition` consistent with pyomo LOPF.

`status`, `termination_condition` are either `ok, optimal` for optimal solutions or `warning, infeasible` for infeasible ones. (NB: Unlike gurobi and cbc, glpk returns `ok, infeasible` for infeasible solutions. For nomopyomo I've made this `warning, infeasible` for all solvers, for consistency.)

For gurobi there are two additional options:

If gurobi is using the barrier method and has a sub-optimal solution, nomopyomo will still read the solution and return `warning, suboptimal`. In this case pyomo would return `warning, other`. This was caught by the pyomo LOPF, see here

https://github.com/PyPSA/PyPSA/blob/54d16deb18a2a07f68efb6f83c2c632933df966e/pypsa/opf.py#L1589

so that PyPSA-Eur could still use the solutions, see here

https://github.com/PyPSA/pypsa-eur/blob/8bd72b47f0bb9bfcde6931b33bc1b1274eb26566/scripts/solve_network.py#L294

since the sub-optimal solutions were often very close to optimal (e.g. gap between primal and dual objective of 1e-3). Some model.energy solutions were also returning sub-optimal, and need to be caught.

If gurobi is using the barrier method, sometimes it cannot distinguish between an infeasible and an unbounded problem, so returns `warning, infeasible or unbounded`. I *think* pyomo just return `warning, infeasible` in this case.

## Checklist

### DIFF
--- a/pypsa/linopt.py
+++ b/pypsa/linopt.py
@@ -563,14 +563,14 @@ def run_and_read_cbc(n, problem_fn, solution_fn, solver_logfile,
 
     if data.startswith("Optimal - objective value"):
         status = "ok"
-        termination_condition = status
+        termination_condition = "optimal"
         objective = float(data[len("Optimal - objective value "):])
     elif "Infeasible" in data:
+        status = "warning"
         termination_condition = "infeasible"
-        status = 'infeasible'
     else:
+        status = 'warning'
         termination_condition = "other"
-        status = 'other'
 
     if termination_condition != "optimal":
         return status, termination_condition, None, None, None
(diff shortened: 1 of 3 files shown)


nworbmot:
Oops, testing is failing because for unit commitment GLPK returns "integer optimal" not "optimal" and I've made the scripts stricter about the output message. I'll fix this today.

pypsa/linopt.py
```diff
@@ -685,12 +691,22 @@ def run_and_read_gurobi(n, problem_fn, solution_fn, solver_logfile,
...
-        return status, termination_condition, None, None, None
-    else:
+    termination_condition = statusmap[m.status]
+
+    if termination_condition == "optimal":
         status = 'ok'
+    elif termination_condition == 'suboptimal':
+        status = 'warning'
+    elif termination_condition == "infeasible":
+        status = 'warning'
```
FabianHofmann: I think these lines can be deleted or?

nworbmot: Yes, but I'd rather keep them to show what possible termination conditions are possible, especially if we want to change the behaviour later.

nworbmot:
Yes! I've checked it on several model.energy cases where the solution was sub-optimal.

PR #116, opened 2020-01-02
linopf: Correct marginal prices by snapshot weightings

nworbmot:
Marginal prices should be corrected by the snapshot weightings.
This is then consistent with the marginal prices in the Pyomo LOPF:
https://github.com/PyPSA/PyPSA/blob/b76798d2347101dd8e8239251ae7ebe6a1643d23/pypsa/opf.py#L1359
and the original nomopyomo:
https://github.com/PyPSA/nomopyomo/blob/1e9ef2ad58ae864d2baf826c58af7f5931229543/nomopyomo.py#L567
It means the prices have a clear EUR/MWh definition regardless of the snapshot weighting. It also makes them stay roughly the same when snapshots are aggregated.
Dual variables are anyway only defined up to a constant (since you can multiply the whole constraint by a constant).
It raises the question of whether to correct the other dual values saved by PyPSA (e.g. line mu's) by the snapshot weightings. I'm not sure about this. First we should make the behaviour consistent between pyomo and nomopyomo versions of PyPSA, which is what this PR does.

## Checklist

### DIFF
--- a/pypsa/linopf.py
+++ b/pypsa/linopf.py
@@ -794,7 +794,7 @@ def network_lopf(n, snapshots=None, solver_name="cbc",
          solver_logfile=None, extra_functionality=None,
          extra_postprocessing=None, formulation="kirchhoff",
          keep_references=False, keep_files=False,
-         keep_shadowprices=['Bus', 'Line', 'GlobalConstraint'],
+         keep_shadowprices=['Bus', 'Line', 'Transformer', 'Link', 'GlobalConstraint'],
          solver_options=None, warmstart=False, store_basis=False,
          solver_dir=None):
     """
--- a/pypsa/linopf.py
+++ b/pypsa/linopf.py
@@ -746,6 +746,9 @@ def assign_solution(n, sns, variables_sol, constraints_dual,
     for c, attr in sp:
         map_dual(c, attr)
 
+    #correct prices for snapshot weightings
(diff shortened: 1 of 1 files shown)


pypsa/linopf.py
```diff
@@ -746,6 +746,9 @@ def map_dual(c, attr):
     for c, attr in sp:
         map_dual(c, attr)
 
+    #correct prices for snapshot weightings
+    n.buses_t.marginal_price.loc[sns] = n.buses_t.marginal_price.loc[sns].divide(n.snapshot_weightings.loc[sns],axis=0)
```
FabianHofmann:
```suggestion
    n.buses_t.marginal_price.loc[sns] = n.buses_t.marginal_price.loc[sns].divide(n.snapshot_weightings.loc[sns], axis=0)
```

nworbmot:
@FabianHofmann One inconsistency between nomopyomo at pyomo formulations is that pyomo also reads out the `mu_lower` and `mu_upper` shadow prices for non-extendable / fixed lines, transformers and links, whereas nomopyomo doesn't. The pyomo formulation does this by generating constraints for the fixed branches, see e.g.
https://github.com/PyPSA/PyPSA/blob/98f53ffa7fb02fae7172df3b61c86eab676fce11/pypsa/opf.py#L972
nomopyomo doesn't see these shadow prices, because the upper and lower limits are defined with the variable, rather than as extra constraints. Do you think this would be easy to change to be consistent with the pyomo formulation?

nworbmot:
I'll open a separate issue on this.

PR #134, opened 2020-02-03
fix groupby for pandas v1.0.0

FabianHofmann:
- minor: fix deprecation for logger

I really don't know how exactly the `groupby` changed in the new pandas version, but applying a function on each row in a group has changed. Meanwhile this solves it, however has a slightly worse performance for large networks


nworbmot:
Looks good to me, thanks for fixing this!

fneum:
Does this require pandas 1.0.0?

FabianHofmann:
> Does this require pandas 1.0.0?

No, this works also with old pandas versions

PR #137, opened 2020-02-06
WIP: Switches

### LINKED ISSUE #89: Adding Switches
Add switches as a new component. Each line would have two switches with status as open or closed. An example csv/dataframe would look as follows:

| name | line | connected_bus | disconnected_bus | status | b0b1 |
| --- | --- | --- | --- | --- | --- | 
| s1 | 63 | 13915_LV | 13915_LV_s1 | open | bus0 |
| s2 | 64 | 13911_LV | 13911_LV_s2 | open | bus1 |

- Two open switches place a line out of service by removing from the net.lines df and into an 'os_lines' dataframe of the network object for accounting purposes. This way the line can be appended to the net.lines df from the os_lines df if brought back into service. 
- One open switch creates an auxiliary bus (similar to pandapower) referred to as "disconnected_bus" in the above mentioned dataframe. 

I currently have working code for the above logic for my own "grid" class which I inherited from PyPSA's Network class. Would be great to hear your thoughts.

jankaeh:
Closes #89.

## Checklist


jankaeh:
Hi, I realized that adding electrical components needs to take switches into account also... They should all be opned before adding components and "buses_only_logical" need to be readded if they are connected to electrical elements. I added a minimal example for switches, where I comment on this. So there definetly is some work to do.

jankaeh:
Hi, I just pushed the current state with some issues left. Maybe you wait with reviewing until I've adressed those points.

fneum:
No worries! While you are at it, it would be really nice if you could create a unit test in the `test` directory for instance using your example. You can also draw inspiration from the other test cases.

jankaeh:
Phew, I think I covered import/export and adding/removing components with the last commits... Definetly needs testing. And the unit test is still missing. Hope to have time for it in the next few days.

nworbmot:
Sorry, only just looking at this. Thanks for contributing, but I have a few questions. Is the idea to implement fictitious buses to implement the switches in the admittance matrix (e.g. if only one line switch is open)? Will there also be switches for loads and generators? Are you only planning to implement this for the power flow, or also for the OPF? Thanks for your time to answer the questions!

jankaeh:
Sorry, it's been a while... so the idea is to not differ between switches by attached components ("line switch", "generator switch", etc.). Instead switches just connect buses. I tried to document this (https://github.com/jankaeh/PyPSA/blob/switches/doc/components.rst#switch):
"Switches are logical elements between a bus0 to a bus1. They have a status that indicates if they are open (0) or closed (1). Logical elements do not have impedances and thus are no part of the admittance matrix. So they are no part of powerflow calculations. But logical elements are part of the network and determine its topology.
The absence of logical elements in the topology of the network requires a logic for adding buses to and removing buses from the network dependent on status of logical elements and at the same time changing the buses of elements that are attached to logical elements."

"The functions network.open_switches(switches) and network.close_switches(switches) can be used to change the status of the given switches and redetermine the network topology accordingly. The status of switches is an output of these functions, but you may also change the status manually and then call network.switching() to apply all status of logical elements and determine the resulting topology. Changing the status without calling network.switching() will have no effect."

So, regarding your question "is the idea to implement fictitious buses to implement the switches in the admittance matrix", maybe this is clearer:
Every switch has a bus0 and a bus1. In addition a "bus_connected" is determinded, which is active if the switch is closed. The determination of this bus happens dependent on the logical topology. If Switches are connected to each other they share the same bus_connected.
So if a switch is open, its two buses bus0 and bus1 are part of the admittance matrix. Any other element can be connected to bus0/bus1. If this switch is closed, bus0 and bus1 will be removed from the admittance matrix and the bus_connected will be added. All elements from bus0 and bus1 will be attached to bus_connected.

Regarding your question "Are you only planning to implement this for the power flow, or also for the OPF?":
I tried to minimize conflicts and did not touch the calculations - So the pf does not know anything about switches being part of the network. Switching needs to be done before a loadflow to change the topology. So OPF will work as usual also, but switches won't be part of the problem...

nworbmot:
Thanks @jankaeh for these explanations and for the web call just now.
As discussed, I think a design with a clearer separation of input and output data would be better. We also need to make sure that if the calculation breaks off with an error in the middle, that the input data is not corrupted.
I would suggest:
- The "Switch" is a new component with attributes: `name`, `bus0`, `bus1` and `closed` (boolean True or False). This is clearly defined input data.
- When running a PF or OPF, **if** there are switches, then PyPSA makes a copy of the network, alters the bus structure and then runs the regular calculation on the copy. The copy should only be made for the relevant snapshots to save memory, i.e. `network.copy(snapshots=snapshots)` and stored somewhere as an underscore attribute, e.g. `network._fictitious_copy`.
- In the copy of the network the `bus_connected` can be added and other components (loads, generators, etc.) reconnected before the calculation is carried out.
- After the calculation the results can be read back into the original network and e.g. the `network.buses_t.p` and `network.switches_t.p0/1` calculated.
- The code should also handle switches in series/parallel/a Y or Delta gracefully.
We should apply the same design pattern of creating a copy of the network for other components that require fictitious components (e.g. three-winding transformer is modelled by 3x two-winding transformers connected at a fictitious node).

fneum:
Closing stale PR.

PR #158, opened 2020-03-03
Transformer model

TobiasDess:
Signed-off-by: Tobias Dess tobias.dess@elena-international.com
Changes proposed in this Pull Request

    Implentation of off-nominal voltage level of transformer in pf.py
    Test Skript against NEPLAN for several different parameter combinations

Question: Are under author all contributer listet or just the main maintainer? I just wrote myself there, since I don#t know better. If this is not OK, feel free to change it...
Checklist

    [I am happy for feedback on that] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in doc.
    [X ] Unit tests for new features were added (if applicable).
    [not needed ] Newly introduced dependencies are added to environment.yaml, environment_docs.yaml and setup.py (if applicable).
    [See below ] A note for the release notes doc/release_notes.rst of the upcoming release is included.

    Transformer Model can now respect the off-nominal voltage level (e.g. 22 kV/0.45 kV Transformer between the 20kV and 0.4 kV voltage level). Implementation is unit-tested against NEPLAN 10.8.2.4. in normal and tap-switched state.


nworbmot:
Thanks Tobias! It's a bit difficult for me to judge the code since it's a while since I looked at this. It seems the transformer implementations differ from software to software. I'd like it to be compatible with pandapower. Any changes we make, we should be absolutely sure about because we don't want to change the behaviour later.
With that in mind, could you:
- Check how the implementation and results compare to pandapower, e.g. https://github.com/e2nIEE/pandapower/blob/develop/pandapower/build_branch.py
- Does it make sense to alter the tap_ratio with the "off_nominal_ratio" when it is defined in line 704 rather than at tau_hv later? Or does this depend on which side the tap is defined?
- Correct all spellings "off_nomial" -> "off_nominal"
- Are the suffixes in 707-708 correctly defined?
- The try-except is very opaque, perhaps an "if "off_nomial_tap_ratio" in branches" instead?

TobiasDess:
Hello Tom,

thank you for the reply. We needed it compatible with Neplan in a current Project and had good pre-defined test cases for that. I can totally understand your point to keep it in line with pandapower. I will further look into it, when I finished the current deadlines (in around 3-4 weeks). 
To answer your points:

- Check how the implementation and results compare to pandapower, e.g. https://github.com/e2nIEE/pandapower/blob/develop/pandapower/build_branch.py

See above

-     Does it make sense to alter the tap_ratio with the "off_nominal_ratio" when it is defined in line 704 rather than at tau_hv later? Or does this depend on which side the tap is defined?

I defined the  "off_nominal_ratio" from the high voltage side. Since you use the tab ratio in tau is determined by the "tab_side", it would be incorrect. Maybe I will find a more clean way to implement it.

-     Correct all spellings "off_nomial" -> "off_nominal"

Thanks

-     Are the suffixes in 707-708 correctly defined?

I will check it

-     The try-except is very opaque, perhaps an "if "off_nomial_tap_ratio" in branches" instead?

Valid point

So thank you for your work in PyPSA until now. 
Tobias

TobiasDess:
Hi Tom (or other maintainer),

sorry for taken so long.

I rebased my fork on the current master to ensure maximum compatibility and avoid submitting a new Pull request. Hope this procedure was ok? Or would it be better to start a new PR?

I took on your feedback and changed the tests in order to demonstrate synchronicity with Pandapower. I only included the tests in the final PR which cover the most areas. all small tests leading up to this were not interesting for the project? I thought that will be better to save pytest time and CI time.

To answer your questions before:

- Check how the implementation and results compare to pandapower, e.g. https://github.com/e2nIEE/pandapower/blob/develop/pandapower/build_branch.py

-> Checked the implementation, but the software architecture was to different to copy it somehow. Espechially the design decision to seperate between tau_hv and tau_lv was kind of incompatible. To synchronice the implementation on a code level would be way more invasive. I included a test case for both sides of the tab, to demonstrate the same end results

- Does it make sense to alter the tap_ratio with the "off_nominal_ratio" when it is defined in line 704 rather than at tau_hv later? Or does this depend on which side the tap is defined?

--> Th definition of the off_nominal_tab_ratio is independent of the tab side and independent o a tab present. Since the calculation of the tab ratio doest take the side into account in apply_transformer_types(network) and the value has also effects on electrical parameter, this way seemed to me the least invasive way to include this. But if you have a better idea, feel free :)

- Correct all spellings "off_nomial" -> "off_nominal"
--> Done

- Are the suffixes in 707-708 correctly defined?
--> Yes but one of the was unnecessary, I removed it

- The try-except is very opaque, perhaps an "if "off_nomial_tap_ratio" in branches" instead?
--> Done

Hope I could help with this PR

Best regards
Tobias

nworbmot:
Thanks Tobias! Sorry for the slow response, all the potential reviewers of the code are busy with deadlines at the moment, but we hope to review and merge this very soon!

TobiasDess:
Hi Tom,

that's understandable. Happy to hear from you :)

Best Regard
Tobias

PR #162, opened 2020-03-15
Small code cleanup

fneum:
- remove unused imports
- remove unused variables
- improve type checks
- remove reimports
- misc code style issues

### DIFF
--- a/test/test_pf_against_pandapower.py
+++ b/test/test_pf_against_pandapower.py
@@ -1,9 +1,4 @@
-from __future__ import absolute_import
-
-
 import pypsa
-
-import pandas as pd
 
 import numpy as np
 
--- a/test/test_pf_against_pandapower.py
+++ b/test/test_pf_against_pandapower.py
@@ -12,66 +12,89 @@ import six
 
 import pytest
 
-@pytest.mark.skipif(six.PY2 and not pandapower_compatible,
-                    reason="Pandapower 2.0.0 dropped support for Python 2.")
+
+@pytest.mark.skipif(
+    six.PY2 and not pandapower_compatible,
+    reason="Pandapower 2.0.0 dropped support for Python 2.",
+)
 def test_pandapower_case():
 
-    #more complicated examples like
-    #net = pandapower.networks.example_simple()
-    #can be used once the import of e.g. switches is perfected
+    # more complicated examples like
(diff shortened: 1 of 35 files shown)


fneum:
Where are the `from __future__ import ...` still necessary and where not?

FabianHofmann:
The `__future__` imports are all referring to compatibility between python 2 and python 3. As far as I see there remain, the following future imports:

* `absolute_import` 
* `print` 
* `division`

All of those are in the [MandatoryRelease](https://docs.python.org/3/library/__future__.html) of python 3, meaning they are not needed for version > 3.

FabianHofmann:
I'll clean them out and push it here

FabianHofmann:
Also not needed anymore:

* `from six import string_types` &rarr; use standard `str` instead
* `from six.moves import range` &rarr; use standard `range`
* `from six.moves import map` &rarr; use standard `map`

Often unnecessary 
* `from six import iteritems` and `iterkeys` &rarr; use the .items() function for dicts and series

But perhaps we keep this for another PR, since it requires modifying the code a bit

test/test_pf_against_pandapower.py
```diff
@@ -17,66 +12,89 @@
...
+    )
+    np.testing.assert_array_almost_equal(
+        n.transformers_t.p1.loc["now"], net.res_trafo.p_lv_mw
+    )
+    np.testing.assert_array_almost_equal(
+        n.transformers_t.q0.loc["now"], net.res_trafo.q_hv_mvar
+    )
+    np.testing.assert_array_almost_equal(
+        n.transformers_t.q1.loc["now"], net.res_trafo.q_lv_mvar
+    )
```
FabianHofmann: I guess this is only line breaks or?

fneum: Yes, I have mapplied black coding style to this script as I was unhappy with the formatting of lines 41ff.

fneum:
Thanks for going through the examples! It's a bit odd that you need the `cufflinks` although it is never really called anywhere.

FabianHofmann:
yes, it's because some interactive are called via `pd.DataFrame.iplot()`. This does not come with the standard pandas package, cufflinks has to be imported to enable it

fneum:
Will merge anyway since failing PR CI is only due to Mac hickups.

PR #170, opened 2020-03-23
linopf: assign p0/p1 only for attr s

fneum:
I had defined a new time-dependent variable on the lines:
```py
pypsa.linopt.define_variables(network, 0, 1000, "Line", attr='loss')
```
The function `pypsa.linopt.assign_solution` then wrongly assigned the variable values of "loss" to "p0" and "p1" as it only checks whether the component belongs to passive branch components and not the attribute itself.
With this PR I'd still be able to retrieve the variable values after optimisation with 
```py
network.sols["Line"].pnl["loss"]
```
and makes the function a bit more flexible.

[Sorry, that this comes right after the release]


FabianHofmann:
Ah damn! Sorry I missed that when writing it down! I think it makes sense to have the same check in line 697 right?

fneum:
No worries, I've extended the fix to links.

PR #184, opened 2020-06-17
Support for the FICO Xpress solver

merraksh:

Provide support fo the FICO Xpress commercial solver, in a way similar to the support given by the other two commercial solvers (Cplex, Gurobi) via a problem file to be read and solved.

## Checklist

- Documented new function `run_and_read_xpress`;
- A note for the release notes `doc/release_notes.rst` of the upcoming release is included.

## Acknowledgments
Acknowledgment to the team at Transnet for their significant contribution to this Xpress-PyPSA binding.


doc/release_notes.rst
```diff
@@ -7,6 +7,8 @@ Upcoming Release
 
 .. warning:: The features listed below are not released yet, but will be part of the next release! To use the features already you have to install the ``master`` branch, e.g. ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
 
+* Added support for the FICO Xpress commercial solver.
```
FabianHofmann, line 10:
```suggestion
* Added support for the FICO Xpress commercial solver for `pyomo=False`
```

pypsa/linopt.py
```diff
@@ -803,3 +803,74 @@ def run_and_read_gurobi(n, problem_fn, solution_fn, solver_logfile,
...
+    if termination_condition not in ["optimal"]:
+        return status, termination_condition, None, None, None
+
+    var = [str(v) for v in m.getVariable()]
+    variables_sol = pd.Series(m.getSolution(var), index=var).pipe(set_int_index)
+
+    try:
+        dual = [str(d) for d in m.getConstraint()]
+        constraints_dual = pd.Series(m.getDual(dual), index=dual).pipe(set_int_index)
+    except:
```
FabianHofmann, line 867: Could you add the error type which is raise here?

merraksh: Good point, just pushed. It should catch xpress.SolverError when getDual() fails (due to e.g. unsuccessful solve). All other types of errors (InterfaceError, ModelError) are not caught, but are much more unlikely: m.getConstraint() always returns a list, possibly empty.

FabianHofmann:
Great! thanks for the contribution!

merraksh:
Wonderful, thanks for the quick response.

PR #185, opened 2020-06-22
Inverter controller strategies

### LINKED ISSUE #122: Addition of Control strategies as attributes of Load and Storage_unit components
I think it is a nice idea to implement some Reactive Power Control strategies in pypsa. 

 Because as renewable energy utilization is increasing, number of inverters being integrated to the grid is also increasing through integration of renewable energy power plants, electric vehicle's, chp plants and maybe many more devices containing inverters in the future, which will cause voltage and current violations in the grid. Ex: if more renewable is comming to the grid voltage at the connection point will go higher than the normal range or in case of electric vehicles vice versa and same case for current.

To avoid such issues some inverter control strategies are considered to support the grid by providing reactive power or consuming reactive power in inverters in order to keep the voltage/current  in the allowed range. 

Currently, application of these strategies are done manually in pypsa by making a for loop and iterating for each snapshot and applying control strategy in each snapshot manually which takes time and also slows down the simulation. If they can be implemented in pypsa network main code, it helps alot in reducing the simulations time and code sizes.

idea:

My idea is to apply the control strategies as attributes in load and storage_units components. Because  renewable energy integration can come under load with negative p and also chp plants and electric vehicles come under load or storage_units. So if control strategies are implemented in load and storage unit components, they will consider the chosen control strategy in each snapshot be default and there is no need for further "for loops and iterations" after the power flow.

I have coded the control strategies and they are working fine. If you also think that it is a good idea, i am interested to implement them in pypsa main code. Any hint or examples regarding on how to do it in the main pypsa code would be appreciated.

yama519:
Closes #122.

## Integration of the following inverter control strategies :

1: Fix power factor ,
2: Power factor as a function of active power PF(P) 
3: Reactive power as a function of voltage Q(U)

as new attributes of  controllable_one_port_components (Load, Generator, Store, StorageUnit).

Currently, the above mentioned strategies are implemented, later on  power curtailment P(U) and On Load Tap  changer (OLTC) transformer will be also added.

Please let me know if a brief presentation, documentation or any other supportive document for calrification of the controllers  implementation structure in PyPSA is needed.

## Checklist


pypsa/components.py
```diff
@@ -784,7 +784,7 @@ def copy(self, with_time=True, snapshots=None, ignore_standard_types=False):
 
         Examples
         --------
-        >>> network_copy = network.copy()
+        >>> network_copy = fy()
```
fneum, line 787: Was this by accident?

binder/environment.yml
```diff
@@ -18,5 +18,6 @@ dependencies:
   - glpk
   - gurobi::gurobi
   - pip:
+  - deepcopy
```
fneum, line 21: Isn't this included in Python in any case?
```py
from copy import deepcopy
```

doc/release_notes.rst
```diff
@@ -22,6 +22,11 @@ Upcoming Release
 
 * When solving ``n.lopf(pyomo=False, store_basis=True, solver_name="cplex")`` an error raised by trying to store a non-existing basis is caught.
 
+* Integration of inverter control strategies and on load tap changer transformer for non linear power flow:
+   Inverter control strategies such as: fix power factor, reactive power as a function of voltage Q(U),
+   Active power as a function of voltage P(U) and  power factor (PF) as a function of active power PF(P) can be
+   can be implemented on Load, Generator, Store and StorageUnit components as new attributes for these components
+   and also application of On-Load tap changer transformer as attribute of transformer would be interesting to be implemented.
```
fneum, line 29:
```suggestion
* Integration of inverter control strategies and on-load tap changer (OLTC) transformer for nonlinear power flow:
   Inverter control strategies such as: fixed power factor, reactive power as a function of voltage Q(V),
   active power as a function of voltage P(V) and power factor (PF) as a function of active power PF(P) can be
   can be implemented on Load, Generator, Store and StorageUnit components by specifying new attributes.
```

Removed last bit on OLTCs because release note should only include new features.

PF abbreviation for power factors leads to some ambiguity with the power flow function `n.pf()`. Maybe we can use another or no abbreviation?

I think, we have used V for voltage so far, not U.

environment.yaml
```diff
@@ -16,3 +16,4 @@ dependencies:
 #  - coincbc
   - glpk
   - gurobi::gurobi
+  - deepcopy
```
fneum, line 19: see above, don't think it's necessary

environment_docs.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
   - pyomo
   - cartopy>=0.16
   - coincbc
+  - deepcopy
```
fneum, line 18: see above, don't think it's necessary

examples/minimal_example_pf.py
```diff
@@ -32,7 +32,7 @@
 network.add("Generator","My gen",
             bus="My bus 0",
             p_set=100,
-            control="PQ")
+            control="PQ", q_set=0)
```
fneum, line 35: Did the default value change? I think `q_set` automatically defaults to 0.

```suggestion
            control="PQ")
```

pypsa/component_attrs/generators.csv
```diff
@@ -2,37 +2,49 @@ attribute,type,unit,default,description,status
...
-ramp_limit_start_up,float,per unit,1.,"Maximum active power increase at start up, per unit of the nominal power.   Only read if committable is True.",Input (optional)
-ramp_limit_shut_down,float,per unit,1.,"Maximum active power decrease at shut down, per unit of the nominal power.   Only read if committable is True.",Input (optional)
-p,series,MW,0.,active power at bus (positive if net generation),Output
-q,series,MVar,0.,reactive power (positive if net generation),Output
-p_nom_opt,float,MW,0.,Optimised nominal power.,Output
+ramp_limit_start_up,float,per unit,1,"Maximum active power increase at start up, per unit of the nominal power.   Only read if committable is True.",Input (optional)
+ramp_limit_shut_down,float,per unit,1,"Maximum active power decrease at shut down, per unit of the nominal power.   Only read if committable is True.",Input (optional)
+type_of_control_strategy,string,n/a,n/a,"Inverter control type, supported types are: q_v (v1, v2, v3, v4, s_nom, power_factor, damper), fixed_cosphi (power_factor) , and cosphi_p (set_p1,  set_p2, s_nom, power_factor_min, p_ref) .",Input (optional)
```
fneum, line 30: I could imagine renaming `type_of_control_strategy` to `control_strategy` throughout for brevity.

pypsa/component_attrs/generators.csv
```diff
@@ -2,37 +2,49 @@ attribute,type,unit,default,description,status
...
+power_factor,static or series,n/a,1,"Power factor is default to 1 and can take values (1 = >  power_factor  >  0), required when the type_of_control_strategy is chosen 'fixed_cosphi' and 'q_v'.",Input (optional)
```
fneum, line 31:
```suggestion
power_factor,static or series,n/a,1,"Power factor is default to 1 and can take values (0,1], required when the ``type_of_control_strategy`` is chosen ``fixed_cosphi`` and ``q_v``.",Input (optional)
```
could you do the highlighting with two `` like in the example above? Also applies to the other explanations.

pypsa/component_attrs/generators.csv
```diff
@@ -2,37 +2,49 @@ attribute,type,unit,default,description,status
...
+s_nom,float,MVA,1,"Inverter nominal apparent power, required when the type_of_control_strategy is chosen 'q_v'  or  'cosphi_p'.",Input (optional)
+v1,float,n/a,0.9,"Minimum voltage magnitude per unit for left curve of q_v controller, required when the type_of_control_strategy is chosen  'q_v'.",Input (optional)
```
fneum, line 33: I find it not immediately intuitive, what v1,v2,v3,v4 mean. Maybe the description has to be more precise, e.g. "left curve"?

setup.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-
+from copy import deepcopy
```
fneum, line 2: why do you need this in `setup.py`?

test/networks/results_no_c/citygrid/generators.csv
```diff
@@ -0,0 +1,6 @@
+name,bus,control,min_up_time,min_down_time,initial_status
+EEA1,N10,PQ,0,0,1
+EEA11,N5,PQ,0,0,1
+EEA12,N8,PQ,0,0,1
+EEA121,N15,PQ,0,0,1
+NE_S1,N4,Slack,0,0,1
```
fneum, line 6: can omit `min_up_time` and `min_down_time` because they do not deviate from defaults

pypsa/inverter_functions.py
```diff
@@ -0,0 +1,512 @@
...
+import logging
+import numpy as np
+from copy import deepcopy
+logger = logging.getLogger(__name__)
+
+
+def fixed_cosphi(p_input, now, c_attrs, c_list_name):
+    """
+    fix power factor inverter controller.
+    reference : https://www.academia.edu/24772355/
```
fneum, line 11: this link is not accessible without a google or facebook account? Is this on arxiv or open access somewhere?

Generally, I would like it if the documentation and docstrings were self-contained with academic papers listed as additional references.

Can you explain the functionality in a few sentences directly in the docstring?

pypsa/inverter_functions.py
```diff
@@ -0,0 +1,512 @@
...
+        Is the new reactive power that will be set as new q_set in each
+        controlled component as a result of applying this controller.
+    """
+    # needed parameters
+    p = p_input.loc[now, c_attrs.index]
+    power_factor = c_attrs['power_factor']
+    s_nom = c_attrs['s_nom']
+    setting_p_set_required = False
+    new_p_set = 0
+    Q_inv_cap, Q_allowable, Q_max = find_allowable_q(p, power_factor, s_nom)
```
fneum, line 37: For code style, please avoid starting variable names with capital letters. `Q_max`->`q_max`

pypsa/inverter_functions.py
```diff
@@ -0,0 +1,512 @@
...
+        setting_p_set_required = True
+        new_p_set = adjust_p_set(s_nom, q_set, p, c_list_name, 'fixed_cosphi')
+
+    return q_set, new_p_set, setting_p_set_required
+
+
+def cosphi_p(p_input, now, df, df_t, c_attrs, time_varying_p_set, c_list_name):
+    """
+    Power factor as a function of active power (cosphi_p) controller.
+    reference : https://ieeexplore.ieee.org/document/6096349.
```
fneum, line 52: Same as above. Provide a brief description of the control strategy here, but feel free to keep the paper as additional reference. Maybe provide the doi link instead.

pypsa/inverter_functions.py
```diff
@@ -0,0 +1,512 @@
...
+    logger.info(" Some p_set in '%s' component with %s due to reactive power "
+                "compensation priority. ", c_list_name, log_info)
+
+    return new_p_set
+
+
+def q_v(c_list_name, now, n_trials_max, n_trials, p_input, v_pu_bus, c_attrs):
+    """
+    Reactive power as a function of voltage Q(U).
+    reference : https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=6096349
```
fneum, line 156: Same as above. Provide a brief description of the control strategy here, but feel free to keep the paper as additional reference. Maybe provide the doi link instead.

pypsa/pf.py
```diff
@@ -170,8 +169,8 @@ def _network_prepare_and_run_pf(network, snapshots, skip_pre, linear=False,
     if not linear:
         return Dict({ 'n_iter': itdf, 'error': difdf, 'converged': cnvdf })
 
-def network_pf(network, snapshots=None, skip_pre=False, x_tol=1e-6, use_seed=False,
-               distribute_slack=False, slack_weights='p_set'):
+def network_pf(network, snapshots=None, skip_pre=False, x_tol=1e-6, x_tol_outer=1e-3, use_seed=False,
+               distribute_slack=False, slack_weights='p_set', inverter_control=False):
```
fneum, line 173: The new arguments should be documented in the docstring below.

pypsa/pf.py
```diff
@@ -316,8 +315,8 @@ def sub_network_pf_singlebus(sub_network, snapshots=None, skip_pre=False,
     return 0, 0., True # dummy substitute for newton raphson output
 
 
-def sub_network_pf(sub_network, snapshots=None, skip_pre=False, x_tol=1e-6, use_seed=False,
-                   distribute_slack=False, slack_weights='p_set'):
+def sub_network_pf(sub_network, snapshots=None, skip_pre=False, x_tol=1e-6, x_tol_outer=1e-3, use_seed=False,
+                   distribute_slack=False, slack_weights='p_set', inverter_control=False):
```
fneum, line 319: The new arguments should be documented in the docstring below.

pypsa/components.py
yama519: yes,
i did not notice it.

binder/environment.yml
yama519: will fixt it!

pypsa/inverter_functions.py
yama519: i will consider it.

examples/minimal_example_pf.py
yama519: no, it did not, no problem if removed. i will update it

pypsa/component_attrs/generators.csv
yama519: yes, no problem, i will update them with " "

pypsa/component_attrs/generators.csv
yama519: Yes you are right, i was not sure to add description here or in https://pypsa.readthedocs.io/en/latest/power_flow.html. because for understanding  v1 ...v2 and other controller parameters, controller droop curve in a graph is required  which is not possible here.

I think if i add a brief decription here and then put a link to pypsa.readthedocs for further details would make sense?

setup.py
yama519: i was confused, considering pull request checklist, i will remove it.

FabianHofmann:
Hey @yama519  I'm also having a small look through the code. For the generic style, I would like to point out some tips that would help to improve the code and align it to the pypsa code. 

For example the function `cosphi` seems to have very bloated argument list. If I see correctly it should be possible to boil it down to 
```
def cosphi_p(n, snapshot, c, index=None):
```
where `n` is the network, `snapshot` the current snapshot to consider, `c` the component name, e.g. 'Generator', 'StorageUnit'. 
The other variables are now given by 
* `p_input` -> `n.pnl(c).p`
* `df` -> n.df(c)
* `df_t` -> `n.pnl(c)`
* `c_attrs` -> `n.df(c)` (seems to be) 
* `index` can determine a subset of `n.df(c).index`, replacing the `c_attrs.index`. Default should be `n.df(c).index`
* The `time_varying_p_set` can be made obsolete using the function `pypsa.descriptors.get_switchable_as_dense()` in the `apply_controller` function. It combines time varying and static values. Import it as `as_dense` then the values are combined with `as_dense(n, c, attr, snapshot)`

This should be applied to the other functions as well. 

It seems it would be handier to make the inplace changes to in already in `cosphi_p`, `q_v` and `fixed_cosphi`. Then predefining a lot of variables in `apply_controller` could be removed as well as `_set_controller_outputs_to_n`. You could rename the functions like `cosphi_p` -> `apply_cosphi_p` for making the inplace changes clear. 


The `parameter_dict` dict is a real monster :) Isn't there a more elegant way?

Great that you added a test!

yama519:
@FabianHofmann @fneum 

Thanks for your comments and suggenstions:)

Considering the comments and suggestions i have updated the code. Although i added the required descriptions and updated the doc strings but i will add a PDF file which explains the concept of each controller and its application in PyPSA which might be also useful during your review.

During my editing i have realized that i still need to work on Store and StorageUnit components and the changes i will make to complete these two components would be on function _set_controller_outputs_to_n function in control.py . The current code works fine for Load and Generator components as in the provided examples and the tests.

Please let me know if you think or decided that application of inverter control idea in any of the four considered components is not necessary due to duplication, for instance in Store and StoragUnit or in Load which is a rare case.

Thank you

yama519:
> Hey @yama519 I'm also having a small look through the code. For the generic style, I would like to point out some tips that would help to improve the code and align it to the pypsa code.
> 
> For example the function `cosphi` seems to have a very bloated argument list. If I see correctly it should be possible to boil it down to
> 
> ```
> def cosphi_p(n, snapshot, c, index=None):
> ```
> 
> where `n` is the network, `snapshot` the current snapshot to consider, `c` the component name, e.g. 'Generator', 'StorageUnit'.
> The other variables are now given by
> 
> * `p_input` -> `n.pnl(c).p`
> * `df` -> n.df(c)
> * `df_t` -> `n.pnl(c)`
> * `c_attrs` -> `n.df(c)` (seems to be)
> * `index` can determine a subset of `n.df(c).index`, replacing the `c_attrs.index`. Should default to  `n.df(c).index`
> * The `time_varying_p_set` can be made obsolete using the function `pypsa.descriptors.get_switchable_as_dense()` in the `apply_controller` function. It combines time varying and static values. Import it as `as_dense` then the values are combined with `as_dense(n, c, attr, snapshot)`
> 
> This should be applied to the other functions as well.
> 
> It seems it would be handier to make the inplace changes already in `cosphi_p`, `q_v` and `fixed_cosphi`. Then predefining a lot of variables in `apply_controller` could be removed as well as `_set_controller_outputs_to_n`. You could rename the functions like `cosphi_p` -> `apply_cosphi_p` for making the inplace changes clear.
> 
> The `parameter_dict` dict is a real monster :) Isn't there a more elegant way?
> 
> Great that you added a test!

Hey @FabianHofmann, Sorry for late reply
Thanks again for your hints, you can see them all in the code,

Regarding your question about 'parameter_dict', which is currently renamed to "prepare_controlled_index_dict" 
yes there are other ways as well, as follow:

- Using 'prepare_controlled_index_dict' to only include controlled indexes inside each controller name (dictionary keys)
  which can be seen in the current code (i prefered this method).

- including the content of 'prepare_controlled_index_dict' in 'apply function', but this way, would make the power flow slower, 
 since the content of 'prepare_controlled_index_dict' need to be done only once such as filtering controlled indexes, assertions, 
 n_trials_max and so on, whereas inside 'apply function' they will repeated per snapshot and per iterations.

- Another solution would be to include the controlled indexes in n.pf(controlled_indexes=....), in this case 
 'prepare_controlled_index_dict' will only return n_trial_max.

yama519:
@FabianHofmann @fneum 
Here is the link for the PDF document which explains the application of control strategy logic in PyPSA

https://github.com/yama519/Supportive-docs/blob/master/Read_Controllers.pdf 

Please let me know if something is not clear or you need further docmentation, or further support from my side to convey the main logic of my idea.

Thank You

yama519:
Hi,

I was confused reagarding Store and StorageUnit components.
I thought that state of charge is also considered in pf and as i did not consider it in the current code

> @FabianHofmann @fneum
> 
> Thanks for your comments and suggenstions:)
> 
> Considering the comments and suggestions i have updated the code. Although i added the required descriptions and updated the doc strings but i will add a PDF file which explains the concept of each controller and its application in PyPSA which might be also useful during your review.
> 
> During my editing i have realized that i still need to work on Store and StorageUnit components and the changes i will make to complete these two components would be on function _set_controller_outputs_to_n function in control.py . The current code works fine for Load and Generator components as in the provided examples and the tests.
>I rephrase and change this part of my message:, I was confused about the use of state of charge for Store and StoragUnit components, and thought that, state of charge should be also considered as it is in LOPF which i did not consider, but since pf outputs for Store and StorageUnit contains only  p and q as outputs, therefore the current code is correct.
> Please let me know if you think or decided that application of inverter control idea in any of the four considered components is not necessary due to duplication, for instance in Store and StoragUnit or in Load which is a rare case.
> 
> Thank you

yama519:
Hi, @fneum 

I noticed that Travis CI test was failing for my test script "test_control_strategies_against_cerberus.py" file, i fixed it but updating the files to PyPSA 0.17.1 caused Travis Ci failure for "test_sclopf_scigrid.py" , I hope this is not a problem for PR review.

Thank you!

yama519:
Hi @FabianHofmann
I have updated the doc string at some functions to make them easy understandable.  And in 'prepare_controlled_index_dict' only controlled index are stored (instead of 'c_attrs' which was the complete df of controlled indexes) and later they are passed for each controller in 'apply_controller' function.  But regarding part of the suggestions saying 'inplace changes already inside each controller' if i understood correctly meaning that controller outputs should be set to the network inside each controller instead of passing it to 'set_controller_outputs_to_n'. If i consider it then i would need to repeat the content of 'set_controller_outputs_to_n' for each controller, but if i miss understood it and the current codes is ok then please ignore this part of my comment.

Thank You

fneum:
Closing stale PR.

PR #188, opened 2020-06-29
fix assignment of dual variables in lv_limit for final lopf

eb5194:
If a line volume is given (such as lv1.25), its dual variable after the optimisation process \mu in the global constraints is NaN, because the last iteration of the lopf fixes lines and links. Hence, we use the second last dual.

Additionally, if track_iterations is true, the global constraint shadow variables \mu are being tracked as well.

### DIFF
--- a/pypsa/linopf.py
+++ b/pypsa/linopf.py
@@ -1007,6 +1007,7 @@ def ilopf(n, snapshots=None, msq_threshold=0.05, min_iterations=1,
         setattr(n, f"status_{iteration}", status)
         setattr(n, f"objective_{iteration}", n.objective)
         n.iteration = iteration
+        n.global_constraints = n.global_constraints.rename(columns={'mu': f'mu_{iteration}'})
 
 
     if track_iterations:
@@ -1028,8 +1029,6 @@ def ilopf(n, snapshots=None, msq_threshold=0.05, min_iterations=1,
                                 f'and termination {termination_condition}')
         if track_iterations:
             save_optimal_capacities(n, iteration, status)
-            n.global_constraints = n.global_constraints.rename(columns={'mu': f'mu_{iteration}'})
-        final_duals = n.global_constraints.mu.copy() #save second last iteration of duals (needed for lv_limit duals)
         update_line_params(n, s_nom_prev)
         diff = msq_diff(n, s_nom_prev)
         iteration += 1
@@ -1041,11 +1040,6 @@ def ilopf(n, snapshots=None, msq_threshold=0.05, min_iterations=1,
     kwargs['warmstart'] = False
     network_lopf(n, snapshots, **kwargs)
     if track_iterations:
-        n.global_constraints = n.global_constraints.rename(columns={'mu':f'mu_{iteration}'})
-    if 'lv_limit' in n.global_constraints.index:
-        logger.info(f'Resulting dual value for line volume limit (lv_limit) set to the one of iteration {iteration-1} '
-                    'due to its optimisation process.') # otherwise, final dual for lv_limit is NaN.
-        if track_iterations: n.global_constraints.at['lv_limit', f'mu_{iteration}'] = final_duals['lv_limit']
-        else: n.global_constraints.at['lv_limit', f'mu'] = final_duals['lv_limit']
+        save_optimal_capacities(n, iteration, status)
     n.lines.loc[ext_i, 's_nom_extendable'] = True
     n.links.loc[ext_links_i, 'p_nom_extendable'] = True
--- a/pypsa/linopf.py
+++ b/pypsa/linopf.py
@@ -1028,6 +1028,8 @@ def ilopf(n, snapshots=None, msq_threshold=0.05, min_iterations=1,
                                 f'and termination {termination_condition}')
         if track_iterations:
             save_optimal_capacities(n, iteration, status)
+            n.global_constraints = n.global_constraints.rename(columns={'mu': f'mu_{iteration}'})
+        final_duals = n.global_constraints.mu.copy() #save second last iteration of duals (needed for lv_limit duals)
         update_line_params(n, s_nom_prev)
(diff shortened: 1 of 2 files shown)


pypsa/linopf.py
```diff
@@ -1038,5 +1040,12 @@ def save_optimal_capacities(n, iteration, status):
     n.links[['p_nom', 'p_nom_extendable']] = n.links['p_nom_opt'], False
     kwargs['warmstart'] = False
     network_lopf(n, snapshots, **kwargs)
+    if track_iterations:
+        n.global_constraints = n.global_constraints.rename(columns={'mu':f'mu_{iteration}'})
+    if 'lv_limit' in n.global_constraints.index:
+        logger.info(f'Resulting dual value for line volume limit (lv_limit) set to the one of iteration {iteration-1} '
+                    'due to its optimisation process.') # otherwise, final dual for lv_limit is NaN.
+        if track_iterations: n.global_constraints.at['lv_limit', f'mu_{iteration}'] = final_duals['lv_limit']
```
fneum, line 1048:
```suggestion
        if track_iterations:
            n.global_constraints.at['lv_limit', f'mu_{iteration}'] = final_duals['lv_limit']
```
style comment: would indent here

I was wondering whether it would make sense to keep the `mu_{iteration}` of the last iteration as NaN and not override it (so it becomes clear that you use the "last iteration" heuristic).

pypsa/linopf.py
```diff
@@ -1038,5 +1040,12 @@ def save_optimal_capacities(n, iteration, status):
...
+        else: n.global_constraints.at['lv_limit', f'mu'] = final_duals['lv_limit']
```
fneum, line 1049: Wouldn't you want to do this in any case (when iterating) not just if you track iterations?

pypsa/linopf.py
```diff
@@ -1028,6 +1028,8 @@ def save_optimal_capacities(n, iteration, status):
                                 f'and termination {termination_condition}')
         if track_iterations:
             save_optimal_capacities(n, iteration, status)
+            n.global_constraints = n.global_constraints.rename(columns={'mu': f'mu_{iteration}'})
+        final_duals = n.global_constraints.mu.copy() #save second last iteration of duals (needed for lv_limit duals)
```
fneum, line 1032: When you track iterations, don't you get rid of the 'mu' column in the first row and in the second you try to access it? Does that work? I haven't tried.

fneum:
Note: merge master into this branch once #190 is in it, so the tests pass again.

FabianHofmann:
Hey @eb5194 , I would rather keep it general and only add one line 

```
        n.global_constraints[f'mu_{iteration}'] = n.global_constraints.mu
```
to the function `save_optimal_capacities` (the function name does not suit anymore, but it's an inner function). This keeps it simple and does not overwrite the last iteration, which makes it programmatically correct - you don't have a shadow price when you not allow line expansion. The value you want is accessable by `n.global_constraints[f'mu_{n.iteration}']` then. 

(Note also: the line volume constraint does not necessarily have the name 'lv_limit', this is just the name set in pypsa-eur, you would rather look at the `n.global_constraints.type` column).

pypsa/linopf.py
eb5194: If the iterations are being tracked, I was thinking that as well. However, if not, I'd overwrite it.

pypsa/linopf.py
eb5194: Huh, true, but it worked when I tried. That's strange. I'll change it, just to be sure

pypsa/linopf.py
eb5194: Not sure I get your question. The name of the column in n.global_constraints is different, that's why I added the if/else, and the logger.info is displayed in any case. But if we track iterations and want to remain with NaN for the last iteration, this part will be different.

eb5194:
> Hey @eb5194 , I would rather keep it general and only add one line
> 
> ```
>         n.global_constraints[f'mu_{iteration}'] = n.global_constraints.mu
> ```
> 
> to the function `save_optimal_capacities` (the function name does not suit anymore, but it's an inner function). This keeps it simple and does not overwrite the last iteration, which makes it programmatically correct - you don't have a shadow price when you not allow line expansion. The value you want is accessable by `n.global_constraints[f'mu_{n.iteration}']` then.
> 
> (Note also: the line volume constraint does not necessarily have the name 'lv_limit', this is just the name set in pypsa-eur, you would rather look at the `n.global_constraints.type` column).

Sounds good! But if iterations are not being tracked, leave it then with a NaN?

FabianHofmann:
> Sounds good! But if iterations are not being tracked, leave it then with a NaN?

Yes, that would be the correct way.

pypsa/linopf.py
```diff
@@ -1038,5 +1039,7 @@ def save_optimal_capacities(n, iteration, status):
     n.links[['p_nom', 'p_nom_extendable']] = n.links['p_nom_opt'], False
     kwargs['warmstart'] = False
     network_lopf(n, snapshots, **kwargs)
+    if track_iterations:
+        save_optimal_capacities(n, iteration, status)
```
FabianHofmann, line 1043:
```suggestion
```
This seems to be a duplicate of line 1030/1031

eb5194: it is, but otherwise, after the last lopf, the duals remain un-iterated as network_lopf is run again in line 1041. (lines 1026/27 within the iterative optimisation process before), so I prefer to rename the mus again for consistency, but we can leave the last iteration un-indexed as well

FabianHofmann: Alright, I get you, but on the other hand the 'last' index is exactly the final outcome (resulting mu), right?

eb5194: It isn't for the global_constraints (there is no mu without index), but it is for the other results that are being tracked (such as objective, status, etc.) I think it's easier to remove these two lines, you're right

PR #197, opened 2020-08-10
Add nominal bounds for bus & carrier

FabianHofmann:

Support lower and upper capacity constraints per bus and carrier. With this feature the combined capacity of components of the same carrier at the same bus can be constrained by setting values in `n.buses['nom_max_{carrier}']` / `n.buses['nom_min_{carrier}']` (where carrier is a dummy for wind, solar etc. )

Test it with 

```
import pypsa
from pathlib import Path

n = pypsa.Network(str(Path(pypsa.__file__).parent.parent / 'examples'/
                      'ac-dc-meshed'/ 'ac-dc-data'))

n.add('Generator', 'Frankfurt Wind 2', bus='Frankfurt', capital_cost = 120,
      carrier='wind', p_nom_extendable=True)

n.buses.loc[['Frankfurt', 'Manchester'], 'nom_min_wind'] = 2000
n.buses.loc[['Frankfurt'],'nom_max_wind'] = 2200

n.lopf(pyomo=False)

print(n.generators.p_nom_opt)

```

## Checklist

- [x] ~Unit tests for new features were added (if applicable).~
- [x] ~Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).~

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -7,7 +7,7 @@ Upcoming Release
 
 .. warning:: The features listed below are not released yet, but will be part of the next release! To use the features already you have to install the ``master`` branch, e.g. ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
 
-* First new feature
+* When solving ``n.lopf(pyomo=False)``, PyPSA now supports setting lower and upper capacity bounds per bus and carrier. These are specified in the columns ``n.buses['nom_min_{carrier}']`` and ``n.buses['nom_max_{carrier}']`` respectively. For example, if multiple generators of carrier "wind" are at bus "bus1", the combined capacity is limited to 1000 MW by setting ``n.buses.loc['bus1', 'nom_max_wind'] = 1000`` (a minimal capacity is forced by setting ``n.buses.loc['bus1', 'nom_min_wind']``). In the same manner the combined ``p_nom`` of components ``StorageUnit`` and ``e_nom`` of components ``Store`` can be limited.  
 
 PyPSA 0.17.1 (15th July 2020)
 =============================
--- a/examples/notebooks/capacity-constraint-per-bus.ipynb
+++ b/examples/notebooks/capacity-constraint-per-bus.ipynb
@@ -1,0 +1,135 @@
+{
+ "cells": [
+  {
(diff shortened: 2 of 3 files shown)


fneum (changes requested):
There's some code duplication in the two loops. I was wondering whether one could simply reduce this by looping over the senses:

```py
for bound, sense in [("max", "<="), ("min", ">=")]:
    for carrier in n.carriers.index:
        col = f'nom_{bound}_{carrier}'
...
        define_constraints(n, lhs, sense, rhs, 'Bus', 'mu_' + col)
```

The other difference is how limits at buses with no extendable components are handled:

Upper limits are less strict and handle exceptions:

```py
        lhs = lhs[lhs != '']
        rhs = rhs.reindex(lhs.index)
```

whereas lower bounds don't and are more restrictive

```py
        assert (lhs != '').all(), (
            f'No extendable components of carrier {carrier} on bus '
            f'{list(lhs[lhs == ""].index)}')
```

Would be nice to unify this, right?

FabianHofmann:
Yes, indeed we can unify the two loops. Give me a second

fneum:
If you merge the master, tests should pass.

While you are at it can you add documentation of `nom_max_{carrier}` to the component attributes at

https://github.com/PyPSA/PyPSA/blob/master/pypsa/component_attrs/buses.csv

?

FabianHofmann:
> If you merge the master, tests should pass.
> 
> While you are at it can you add documentation of `nom_max_{carrier}` to the component attributes at
> 
> https://github.com/PyPSA/PyPSA/blob/master/pypsa/component_attrs/buses.csv
> 
> ?

Unfortunate that's will end up in a weird extra column, n.buses['nom_min_{carrier}'] (with exactly this name) when loading any network. So I wouldn't do it. Probably it makes more sense to add a proper example jupyter notebook how to use it.

fneum:
> Unfortunate that's will end up in a weird extra column

Ah, I see. I just wanted this functionality documented somewhere else than in the release notes. An example would be fine. I think we could just use your example at the top of the PR.

fneum (approved):

PR #205, opened 2020-11-24
contigency: fix assignment in empty series

FabianHofmann:
In some newer pandas versions (not the newest) the assignment in a multiindexed series fails when using the pattern 

```
ds.at[key] = value
```

instead one has to use 

```
ds[key] = value
```
which is compatible with all pandas versions. This pr solve this issue for a line contigency.py. The line caused an error in the tests https://travis-ci.org/github/PyPSA/PyPSA/builds/730168599

**EDIT**: the second pattern is not compatitble either for all, instead restructure the code a bit.

### DIFF
--- a/pypsa/contingency.py
+++ b/pypsa/contingency.py
@@ -125,19 +125,13 @@ def network_lpf_contingency(network, snapshots=None, branch_outages=None):
         branch_outages = passive_branches.index
 
 
-    p0_base = pd.Series(index=passive_branches.index)
-
-    for c in network.passive_branch_components:
-        pnl = network.pnl(c)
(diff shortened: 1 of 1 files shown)


FabianHofmann:
ah, I see this was kind of already solved in https://github.com/PyPSA/PyPSA/commit/77c00f7016bbbe6c5baefd7af45fa4b203dbb3ab

fneum:
> ah, I see this was kind of already solved in 77c00f7

yes, but yours also works with older pandas versions than 1.1.0, which is nicer! Did you check the minimum?

fneum (approved):

PR #206, opened 2020-11-30
Initialize attributes in constructor

lukasgebhard:
It is [best practice to initialize class attributes in the constructor](https://softwareengineering.stackexchange.com/questions/357937/is-it-really-correct-to-declare-all-instance-attributes-in-init). Then, the compiler can warn developers if they try to read/write non-existing attributes (e.g., `network.busses` instead of `network.buses`). Such compiler checks are enabled by default in, e.g., [PyCharm](https://www.jetbrains.com/pycharm/).

### DIFF
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -208,6 +208,28 @@ class Network(Basic):
     def __init__(self, import_name=None, name="", ignore_standard_types=False,
                  override_components=None, override_component_attrs=None,
                  **kwargs):
+        self.buses = None
+        self.loads = None
+        self.generators = None
+        self.stores = None
+        self.storage_units = None
+        self.transformers = None
+        self.transformer_types = None
+        self.lines = None
+        self.line_types = None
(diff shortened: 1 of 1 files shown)


fneum:
@FabianHofmann do you have an opinion on this?

FabianHofmann:
Since we are having a flexible component structure, this is a bit trickier. 
Atm we are initializing all components by calling `_build_dataframes()` in the `__init__`. 

If we want to do something likewise, we should go for using  `__slots__`  which can reduce the RAM usage significantly, see [here](https://book.pythontips.com/en/latest/__slots__magic.html). This could look like this 

```python 
slots = [c['list_name'] for c in n.components.values()]
slots.remove('networks')
slots += [s + '_t' for s in slots] (+ possible other stuff like srid) 
__slots__ = slots 
```
right at the beginning of the network class definition. But it requires that `__slots__` can be modified when overwriting components, which I am not sure of.

Issue #37, opened 2018-06-19
Separate snapshot weighting from elapsed time


nworbmot:
It was discussed on the [PyPSA mailing list](https://groups.google.com/forum/#!topic/pypsa/oSKrOidIwBY) to separate the snapshot_weighting for each snapshot into two separate values, one that determines the prefactor for all snapshot-specific quantities in the objecting function, and one that corresponds to the snapshot duration, used e.g. by storage to determine how much energy is stored for a given power in each snapshot. This seems like a good idea. network.snapshot_weightings would then become a DataFrame rather than a series.

FabianHofmann:
feature added in the current master

PR #211, opened 2021-01-14
Multi decade

lisazeyen:
Closes # (if applicable).
#36 
Integrate pathway/multi-decade optimisation to PyPSA, allowing the optimisation overall several years with different investment periods.

- for all components the parameters "lifetime" and "build_year" are added. "build_year" is the first year of operation, e.g. build_year=2020, lifetime=10 than the component is active until  01-01-2030
-  'investment_periods' and 'investment_period_weightings' are added analogous to 'snapshots' and snapshots_weightings'. They represent the time of investment and the weighting of the investment period. They are set analogs to snapshots with 'n.set_investment_periods'. The default value for the 'investment_period_weightings' is currently 1. This could be changed to the time span between the investment points, question would be how to value the last time period, e.g. investment_periods=[2020,2025, 2040,2050] than the weightings could be [5, 5, 15, 10, ?1?] 
-  investment_periods are not specified, it falls back to "normal pypsa" optimisation with one investment at the first snapshot
-  currently only for pyomo=False
-  the structure of how the capacities are stored is not changed. Every component has an optimal capacity which can be used during the time span build year and build year plus lifetime with certain capacity and costs. I wrote a little function which returns as rows investment periods and as columns the different components if one wants to see in a clearer way when the capacity can be used.
- new attribute for GlobalConstraints called "investment_period", which specifies the time frame in which the constraint is valid, if not specified, Constraint applies to all investment periods
- it is possible to specify a CO2 emission budget when adding a global constraint (type="Budget"), which limits the total emissions over all investment periods, as well as defining a limit for single investment periods (as before type="primary energy")
- there is a [jupyter notebook](https://github.com/lisazeyen/PyPSA/blob/multi-decade/test/multi-decade-test.ipynb) and a [python](https://github.com/lisazeyen/PyPSA/blob/multi-decade/test/multi-decade-test.py) test script, paths to the pypsa-eur directory for testing have to be modified
- 
## Checklist
- [ ] global constraint for limiting the installable capacity of one technology (e.g. 'onwind') at each node for each investment_period
- [ ] testing (transformers, committable generators, IO)
- [ ] split snapshot_weightings into weighting during time period and weighting between time periods issue #37 , [discussion](https://github.com/PyPSA/EnergyModels.jl/issues/14), [modelling seasonal storage Leander](https://arxiv.org/abs/1710.07593)
- [ ] change scripts for pyomo=True, currently only pyomo=False


test/test_pf_against_pandapower.py
```diff
@@ -1,3 +1,17 @@
...
+             '/home/ws/bw0928/miniconda3/envs/test-pypsa/lib/python3.6/lib-dynload',
+             '',
+             '/home/ws/bw0928/.local/lib/python3.6/site-packages',
+             '/home/ws/bw0928/miniconda3/envs/test-pypsa/lib/python3.6/site-packages',
+             '/home/ws/bw0928/miniconda3/envs/test-pypsa/lib/python3.6/site-packages/IPython/extensions',
+             '/opt/gurobi900/linux64/bin',
+             '/opt/ibm/ILOG/CPLEX_Studio129/cplex/python/3.7/x86-64_linux',
+             '/home/ws/bw0928/Dokumente/MA/Code',
+             '/home/ws/bw0928/.ipython']
+
```
nworbmot, line 14: Try and avoid adding code only specific to your computer setup.

nworbmot (commented):
I've played with the example and looked through most of the code and I'm super impressed! Great job Lisa! I actually wouldn't change very much, just the default behaviour for old users and introduce a very conservative suggestion for the ``global_constraints``.

## Naming schemes

I agree with your naming: we need ``investment_periods`` to distinguish from ``snapshot_periods`` that run through the year. Old PyPSA "overnight" behaviour is often called "multi-period" optimisation, but we need to make clear that we also do multiple investment periods here.

## Default behaviour

The guiding principle is that old users shouldn't have to make any big changes, and that there should be no surprises.

I suggest that n.snapshots stays the same, and can be either MultiIndex or Index.

All PyPSA functions should by default treat MultiIndex snapshots just the same as Index snapshots. This is the case for old PyPSA ``n.lopf(pyomo=False)``, but ``n.lopf(pyomo=True)``, ``n.lpf()`` and ``n.pf()`` all break when using MultiIndex -we should fix this in future releases (but not now).

The multi-period optimisation is the only one to treat MultiIndex differently, and must be activated with:

``n.lopf(pyomo=False, multi_investment_period=True)``

When this is activated, it checks that ``n.snapshots`` is MultiIndex, that level 0 is integer, and that the integers are increasing. It also checks that all investment periods appear in the index of ``n.investment_period_weightings``.

If ``n.lopf(pyomo=True, multi_investment_period=True)`` is called, an error is thrown.

### Snapshot weightings

This is a parallel change to the investment periods. In PyPSA Version 0.18.0 we will replace the pd.Series ``n.snapshot_weightings`` with a pd.DataFrame with three columns ``[objective_weightings, generator_weightings, store_weightings]``. This separates out the three current use cases of weightings in the LOPF:

- ``objective_weightings``: for operational costs in the objective function
- ``store_weightings``: for elapsed time in the filling of ``Store`` and ``StorageUnit`` components
- ``generator_weightings``: for the contribution of generators to ``GlobalConstraint`` components for e.g. CO2 budgets

Example use cases:

- case 1 3-hourly sampling for full year: ``objective_weighting = generator_weighting = store_weighting = 3``.
- case hourly sampling for 10 days to represent full year with cyclic storage: ``objective weighting = generator_weighting = 52. , 
store_weighting = 1.``

BUT this case fails if e.g. a ``Store`` represents non-cyclic depletion of e.g. biomass or some other finite resource, in which case you want ``store_weighting = 52``, like ``generator_weighting``. This interface need all storage doing short-term cycling, or all storage non-cyclic (depleting a finite resource). We should fix this when we implement Kotzur's seasonal storage scheme.

### Investment period weightings

I think the way Lisa implemented this looks good, maybe change the names of weightings to be similar to the ``snapshot_weightings``?

I also suggest not to have ``n.investment_periods``, but track these only via level 0 of ``n.snapshots``.

### Behaviour when an empty network is generated

If you call ``n = pypsa.Network()`` by default will be created:

- ``n.snapshots = pd.Index(["now"])``
- ``n.snapshot_weightings = pd.DataFrame(1.,1.,1.,index=["now"],
                                     columns=["objective_weightings","store_weightings","generator_weightings"])``
- ``n.investment_period_weightings = pd.DataFrame(columns=["objective_weightings","store_weightings","generator_weightings"])``

### Behaviour when a network from PyPSA Version < 0.18.0 is imported

Create new empty ``n.investment_period_weightings``.

The values of pd.Series ``n.snapshot_weightings`` are copied across all three values of the new pd.DataFrame ``n.snapshot_weightings``.

## Global constraints for ``p_nom_max`` at each bus, carrier and investment_period

I suggest to keep these all in ``n.global_constraints`` since they are global in the sense of affecting multiple components at once. We need to add to the ``GlobalConstraint`` table attributes ``bus``, ``carrier`` and ``investment_period`` and type ``multi_investment_period_p_nom_max`` or similar.

Yes, this will create lots of global constraints, but introducing e.g. ``n.global_constraints_p`` doesn't help since you need 3-d information.

This way is minimally invasive.

## Treatment of sub_networks

I think the current way is good. The sub_networks are generated as needed, and the last one to be generated is stored. There is no need to store them from all investment periods, i.e. no need for ``n.sub_networks_p``.

The sub_network is anyway more of a hidden variable.

## Storage behaviour between investment periods

Did I understand correctly that cyclic storage is cyclic within each investment period? This behaviour would be fine.

If the storage is not cyclic, is energy transfered between investment periods? Probably this is not such a good idea for e.g. depletable biomass. But maybe we should model depletable biomass as a generator, rather than as a store that starts full. Then the generator can fill the storage at the start of each year?

## Check objective function

Since the objective function is quite complicated now, please check for some non-trivial examples that the Gurobi objective function is also what you calculated from the results. This should help to find bugs.

## Coding style

Try to avoid abbreviations like ``inv_p`` and ``sns`` in API functions like ``network.determine_network_topology``. Here we should be explicit for understandability. You can abbreviate inside functions, that's fine.

Also try to avoid committing code very specific to your own setup, like paths to libraries.

FabianHofmann:
Some thoughts from my side: 

1. Why do we need an extra flag `multi_investment_periods` in the `n.lopf`? we could just say if snapshots have a multiindex -> solve in multi-period mode, if  single-level index -> in overnight mode? 
2. I would advocate for a shorter name than `investment_period`, I think we're better off with `period` or `span` which is much snappier. The information that it is about the investment should be clear once you read the doc (we also don't say `operational_snapshot`). 
3. For not having a breaking change with the `snapshot_weightings`, we can add a new attribute `weightings` (a new class or a `descriptor.Dict`) which contains two object `snapshot` and `period`. Then `n.weightings.snapshot` is a DataFrame as Tom described. In case the users only defines the `snapshot_weightings` all columns in n.weightings.snapshot default to `n.snapshot_weightings`. The second, `n.weightings.period` is a Series with the duration of the investment period. Which brings me to the next question
4. Do we need specific generator, store and objective weightings for the investment period? I can hardly think of an example where you would have different generator weighting then objective weightings for an investment period. 
5. For generalization the `determine_network_topology()` function should not take the investment_period as an argument but `branches`. So one pulls out the calculation of the active branches out of the function.

nworbmot:
After our meeting we decided:

1. Keep the naming ``investment_periods`` to distinguish operational periods/snapshots from investment periods.
2. Cyclic storage should be cycling within investment period.
3. Non-cyclic storage should have two options with True/False switch: one where ``e`` is set to ``e_initial`` only at the start of the whole simulation and energy can be passed between investment periods and one where ``e`` is reset to ``e_initial`` at the start of each investment period.
4. We should make e.g. ``n.pf()`` work for multiIndex so that you can take results from ``multi_investment_periods=True`` and check non-linear power flow.
5. Long-term: make ``pyomo=False`` the default.
6. Long-term: make ``nomopyomo`` useable outside PyPSA framework so users can become familiar with it, then PyPSA uses it as a library. Also other software can then use it's speed and low resource usage.

PR #212, opened 2021-01-19
FutureWarning

martacki:
The following FutureWarning affects `io.py` and `linopf.py`:
`Index.__or__ operating as a set operation is deprecated, in the future this will be a logical operation matching Series.__or__.  Use index.union(other) instead` & `Index.__and__ operating as a set operation is deprecated, in the future this will be a logical operation matching Series.__and__.  Use index.intersection(other) instead`

As suggested, `.union()` instead of a logical `or`/ `|` and `.intersection()` instead od a logical `and`/`&` should avert the problem

If you find other scripts that print the same Warning, comment or include them into this PR? Thanks

### DIFF
--- a/pypsa/io.py
+++ b/pypsa/io.py
@@ -766,9 +766,9 @@ def import_series_from_dataframe(network, dataframe, cls_name, attr):
         dataframe = dataframe.reindex(network.snapshots, fill_value=default)
 
     if not attr_series.static:
-        pnl[attr] = pnl[attr].reindex(columns=df.index | columns, fill_value=default)
+        pnl[attr] = pnl[attr].reindex(columns=df.index.union(columns), fill_value=default)
     else:
-        pnl[attr] = pnl[attr].reindex(columns=(pnl[attr].columns | columns))
+        pnl[attr] = pnl[attr].reindex(columns=(pnl[attr].columns.union(columns)))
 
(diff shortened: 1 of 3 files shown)


FabianHofmann:
looks good. Did you have a `pytest` run? there you see all the deprecation warnings.

fneum:
Interesting. Here's the referring issue https://github.com/pandas-dev/pandas/issues/36758. Do not merge yet, as there are many more cases of set operations on `pd.Index` objects in the code. Can you push them @eb5194? Also, we need to go through PyPSA-Eur and PyPSA-Eur-Sec. Dooh that will make the code verbose.......

eb5194:
> Interesting. Here's the referring issue [pandas-dev/pandas#36758](https://github.com/pandas-dev/pandas/issues/36758). Do not merge yet, as there are many more cases of set operations on `pd.Index` objects in the code. Can you push them @eb5194? Also, we need to go through PyPSA-Eur and PyPSA-Eur-Sec. Dooh that will make the code verbose.......

I can, but it's hard to detect all of them. I ran the tests, but that only showed one more of this one. However, there are many other DeprecationWarnings. I could try to resolve those as well in the same PR.

fneum:
Thanks @martacki. Only found a couple more in the unit commitment section.

PR #229, opened 2021-04-30
Add rudimentary test for import and export functionality

fneum:
Adds three tests that check whether exports and imports as CSV, netCDF, and HDF5 run through. No content check.


FabianHofmann (approved):
Good idea!

PR #232, opened 2021-05-03
deprecate geo.area_from_lon_lat_poly

fneum:
widely unused and dependencies that are not tracked.

@FabianHofmann is this a good way to also handle deprecations in the future?


FabianHofmann (approved):
I like the deprecated decorator! We should not forget to add new dependencies to the conda recipe as well

PR #235, opened 2021-05-06
Multi indexed snapshots

FabianHofmann:

Second step of preparing #211 :

Support multiindexed snapshots. All functionalities except for optimization with `pyomo` now work with multi-indexed snapshots (including io, clustering, pf). 

Note in the `io.py` the functions `get_investment_periods`/`set_investment_periods` are already pulled from #211 but not used anywhere (yet). 

## Checklist

### DIFF
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -389,7 +389,11 @@ class Network(Basic):
         -------
         None
         """
-        self._snapshots = pd.Index(value)
+        if isinstance(value, pd.MultiIndex):
+            assert value.nlevels == 2, "Maximally two levels of MultiIndex supported"
+            self._snapshots = value.rename(['period', 'snapshot'])
+        else:
+            self._snapshots = pd.Index(value, name='snapshot')
 
         self.snapshot_weightings = (
             self.snapshot_weightings.reindex(self._snapshots, fill_value=1.))
--- a/pypsa/io.py
+++ b/pypsa/io.py
@@ -334,7 +334,7 @@ def _export_to_exporter(network, exporter, basename, export_standard_types=False
     if isinstance(network.snapshot_weightings.index, pd.MultiIndex):
         network.snapshot_weightings.index.rename(["period", "snapshot"], inplace=True)
     else:
-        network.snapshot_weightings.index.rename("snapshot")
+        network.snapshot_weightings.index.rename("snapshot", inplace=True)
     snapshots = network.snapshot_weightings.reset_index()
     exporter.save_snapshots(snapshots)
 
--- a/pypsa/io.py
+++ b/pypsa/io.py
@@ -329,9 +329,9 @@ def _export_to_exporter(network, exporter, basename, export_standard_types=False
 
     #now export snapshots
     if isinstance(network.snapshot_weightings.index, pd.MultiIndex):
-        network.snapshot_weightings.index.rename(["investment_periods", "snapshots"], inplace=True)
+        network.snapshot_weightings.index.rename(["period", "snapshot"], inplace=True)
     else:
-        network.snapshot_weightings.index.rename("snapshots")
+        network.snapshot_weightings.index.rename("snapshot")
     snapshots = network.snapshot_weightings.reset_index()
     exporter.save_snapshots(snapshots)
 
@@ -387,7 +387,7 @@ def _export_to_exporter(network, exporter, basename, export_standard_types=False
                     col_export = pnl[attr].columns[(pnl[attr] != default).any()]
 
             if len(col_export) > 0:
-                df = pnl[attr][col_export]
+                df = pnl[attr].reset_index()[col_export]
                 exporter.save_series(list_name, attr, df)
             else:
                 exporter.remove_series(list_name, attr)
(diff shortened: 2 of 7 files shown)


pypsa/components.py
```diff
@@ -389,7 +389,11 @@ def set_snapshots(self, value):
         -------
         None
         """
-        self._snapshots = pd.Index(value)
+        if isinstance(value, pd.MultiIndex):
```
lisazeyen, line 392: I would suggest to seperate the reindexing of the snapshots also depending on the type of the "value". By that one could more easily switch from a single-index network to a multi-index network without loosing the time-dependent data, e.g.         
```
 n = pypsa.Network("examples/scigrid-de/scigrid-with-load-gen-trafos") 
 n.snapshot_weightings["objective"] = 2. 
 multi_index =  pd.MultiIndex.from_product([['first'], n.snapshots])
 n.set_snapshots(multi_index)
```
One would loose the information about the objective weightings. I would suggest to change the setting of the snapshots to 
`def set_snapshots(self, value):
        """
        Set the snapshots and reindex all time-dependent data.

        This will reindex all pandas.Panels of time-dependent data; NaNs are filled
        with the default value for that quantity.

        Parameters
        ----------
        snapshots : list or pandas.Index
            All time steps.

        Returns
        -------
        None
        """
        if isinstance(value, pd.MultiIndex):
            assert value.nlevels == 2, "Maximally two levels of MultiIndex supported"
            self._snapshots = value.rename(['period', 'snapshot'])

            self.snapshot_weightings = (
                self.snapshot_weightings.reindex(self._snapshots, level=1, fill_value=1.))
        else:
            self._snapshots = pd.Index(value, name='snapshot')

            self.snapshot_weightings = (
                self.snapshot_weightings.reindex(self._snapshots, fill_value=1.))

        for component in self.all_components:
            pnl = self.pnl(component)
            attrs = self.components[component]["attrs"]

            for k,default in attrs.default[attrs.varying].iteritems():
                if isinstance(value, pd.MultiIndex):
                    pnl[k] = pnl[k].reindex(self._snapshots, level=1).fillna(default)
                else:
                    pnl[k] = pnl[k].reindex(self._snapshots).fillna(default)`

FabianHofmann: That's a good point. 
But in this case you expect snapshots to be a single index and value a single- or multi-index. What if snapshots are already a multiindex? I think it is not elegant to catch all these cases in `set_snapshots` but later in `set_investment_period` we can implement a convenience case which switches all single-indexed snapshots to multi-indexes. It would look like (I already tested it)

```python 
def set_investment_period(n, periods):
    """
    Set the investment periods of the network.

    If the network snapshots are a pandas.MultiIndex, the investment periods
    have to be a subset of the first level. If snapshots are a single index,
    they and all time-series are repeated for all periods. This changes
    the network snapshots to be a MultiIndex (inplace operation) with the first
    level being the investment periods and the second level the snapshots.

    Parameters
    ----------
    n : pypsa.Network
    periods : list
        List of periods to be selected/initialized.

    Returns
    -------
    None.

    """
    if isinstance(n.snapshots, pd.MultiIndex):
        assert set(periods).issubset(n.snapshots.unique('period')), (
            "Not all investment periods are in level `period` of snapshots.")
        n._investment_periods = periods
    else:
        # Convenience case:
        # repeat the timeseries for each period > all multiindexed
        for component in n.all_components:
            pnl = n.pnl(component)
            attrs = n.components[component]["attrs"]

            for k,default in attrs.default[attrs.varying].iteritems():
                pnl[k] = pd.concat({p: pnl[k] for p in periods})

        n._snapshots = pd.MultiIndex.from_product([periods, n.snapshots],
                                                  names=['period', 'snaphots'])
        n._snapshot_weightings = pd.concat({p: n.snapshot_weightings for p in periods})
```

Then we separate the cases: 
1. `set_snapshots` -> only hard reindexing 
2. `set_investment_periods` -> if network is single-indexed convert all to multi-indexes with repeated time-series/weightings

FabianHofmann: Then with the `class property` set up, you could simply convert your ordinary network to a multi-decade with 

```
n.investment_period = [2020, 2030, 2040]
```

lisazeyen: Yes, I like the class property and it does make it much easier! But I think it does make sense to have this property also without the multi investment periods, e.g. if you read in an old pypsa network with single index and you want to change it to a multi index without loosing all the time-dependent data. I actually wrote a small function in the multi-decade branch which catches all the cases (also multi-index back to single index) [here](https://github.com/lisazeyen/PyPSA/blob/b737a9974ee5d55eb6f11311da779d598c7bf025/pypsa/descriptors.py#L389)

```
def reindex_df(df, new_index, fill_value=1.):
    # reindex multiindex on level drops missing values, therefore seperate the cases
    # both multiindex or both single
    if isinstance(new_index, type(df.index)):
        df = df.reindex(new_index, fill_value=fill_value)
    # if multiindex -> index
    elif type(df.index) == pd.MultiIndex:
        df = df.droplevel(0).reindex(new_index, fill_value=fill_value)
    # index -> multiindex
    else:
        df = df.reindex(new_index, level=1, fill_value=fill_value)

    return df
```

fneum: I understand both perspectives and do not have a strong opinion on this. I tend to agree that `set_snapshots` should be for hard reindexing and not include the time-series copying logic. I am not sure if copying is always the desired behaviour and it has the advantage that this way there is only one way to access the copying logic.

The way to convert an old pypsa network to a simple multi-indexed network would be `n.investment_periods = ['first']`. Isn't the use case for this specific case also limited because even with single-index snapshots everything runs normally?

At first, I also thought that it would be a good idea to have a function that converts back to a single index (if there is only one investment period) for compatibility with older PyPSA versions, but I think the compatibility is already broken with the split `snapshot_weightings`, right? A compatibility function (e.g. `store_as_less_0_18()` or so) would need to address this too and restoring compatibility would only be possible when weightings are identical.

fneum (changes requested):

pypsa/io.py
```diff
@@ -305,7 +331,17 @@ def _export_to_exporter(network, exporter, basename, export_standard_types=False
     exporter.save_attributes(attrs)
 
     #now export snapshots
-    exporter.save_snapshots(network.snapshot_weightings.rename_axis('name'))
+    if isinstance(network.snapshot_weightings.index, pd.MultiIndex):
+        network.snapshot_weightings.index.rename(["period", "snapshot"], inplace=True)
+    else:
+        network.snapshot_weightings.index.rename("snapshot")
```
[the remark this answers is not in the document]
FabianHofmann: Good one!

fneum (approved):

PR #243, opened 2021-05-21
Exclude store links from fixing p_nom at last iteration

chiaroo:
The last linopf iteration fixes the link and line capacities to ensure consistency of impedances and power flow. However, when using store components, it also fixes the capacities of their respective in- and outflow links. Filtering by carrier solves this problem.
This is a quick fix which has worked well for me, I'm sure there's a more elegant solution than using a for loop - however, it is nice that no error messages are displayed, using the loc approach.

Closes # (if applicable).

Filtering by carrier before fixing p_nom links

## Checklist

- ~[ ] Unit tests for new features were added (if applicable).~
- ~[ ] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).~


pypsa/linopf.py
```diff
@@ -1070,7 +1070,10 @@ def save_optimal_capacities(n, iteration, status):
     s_nom_orig = n.lines.s_nom.copy()
     p_nom_orig = n.links.p_nom.copy()
     n.lines[['s_nom', 's_nom_extendable']] = n.lines['s_nom_opt'], False
-    n.links[['p_nom', 'p_nom_extendable']] = n.links['p_nom_opt'], False
+    for i in n.links.index:
+      if n.links.carrier == "DC":
+        n.links.loc[i,'p_nom'] = n.links.loc[i,'p_nom_opt']
+        n.links.loc[i,'p_nom_extendable'] = False
```
fneum, line 1076: Great! I would just comment that such loops are quite inefficient. The better way to do it would be:

```suggestion
dc_links_b = n.links.carrier == "DC"
n.links.loc[dc_links_b, "s_nom"] = n.links.loc[dc_links_b, "s_nom_opt"]
n.links.loc[dc_links_b, "s_nom_extendable"] = False
```
@FabianHofmann I was wondering whether we could also lift the capacity fix of links overall? The reason for fixing was just to prevent AC lines from changing impedances.

fneum (changes requested):

chiaroo:
Another thing I noticed about this: The capital costs of lines and links are not reflected in the n.objective of the output network, since they are not prone to optimisation during the last iteration. This can be easily observed when tracking the objective values of the iterations in the log, but also when manually summing up the costs in the output network. 

If this is not internalised by the user (and I couldn't find anything about this in the documentation) and grid expansion is allowed, system costs may be indicated as too low. 
This could be solved by implemented something like this as the end of the linopf script, before exporting the output network:
`obj_links = (n.links[n.links.carrier=="DC"]['p_nom_opt'] - n.links[n.links.carrier=="DC"]['p_nom_min']) * n.links.loc[n.links.carrier=="DC"]['capital_cost']    
obj_lines = (n.lines['s_nom_opt'] - n.lines['s_nom_min']) * n.lines['capital_cost']
n.objective = n.objective + obj_lines.sum() + obj_links.sum()`

pypsa/linopf.py
```diff
@@ -1070,7 +1070,9 @@ def save_optimal_capacities(n, iteration, status):
     s_nom_orig = n.lines.s_nom.copy()
     p_nom_orig = n.links.p_nom.copy()
     n.lines[['s_nom', 's_nom_extendable']] = n.lines['s_nom_opt'], False
-    n.links[['p_nom', 'p_nom_extendable']] = n.links['p_nom_opt'], False
+dc_links_b = n.links.carrier == "DC"
+n.links.loc[dc_links_b, "p_nom"] = n.links.loc[dc_links_b, "p_nom_opt"]
+n.links.loc[dc_links_b, "p_nom_extendable"] = False
```
fneum, line 1075: Requires one more indentation, which is why tests failed.

fneum:
This behaviour is intended. The network attribute `n.objective` only includes the cost of added infrastructure and operation, not the existing infrastructure. The network has an additional attribute `n.objective_constant` which reflects the capital cost of already existing infrastructure.

chiaroo:
Yes, I'm well aware of that, that was not my point. The output network of an ilopf operation is the one from the last iteration, with all grid fixed. Meaning, factually, it does not add any infrastructure anywhere (grid not extendable and set to expansion from earlier iteration), thus,  the actual expansion from earlier iterations, that should be included in the objective function, is not included in the objective function. 
I checked this with an output network of mine, summing up all costs that should occur (capital costs of everything ADDED, not pre-existing + operation), plus, I was wondering why the objective values of earlier iterations displayed in the log are much higher than the final one.

chiaroo:
Bildschirmfoto 2021-05-21 um 16 52 07
Maybe this as a quick demonstration if my issue did not become clear - this is the output of a green field network (thus, no pre-existing grid infrastructure), only vRES as gens + battery & H2 stores components
I summed up the capital costs of generation, of stores and the operation costs of generation; then I added the capital costs of links and lines (these also include the capital costs for electrolysis here, since the fix from post 1 of this pull request hasn't been implemented yet), and the actual objective value.
I must have missed some minor costs somewhere, since 1 and 3 are not exactly equal, but I think the intuition becomes clear

fneum:
Ok, now I understand the point. One aspect I like about the current implementation is that it directly stores the solver's output in `n.objective` and the constant term of the last iteration (can be helpful for debugging). For analysing the network, you would most likely calculate the cost separately split by components etc. anyhow. So I would argue to keep as is (the links should of course still be fixed). Any tie breaking opinions, e.g. @FabianHofmann?

FabianHofmann:
Hey, just from thinking *a bit* about it: 

**Considering the DC Links:** Why are we fixing them in the last iteration in the first place? The non-linearity only comes from the AC-impedances, right?

**Considering the objective value:** @chiaroo you are definitively right, the objective value is not consistent with the pypsa logic (since we're changing switching lines and links to extendable after solving with fixed capacities).  

I see two options: 

1. Do not set `s_nom_extendable`, `p_nom_extendable` to True after solving with fixed capacities
2. Do it like @chiaroo did it.

In general it would be good to have a flag `additional_iteration_with_fixed_capacities` (or something shorter) which allows to skip the last iteration, and directly get the solution of the converged expansion problem.

fneum:
> Why are we fixing them in the last iteration in the first place? The non-linearity only comes from the AC-impedances, right?

Yes, that's what I was thinking. I think it makes sense to not fix any links in the last iteration.

> I see two options:
>    1. Do not set s_nom_extendable, p_nom_extendable to True after solving with fixed capacities
>    2. Do it like @chiaroo did it.

In this case, I prefer 2. The fact that we set `{s,p}_nom_extendable` to True after solving convinced me.

chiaroo:
Hey guys, thanks for the discussion! I agree with your proposals.
I adjusted the code accordingly - all links have been excluded from being fixed, and the n.objective is post-processed by adding the costs for the line extensions.

fneum:
Great, I just had one thought I need to investigate later: If there is a global limit on the volume of transmission expansion (including both links and lines) it will likely be ignored with the proposed implementation (which was probably the original reason why the DC links are also fixed).

chiaroo:
This is funny because these are actually exactly the circumstances how I noticed this. I set a global limit on the store links, and the optimisation always kicked me out when preparing the network for the last iteration with the initial configuration. (To be honest, I don't really understand why, because restricting DC links has always worked well for me.) When excluding them like proposed in the initial commit, everything went through smoothly. 
I haven't checked this for the DC links when completely excluding them from being fixed (and unfortunately I cannot check it right now, because my cluster is under maintenance this week), but I can say that at least for the store links, excluding them worked fine.

fneum:
@chiaroo I took the liberty to directly push onto your patch branch. Given the line volume limit constraint, I think it is advisable to fix the DC links in the last iteration as you initially suggested. Another addition was to subtract the cost added to `n.objective` from `n.objective_constant` if I am not mistaken. Could you check whether the state as is works for you?

chiaroo:
Hey Fabian, thank you so much, the code looks really great! I tested every code snippet manually, since I cannot solve the network remotely. I will re-run it for an actual solving process on saturday when my cluster is back up. Right now, everything seems to work fine.
Just one quick memo - for all my solved network, the `n.objective_constant` is always equal to zero although I have several pre-existing assets in place (grid + hydro capacities; and of course also the costs for the fixed, but expanded branches should be reflected there). This is why I didn't consider it in adjusting the ilopf code. Right now, I don't really have the ressources to investigate why this is the case, I just wonder if this a specific problem of my networks only of if this is a general problem?

fneum (approved):

fneum:
Merging without test as n.ilopf is not yet covered by unit testing.

PR #251, opened 2021-06-17
Example module

FabianHofmann:

Add a new pypsa module "examples". Use is to retrieve/load pypsa example networks. 

```python
```

This is very helpful for developing and for new users to get in touch with the package. It can be easily extended by pypsa-eur and pypsa-eur-sec results, once their netcdf file are made available somewhere (with the current zip files on zenodo, one would have to download the whole set of networks...)  

## Checklist


FabianHofmann:
@fneum could I ask you for a small review whether the general setup is suitable? CI does not want to start....

pypsa/examples.py
```diff
@@ -0,0 +1,101 @@
...
+_data_dir = os.path.join(
+    os.environ.get("XDG_DATA_HOME", os.environ.get("APPDATA", _writable_dir)),
+    "pypsa-examples",
+)
+_data_dir = Path(_data_dir)
+_data_dir.mkdir(exist_ok=True)
+
+
+# TODO: replace ending by 'master/' as soon as merging into master
+_repo_url = "https://github.com/PyPSA/PyPSA/raw/example-module/"
```
fneum, line 30: Even better would be the version that is currently installed, right?

```suggestion
_repo_url = f"https://github.com/PyPSA/PyPSA/raw/v{pypsa.__version__}/"
```

pypsa/examples.py
```diff
@@ -0,0 +1,101 @@
...
+    Load the meshed AC-DC network example of pypsa stored in the PyPSA repository.
+
+    Parameters
+    ----------
+    update : bool, optional
+        Whether to update the locally stored network data. The default is False.
+
+    Returns
+    -------
+    pyspa.Network
```
fneum, line 55:
```suggestion
    pypsa.Network
```
all the examples :)

fneum (commented):
I like the implementation. One aspect to discuss is the use of `.nc` files. I guess it's practical to retrieve only one file. Should we then switch to `.nc` for all the examples (and testing) to avoid the duplication?

pypsa/examples.py
FabianHofmann: ups :-)

pypsa/examples.py
FabianHofmann: on the one hand this is more consistent, on the other you loose the availability to directly retrieve from the master branch. Perhaps a new flag `from_master` could provide a remedy?

FabianHofmann:
> I like the implementation. It's going to be quite useful and I know pandapower has a similar setup. One aspect to discuss is the use of `.nc` files. I guess it's practical to retrieve only one file. Should we then switch to `.nc` for all the examples (and testing) to avoid the duplication?

Yes, i think that would be helpful. But I would do it in another step to have a smoother transition.

fneum:
Yes, should be separated!

pypsa/examples.py
fneum: A `from_master=False` flag sounds good!

FabianHofmann:
alright since the CI won't start off, I'd say we pull this in and test it then

PR #253, opened 2021-06-20
test multi os matrix

fneum:


FabianHofmann:
great let's solve the remaining issue in the merged branch

PR #255, opened 2021-06-25
only test with glpk

fneum:
Drop CBC tests.


fneum:
@FabianHofmann could you have a look? I think this is actually linopt failing.
```
E   TypeError: must be real number, not str

pypsa/linopt.py:369: TypeError
```

FabianHofmann:
I directly pushed the fix here to directly test it

fneum:
So the issue with macos and windows seems to be that there are no pip wheels for `tables` for these combinations of OS and Python 3.9.

https://pypi.org/project/tables/#files

On the other hand there are for `conda`

https://anaconda.org/conda-forge/pytables/files?version=3.6.1

Therefore, I think it's best to remove 3.9 from pip testing and maybe add 3.9 to conda testing? @euronion @FabianHofmann

euronion:
Yes, since we prefer `conda` anyways I think that is a legit approach.

FabianHofmann (approved):

PR #256, opened 2021-06-25
Tests to make more use of pytest functionalities

fneum:


test/test_opf_storage.py
```diff
@@ -1,35 +1,53 @@
...
+    network.lopf(solver_name='glpk', pyomo=True)
+    equal(
+        network.generators_t.p.reindex_like(target_gen_p),
+        target_gen_p,
+        decimal=2
+    )
+
+
+@pytest.mark.skipif(sys.version_info.major < 3,
+                    reason="Optimisation without pyomo requires Python 3.")
```
FabianHofmann, line 45: We only support python 3, right?

FabianHofmann (commented):
Looks good and a lot cleaner.

test/test_opf_storage.py
fneum: That is true.

PR #259, opened 2021-06-30
Multi investment add investment periods attribute

FabianHofmann:

* Add a investment period attribute which serves as a intermediate layer. 
* Add a convenience functions to repeat single-level time-series. 
* Revise multi decade example a bit 

For the convenience case try 
```python
n = pypsa.examples.storage_hvdc(from_master=True)
n.investment_periods = [2020, 2030]
```
Now the snapshots are multi-indexed.

## Checklist

### DIFF
--- a/examples/multi-decade-example.py
+++ b/examples/multi-decade-example.py
@@ -36,8 +36,8 @@ n = pypsa.examples.storage_hvdc()
 print(n.snapshot_weightings)
 
 # snapshots can be set as a multiindex
-# a convenience functionality is provided: when setting the investment periods 
-# directly, the all time-series are repeated for each period. 
+# a convenience functionality is provided: when setting the investment periods
+# directly, the all time-series are repeated for each period.
 single_snapshots = n.snapshots
 n.investment_periods = [2020, 2030]
 print(n.snapshots)
@@ -187,11 +187,11 @@ n.determine_network_topology()
 print(n.buses.sub_network)
 
 # determines network topolgy in first investment period (bus 2 isolated)
-n.determine_network_topology(n.investment_period_weightings.index[0])
+n.determine_network_topology(n.investment_periods[0])
 print(n.buses.sub_network)
 
 # determines network topology in third investment period (all lines are build)
-n.determine_network_topology(n.investment_period_weightings.index[2])
+n.determine_network_topology(n.investment_periods[2])
 print(n.buses.sub_network)
 
 
--- a/examples/multi-decade-example.py
+++ b/examples/multi-decade-example.py
@@ -32,16 +32,17 @@ print(n.snapshots)
 
 # when a pypsa network from an older version is imported, the old snapshot weightings
 # are set row-wise and a warning is given
-n = pypsa.Network("opf-storage-hvdc/opf-storage-data")
+n = pypsa.examples.storage_hvdc()
 print(n.snapshot_weightings)
 
 # snapshots can be set as a multiindex
-# CAUTION time-dependent data is lost from changing single to multiindex
+# a convenience functionality is provided: when setting the investment periods 
+# directly, the all time-series are repeated for each period. 
 single_snapshots = n.snapshots
-snapshots_multiindex = pd.MultiIndex.from_product([[2020, 2030], single_snapshots])
-n.set_snapshots(snapshots_multiindex)
+n.investment_periods = [2020, 2030]
 print(n.snapshots)
 print(n.investment_period_weightings)
+assert all(n.snapshots == pd.MultiIndex.from_product([[2020, 2030], single_snapshots]))
 
 # investment_periods have to be integer and increasing, otherwise there will be an
 # error message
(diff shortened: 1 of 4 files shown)


FabianHofmann:
One question @lisazeyen : Since we have the 'time' attribute in n.investment_period_weightings, is it still necessary that investment period weightings are increasing integers?

lisazeyen:
> One question @lisazeyen : Since we have the 'time' attribute in n.investment_period_weightings, is it still necessary that investment period weightings are increasing integers?

Yes, because otherwise it is not possible to consider assets which have a build_year before the modelling horizon (e.g. a coal plant was built in 1990 with a lifetime of 50 years and modelling horizon starting in 2020.)

FabianHofmann:
mmh okay, but would that not be better implemented as lifetime=20, as this is the effective lifetime within the optimisation horizon? 

The problem I see with the current concept is that we have two measures for the elapsed time per period, one is the difference of integers in `n.investment_periods` and one is the 'time' column in the weightings. I would prefer one.

Two propositions: 
1. Still require integer investment periods to be years (increasing integers). Delete the 'time' column in weightings and get the elapsed time from the integer difference.
2. Do not require integer investment periods (they can be anything, even strings) and control the elapsed time with the 'time' column. Then, the build_year then to be one of the investment periods (nan defaults to the first investment period)  and active assets are calculated the from the elapsed time. After some trying out this would be the corresponding `get_active_assets` functions  
```
def get_active_assets(n, c, investment_period):
    """
    Getter function. Get True values for elements of component c which are active
    at a given investment period. These are calculated from lifetime and build
    year.
    """
    time = n.investment_period_weightings.time.cumsum().shift().fillna(0)
    until_build = n.df(c).build_year.fillna(n.investment_periods[0]).map(time)
    elapsed_time = time[investment_period]

    return (n.df(c).eval("@until_build <= @elapsed_time < @until_build + lifetime")
            .rename(investment_period))
```
For the second case the column 'build_year' would better be renamed as 'build_period' or 'commissioning_period'.

I prefer the second option, since then the multi-investments are more abstract. One could define investment periods as ('first_period', 'second_period', ...) and easily control the elapsed time with the 'time' column. Also not so much double-checks are needed. 
Any thoughts?

examples/multi-decade-example.py
```diff
@@ -32,16 +32,17 @@
...
 print(n.snapshot_weightings)
 
 # snapshots can be set as a multiindex
-# CAUTION time-dependent data is lost from changing single to multiindex
+# a convenience functionality is provided: when setting the investment periods
+# directly, the all time-series are repeated for each period.
 single_snapshots = n.snapshots
-snapshots_multiindex = pd.MultiIndex.from_product([[2020, 2030], single_snapshots])
-n.set_snapshots(snapshots_multiindex)
+n.investment_periods = [2020, 2030]
```
lisazeyen, line 42: Cool!

examples/multi-decade-example.py
```diff
@@ -32,16 +32,17 @@
...
 print(n.snapshots)
 print(n.investment_period_weightings)
+assert all(n.snapshots == pd.MultiIndex.from_product([[2020, 2030], single_snapshots]))
```
lisazeyen, line 45: The subsequent test with non-ascending investment periods no longer generates an error message
`snapshots_multiindex = pd.MultiIndex.from_product([[2030, 2020], single_snapshots])
 n.set_snapshots(snapshots_multiindex)`
It does check when setting the investment periods that they are integer and increasing and if all investment periods are defined in the network snapshots, but it would be good if it also checks the order of the first level index of the snapshots. Otherwise it is possible to have like in this case snapshots with level=0 order [2030,2020] and investment periods with order [2020,2030]

lisazeyen (commented):
This is now much better with the investment_periods! I add later today a test for checking that the order of the first level of the snapshots aligns with the order of the investment periods.
I also would change parts of the `linopf.py` using the n.investment_periods instead of first level of the snapshots.

lisazeyen:
> mmh okay, but would that not be better implemented as lifetime=20, as this is the effective lifetime within the optimisation horizon?
> 
> The problem I see with the current concept is that we have two measures for the elapsed time per period, one is the difference of integers in `n.investment_periods` and one is the 'time' column in the weightings. I would prefer one.
> 
> Two propositions:
> 
> 1. Still require integer investment periods to be years (increasing integers). Delete the 'time' column in weightings and get the elapsed time from the integer difference.
> 2. Do not require integer investment periods (they can be anything, even strings) and control the elapsed time with the 'time' column. Then, the build_year then to be one of the investment periods (nan defaults to the first investment period)  and active assets are calculated the from the elapsed time. After some trying out this would be the corresponding `get_active_assets` functions
> 
> ```
> def get_active_assets(n, c, investment_period):
>     """
>     Getter function. Get True values for elements of component c which are active
>     at a given investment period. These are calculated from lifetime and build
>     year.
>     """
>     time = n.investment_period_weightings.time.cumsum().shift().fillna(0)
>     until_build = n.df(c).build_year.fillna(n.investment_periods[0]).map(time)
>     elapsed_time = time[investment_period]
> 
>     return (n.df(c).eval("@until_build <= @elapsed_time < @until_build + lifetime")
>             .rename(investment_period))
> ```
> 
> For the second case the column 'build_year' would better be renamed as 'build_period' or 'commissioning_period'.
> 
> I prefer the second option, since then the multi-investments are more abstract. One could define investment periods as ('first_period', 'second_period', ...) and easily control the elapsed time with the 'time' column. Also not so much double-checks are needed.
> Any thoughts?

Mhm, I understand that it would be good to have the investment periods a bit more flexible and abstract. The advance of having the investment periods as integers is that you can rerun the same network with different investment periods without always changing the lifetime and build year. And the definition of lifetime and build_year is then aligned with the "myopic" optimisation in the sector-coupling model. The time weightings can be also used to take care of e.g. "end of horizon" effects by weighting the final investment year stronger than the previous once, so I would say we should keep them. If the final year is weighted more strongly, many assets would be non-active in the final investment period. @fneum  any thoughts on this?

examples/multi-decade-example.py
```diff
@@ -32,16 +32,17 @@
 
 # when a pypsa network from an older version is imported, the old snapshot weightings
 # are set row-wise and a warning is given
-n = pypsa.Network("opf-storage-hvdc/opf-storage-data")
+n = pypsa.examples.storage_hvdc()
```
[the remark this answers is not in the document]
FabianHofmann: ah right I forgot, just add a `from_master=True`

FabianHofmann:
> Mhm, I understand that it would be good to have the investment periods a bit more flexible and abstract. The advance of having the investment periods as integers is that you can rerun the same network with different investment periods without always changing the lifetime and build year. And the definition of lifetime and build_year is then aligned with the "myopic" optimisation in the sector-coupling model. The time weightings can also be used to take care of e.g. "end of horizon" effects by weighting the final investment year stronger than the previous once, so I would say we should keep them. If the final year is weighted more strongly, many assets would be non-active in the final investment period. @fneum any thoughts on this?

Good point. It' important to take care of the last period! ...so the 'time' column should stay (which speaks for option 2). 
I also see the benefit of having required integer periods, but I think it is just a bit too fiddly.  
One work-around: we rename `build_year` into `build_period` which indicates the first investment period of the asset. 
Then, given the column 'build_year' (in integers) which is just a dataholder and not used in the optimisation, the following function can be used to get the build_period

```
def determine_build_period(model_start, buildyear_col='build_year'):

    time = n.investment_period_weightings.time.cumsum().shift().fillna(0)

    for c in n.iterate_components(n.one_port_components | n.branch_components):
        if not c.df[buildyear_col].empty:
            c.df.build_period.update((c.df[buildyear_col] - model_start).dropna()
                                     .apply(lambda x: time[time >= x].idxmin()))
```
which maps the build_period to the investment_period which 'follows' the build_year.
For updating/determining the lifetime:
```
def determine_lifetime(model_start, buildyear_col='build_year',
                       lifetime_col='lifetime'):

    time = n.investment_period_weightings.time.cumsum().shift().fillna(0)

    for c in n.iterate_components(n.one_port_components | n.branch_components):
        if not c.df[buildyear_col].empty:
            c.df.lifetime.update(
                (c.df[lifetime_col] + c.df[buildyear_col] - model_start).dropna())

```

FabianHofmann:
In meanwhile: Are you fine with merging this PR? Then we can continue on the upstream branch

PR #264, opened 2021-07-03
add Action to auto-deploy to PyPI

### LINKED ISSUE #254: GH Action to auto-publish releases to PyPI
Once #252 is merged we can automate the release process further:

https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

fneum:
closes #254 

Rather than using twine locally, automate the build & upload of a new release to PyPI with a Github Action. The building process will be tested for each commit. Publish only for tags.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,25 @@
...
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Build distribution
+      run: |
+        python setup.py sdist
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
FabianHofmann, line 19: I am wondering who's username it takes here if we only pass the password? In linopy/atlite we added user name and password as secrets (which are then both passed here)

fneum: Yes, here it's my tokens stored as secret (although I still need to be added as PyPSA maintainer on Test PyPI which is why the CI currently fails)

FabianHofmann: Ah okay, which means that only you and tom can do the tagged releases (which is totally be with me) since it takes the GH username of the one who pushes, right?

fneum: No, I think everyone with push rights to https://github.com/pypsa/pypsa can tag a new release.

PR #289, opened 2021-09-08
introduce hierarchical clustering

martacki:
Introduce hierarchical agglomeraticve clustering method as new function.

Theoretically I could change `affinity` and `linkage` of the functions to optional kwargs & set current fixed ones as default. Current version is the specific case of Ward's method (`linkage="ward"`), and hence compatible only with the euclidean measure (`affinity="euclidean"`).

Locally the function(s) worked fine.

## Checklist

### DIFF
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -356,7 +356,7 @@ def busmap_by_spectral_clustering(network, n_clusters, **kwds):
     A = network.adjacency_matrix(branch_components=["Line", "Link"], weights=weight)
 
     # input arg for spectral clustering must be symmetric, but A is directed. use A+A.T:
-    return pd.Series(sk_spectral_clustering(A+A.T, n_clusters=50).astype(str), index=network.buses.index)
+    return pd.Series(sk_spectral_clustering(A+A.T, n_clusters=n_clusters).astype(str), index=network.buses.index)
 
 @deprecated(deprecated_in="0.19", removed_in="0.20", current_version = pypsa_version,
             details="Use ``kmeans_clustering`` instead.")
@@ -512,6 +512,85 @@ def busmap_by_rectangular_grid(buses, divisions=10):
 def rectangular_grid_clustering(network, divisions):
     busmap = busmap_by_rectangular_grid(network.buses, divisions)
     return get_clustering_from_busmap(network, busmap)
+
+################
+# Hierarchical Clustering
+def busmap_by_hac(network, n_clusters, buses_i=None, branch_components=["Line", "Link"], feature=None):
+    """
+    Create a busmap accroding to Hierarchical Agglomerative Clustering.
+
+    Parameters
+    ----------
+    network : pypsa.Network
+    n_clusters : int
+        Final number of clusters desired.
+    buses_i: None|pandas.Index
+        Subset of buses to cluster. If None, all buses are considered.
+    branch_components: List
+        Subset of all branch_components in the network. Defaults to ["Line", "Link"].
+    feature: None | pandas.DataFrame
+        Feature to be considered for the clustering.
+        The DataFrame must be indexed with buses_i.
+        If None, all buses have the same similarity.
+
+    Returns
+    -------
+    busmap : pandas.Series
+        Mapping of network.buses to k-means clusters (indexed by
+        non-negative integers).
+    """
+
+    if find_spec('sklearn') is None:
+        raise ModuleNotFoundError("Optional dependency 'sklearn' not found."
+            "Install via 'conda install -c conda-forge scikit-learn' "
+            "or 'pip install scikit-learn'")
+
+    from sklearn.cluster import AgglomerativeClustering as HAC
+
+    if buses_i is None:
+        buses_i = network.buses.index
+
+    if feature is None:
+        logger.warning(
+            "No feature is specified for Hierarchical Clustering. "
+            "Falling back to default, where all buses have equal similarity. "
+            "You can specify a feature as pandas.DataFrame indexed with buses_i."
+        )
+
+        feature = pd.DataFrame(index=buses_i, columns=[""], data=0)
+
+    buses_x = network.buses.index.get_indexer(buses_i)
+
+    A = network.adjacency_matrix(branch_components=branch_components).todense()
+    A = sp.sparse.coo_matrix(A[buses_x].T[buses_x].T)
+
+    labels = HAC(n_clusters=n_clusters,
+                 connectivity=A,
+                 affinity='euclidean',
+                 linkage='ward').fit_predict(feature)
+
+    busmap = pd.Series(data=labels, index=buses_i, dtype='str')
+
+    return busmap
+
+def hac_clustering(network, n_clusters, buses_i=None, branch_components=["Line", "Link"], feature=None, line_length_factor=1.0):
+    """
+    Cluster then network using Hierarchical Agglomerative Clustering.
+
+    Parameters
+    ----------
+    network : pypsa.Network
+        The buses must have coordinates x,y.
+
+    Returns
+    -------
+    Clustering : named tuple
+        A named tuple containing network, busmap and linemap
+    """
+
+    busmap = busmap_by_hac(network, n_clusters, buses_i, branch_components, feature)
+
+    return get_clustering_from_busmap(network, busmap, line_length_factor=line_length_factor)
 
 ################
 # Reduce stubs/dead-ends, i.e. nodes with valency 1, iteratively to remove tree-like structures


pypsa/networkclustering.py
```diff
@@ -356,7 +357,7 @@ def busmap_by_spectral_clustering(network, n_clusters, **kwds):
     A = network.adjacency_matrix(branch_components=["Line", "Link"], weights=weight)
 
     # input arg for spectral clustering must be symmetric, but A is directed. use A+A.T:
-    return pd.Series(sk_spectral_clustering(A+A.T, n_clusters=50).astype(str), index=network.buses.index)
+    return pd.Series(sk_spectral_clustering(A+A.T, n_clusters=n_clusters).astype(str), index=network.buses.index)
```
fneum, line 360: Great catch!

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
...
+        buses_i = network.buses.index
+
+    if feature is None:
+        logger.warning(
+            "No feature is specified for Hierarchical Clustering. "
+            "Falling back to default, where all buses have equal similarity. "
+            "You can specify a feature as pandas.DataFrame indexed with buses_i."
+        )
+
+        feature = pd.DataFrame(index=buses_i, columns=[""], data=0)
```
fneum, line 561:
```suggestion
            "You can specify a feature as pandas.Series indexed with buses_i."
        )

        feature = pd.Series(0, index=buses_i)
```

Are multiple features allowed, i.e. can the dataframe have multiple columns? If not, a pd.Series would be more elegant.

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
...
+
+    buses_x = network.buses.index.get_indexer(buses_i)
+
+    A = network.adjacency_matrix(branch_components=branch_components).todense()
+    A = sp.sparse.coo_matrix(A[buses_x].T[buses_x].T)
+
+    labels = HAC(n_clusters=n_clusters,
+                 connectivity=A,
+                 affinity='euclidean',
+                 linkage='ward').fit_predict(feature)
```
fneum, line 571: I would advocate for promoting `affinity` and `linkage` to kwargs.

To catch incompatible combinations, you can do something like:

```py
if linkage == 'ward':
    assert affinity == 'euclidian', "For Ward's method affinitiy needs to be euclidian."
```

If you do this, you could also add a selection of values that can be selected for affinity and linkage.

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
     busmap = busmap_by_rectangular_grid(network.buses, divisions)
     return get_clustering_from_busmap(network, busmap)
 
+################
+# Hierarchical Clustering
+def busmap_by_hac(network, n_clusters, buses_i=None, branch_components=["Line", "Link"], feature=None):
+    """
+    Create a busmap accroding to Hierarchical Agglomerative Clustering.
```
fneum, line 521:
```suggestion
    Create a busmap according to Hierarchical Agglomerative Clustering.
```

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
...
+            "Falling back to default, where all buses have equal similarity. "
+            "You can specify a feature as pandas.DataFrame indexed with buses_i."
+        )
+
+        feature = pd.DataFrame(index=buses_i, columns=[""], data=0)
+
+    buses_x = network.buses.index.get_indexer(buses_i)
+
+    A = network.adjacency_matrix(branch_components=branch_components).todense()
+    A = sp.sparse.coo_matrix(A[buses_x].T[buses_x].T)
```
fneum, line 566: Do you need to convert to dense before you reconvert to sparse? Should be more efficient without, but don't know if possible.

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
...
+
+    labels = HAC(n_clusters=n_clusters,
+                 connectivity=A,
+                 affinity='euclidean',
+                 linkage='ward').fit_predict(feature)
+
+    busmap = pd.Series(data=labels, index=buses_i, dtype='str')
```
fneum, line 573:
```suggestion
    busmap = pd.Series(labels, index=buses_i, dtype=str)
```

pypsa/networkclustering.py
```diff
@@ -513,6 +514,85 @@ def rectangular_grid_clustering(network, divisions):
...
+
+    return busmap
+
+def hac_clustering(network, n_clusters, buses_i=None, branch_components=["Line", "Link"], feature=None, line_length_factor=1.0):
+    """
+    Cluster then network using Hierarchical Agglomerative Clustering.
```
fneum, line 579:
```suggestion
    Cluster the network using Hierarchical Agglomerative Clustering.
```

fneum (changes requested):
Very nice! Just a couple of minor suggestions/comments.

pypsa/networkclustering.py
martacki: would be, but does not work... and multiple columns are allowed. Throws the following error:

```
Expected 2D array, got 1D array instead:
array=[0 0 0 ... 0 0 0].
Reshape your data either using array.reshape(-1, 1) if your data has a single feature or array.reshape(1, -1) if it contains a single sample.
```

pypsa/networkclustering.py
martacki: unfortunately, I couldn't find a workaround... help wanted! :)

best would be, if pypsas adjacency_matrix had the `buses_i` as kwarg... maybe I can do something using `busorder`?

pypsa/networkclustering.py
martacki: ok, thanks for the feedback! I think I also prefer doing it that way; but if'll leave out the assertion error. this is already covered in hac itself :)

fneum: ok, sounds good!

pypsa/networkclustering.py
FabianHofmann:
```suggestion
    A = network.adjacency_matrix(branch_components=branch_components).tocsc()[buses_x, buses_x]
```
what about this? Compressed Sparse Column format allows indexing. And matrix multiplications should work.

fneum: hmm, don't know a fix right now

I had thought about adding a kwarg to the adjacency_matrix function `as_pandas=False` that returns the adjacency matrix as pandas dataframe (similar for the other matrices of a network). Maybe that would help here?

martacki: ahh, thanks. but does not work that way, we need 

```suggestion
    A = network.adjacency_matrix(branch_components=branch_components).tocsc()[buses_x]
```

FabianHofmann: alternatively, pandas also supports sparse arrays which you can index by labels (as normal pandas data frames) 
```
A = pd.DataFrame.sparse.from_spmatrix(A, n.buses.index, n.buses.index).loc[buses_i, buses_i]
```

FabianHofmann: > ahh, thanks. but does not work that way, we need

Ah, probably because of the tuple indexing (I updated the first suggestion)

martacki:
i don't like the repetition of kwarg descriptions in the busmap_by_hac and hac_clustering, similar to busmap_by_kmeans and kmeans_clustering... is there a sparse way to do this?

FabianHofmann (commented):
@martacki does the `.fit_predict` function allow to pass more than one column? otherwise we could change the feature argument to be a series.

martacki:
> @martacki does the `.fit_predict` function allow to pass more than one column? otherwise we could change the feature argument to be a series.

Multiple columns can be passed, yes- plus, passing a pd.Series throws an error.

FabianHofmann (approved):

pypsa/networkclustering.py
```diff
@@ -508,11 +503,132 @@ def busmap_by_rectangular_grid(buses, divisions=10):
...
+        The linkage criterion determines which distance to use between sets of observation.
+        The algorithm will merge the pairs of cluster that minimize this criterion.
+        - ‘ward’ minimizes the variance of the clusters being merged.
+        - ‘average’ uses the average of the distances of each observation of the two sets.
+        - ‘complete’ or ‘maximum’ linkage uses the maximum distances between all observations of the two sets.
+        - ‘single’ uses the minimum of the distances between all observations of the two sets.
+    line_length_factor: float, default=1.0
+        Factor to multiply the spherical distance between two new buses in order to get new line lengths.
+    kwargs:
+        Any remaining arguments to be passed to Hierarchical Clustering (e.g. memory, connectivity).
```
FabianHofmann, line 619: Ah, just saw that the kwargs are not passed to the HAC function. Also, perhaps it would be more clever to set the the default of `branch_components` to `n.branch_components`?

martacki: @FabianHofmann thanks!

fneum:
🎉

PR #308, opened 2021-11-03
add highs solver

pz-max:
Closes # (if applicable).

Add open source [HiGHS](https://github.com/ERGO-Code/HiGHS) solver:
solver_performance_new

Future work can:
- While the above results look incredible, generally all open source solvers struggle when 1 year is simulated in hourly resolution. I.e. Gurobi solves 5 nodes for 1 year in hourly resolution in 1min, while HiGHS and others may need 2+ hours. This performance drop occurs because open source solvers don't address properly the existence of "dense columns".
- Future work can contribute to the HiGHS solver and implement a method for the IPM that does deal with dense columns.
- Other work could implement (advanced) aggressive scaling techniques into PyPSA (to be useful for all solvers) and/or HiGHS. See some ongoing PyPSA discussion here: https://github.com/PyPSA/PyPSA/issues/309

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -16,7 +16,9 @@ Upcoming Release
 
 PyPSA 0.18.1 (15th October 2021)
 ================================
-* Add "highs" solver.
+* Add open source "HiGHS" solver: https://github.com/ERGO-Code/HiGHS. See
+  simple performance analysis for HiGHS, cbc, glpk and gurobi here:
+  https://github.com/PyPSA/PyPSA/pull/308#issue-772907717.
 
 * Add assert: CBC solver does not work with '>' and '<'
 
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -13,6 +13,7 @@ Upcoming Release
 
 PyPSA 0.18.1 (15th October 2021)
 ================================
+* Add "highs" solver.
 
 * Compatibility with ``pyomo>=6.1``.
 
--- a/pypsa/linopt.py
+++ b/pypsa/linopt.py
@@ -603,7 +603,7 @@ def set_int_index(ser):
 
 
 def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
-                        solver_options, warmstart=None, store_basis=True):
+                        solver_options={}, warmstart=None, store_basis=True):
     """
    Highs solver function. Reads a linear problem file and passes it to the highs
     solver. If the solution is feasible the function returns the objective,
@@ -684,6 +684,7 @@ def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
     constraints_dual : series
     objective : float
     """
+    logger.warning("The HiGHS solver can potentially solve towards variables that slightly deviate from Gurobi,cbc,glpk")
     options_fn = "highs_options.txt"
     default_dict = {
         "method": "ipm",
--- a/pypsa/linopt.py
+++ b/pypsa/linopt.py
@@ -615,7 +615,7 @@ def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
     After the "make" run, the 'highs' executables are in highs/build/bin/run/highs
     Once the "ctest" runs make sure you set the path. Path setting notes:
     >>> highs lib folder must be in "LD_LIBRARY_PATH" environment variable
-    >>> To do this, insert the below path at .bashrc (which is hidden in the home folder)
+    >>> To do this, insert the below path in the .bashrc (hidden in the home folder in Linux/Ubuntu)
     >>> LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/<your_path>/HiGHS/build/lib
 
     Architecture
@@ -626,16 +626,15 @@ def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
     your local terminal helps to identify the raw outputs that are useful for
     developing the interface further.
 
-    All functions below the "process = ..." do only read and saves the outputs
-    generated from the HiGHS solver. These parts needs are solver specific and
+    All functions below the "process = ..." do only read and save the outputs
+    generated from the HiGHS solver. These parts are solver specific and
     depends on the solver output.
 
     Solver options
     ---------------
-    Solver options are read by the 1) command window and the 2) option_file.
+    Solver options are read by the 1) command window and the 2) option_file.txt
 
-    1) Full list of solver options (executable by command window) is given here:
-    https://www.maths.ed.ac.uk/hall/HiGHS/HighsOptions.set
+    1) An example list of solver options executable by the command window is given here:
     Examples:
     --model_file arg 	File of model to solve.
     --presolve arg 	    Presolve: "choose" by default - "on"/"off" are alternatives.
@@ -645,10 +644,22 @@ def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
     --options_file arg 	File containing HiGHS options.
     -h, --help 	        Print help.
 
-    2) The options_file is directly written in this function. Note, if you want
-    to make adjustments you need to change the pypsa library in the environment
-    of interest (for instance pypsa in the pypsa-eur needs to be changed in case
-    you want to model smt. with pypsa-eur)
+    2) The options_file.txt gives some more options, see a full list here: 
+    https://www.maths.ed.ac.uk/hall/HiGHS/HighsOptions.set 
+    By default, we insert a couple of options for the ipm solver. The dictionary
+    can be overwritten by simply giving the new values. For instance in PyPSA-Eur,
+    you could write a dictionary replacing some of the default values or adding new
+    options:
+    ```
+    solving:
+        solver:
+            name: highs,
+            method: ipm,
+            parallel: "on",
+            <option_name>: <value>,
+    ```
+    Note, the <option_name> and <value> must be equivalent to the name convention
+    of HiGHS.
 
     Output
     ------
@@ -679,7 +690,7 @@ def run_and_read_highs(n, problem_fn, solution_fn, solver_logfile,
     # update default_dict by solver_dic (i.e. from PyPSA-Eur config.yaml)
     default_dict.update(solver_options)
     method = str(default_dict.pop("method", "ipm"))
-    logger.info(f"Options: \"{default_dict}\". More at: https://www.maths.ed.ac.uk/hall/HiGHS/HighsOptions.set")
+    logger.info(f"Options: \"{default_dict}\". Docstring of function explains how to add/change options.")
     f1 = open("highs_options.txt","w")
     # write dict to a text file with options in each row
     f1.write(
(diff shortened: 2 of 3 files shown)


pz-max:
Will add some other features, like using threads in parallel. Just going through tests.

pypsa/linopt.py
```diff
@@ -601,6 +601,142 @@ def set_int_index(ser):
...
+    variables_sol : series
+    constraints_dual : series
+    objective : float
+    """
+    import logging, re, io, subprocess, sys, os
+
+    # write option file
+    if not os.path.exists(solution_fn):
+        os.mknod(solution_fn)
+    f1 = open("highs_options.txt","w")
```
FabianHofmann, line 668: this should probably be a temporary file

FabianHofmann (dismissed):
This is really nice! 
I think the solver options could be handled in a better way. A suggestion:

1. pop all the command line options from the `solver_options` dict with defaults. In the form `parallel = solver_options.pop(parallel, False)`. All of these are added to the command line (later)
2. Set a defaults dict for the `option_file_options` which include the options which you have in line 670-674. 
3. Update the `option_file_options` dict with the remaining `solver_options`, using `option_file_options.update(solver_options)` 
4. Convert the `option_file_options` to text form and write it to the file. 

In this way, one can easily access command line options and option-file options. What do you think?

FabianHofmann (commented):
That looks good! Let me go one time through the installation instruction and run a optimization. From the code everything looks fine. 

Would you mind if I copy some snippets to the linopy package? Otherwise you could just copy/paste that function into the linopy/solvers.py (so it stays effectively your contribution) and make a PR and the rest I take care of.

pz-max:
HI @FabianHofmann, We are testing the installation today with @fneum. Yeah, I would write a PR in Linopy as well. Probably good to have some overview on the codebase anyways :)

pypsa/linopt.py
```diff
@@ -601,6 +601,165 @@ def set_int_index(ser):
...
+    solution and dual constraint variables. Highs must be installed for usage.
+
+    Installation
+    -------------
+    Installation manual: https://www.maths.ed.ac.uk/hall/HiGHS/
+    After the "make" run, the 'highs' executables are in highs/build/bin/run/highs
+    Once the "ctest" runs make sure you set the path. Path setting notes:
+    >>> highs lib folder must be in "LD_LIBRARY_PATH" environment variable
+    >>> To do this, insert the below path in the .bashrc (hidden in the home folder in Linux/Ubuntu)
+    >>> LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/<your_path>/HiGHS/build/lib
```
fneum, line 619: also the HiGHS executables need to be added to the path. Here are the full installation steps I needed.

```bash
sudo apt-get install cmake # if not installed
git clone git@github.com:ERGO-Code/HiGHS.git
cd HiGHS
mkdir build
cd build
cmake ..
make
ctest
```

Then in `.bashrc` add paths of executables and library:

```sh
export PATH="${PATH}:/foo/HiGHS/build/bin"
export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/foo/HiGHS/build/lib"
```

```sh
source .bashrc
```

Now you should see when running `highs`:

```
Running HiGHS 1.1.1 [date: 2021-11-11, git hash: 95342daa]
```

pz-max:
some to do's https://github.com/PyPSA/linopy/pull/8#issuecomment-966485362

pz-max:
Hi guys,
- I installed the new HiGHs version
- Switched the options to `"write_solution_style": 1,` where output variables good be better read out
- Tested with PyPSA
- Tested with PyPSA-Eur

Some outputs for the PyPSA run:
Solver outputs for variable primal:
comparison_solvers1

Solver outputs for constraints dual:
constraint_dual

The difference for the PyPSA-Eur run compared to Gurobi are also minor. So seems now everything works.

fneum:
Thanks @pz-max for driving this! I would like to also see that the variables are correctly attributed to the same variables. The stats somewhat show that all the values of the variables are there but not whether they are correctly allocated.

pz-max:
Luckily I was taking this snapshot:
image

The new `"write_solution_style": 1,` lead to an output with variable naming such as x300 = ... as the other solvers do. I had to use also `.pipe(set_int_index)` for the variables to associate them right

pz-max:
@fneum, here a more complete version:
[solution_variables.csv](https://github.com/PyPSA/PyPSA/files/7547208/solution_variables.csv)
Note, the index had before cleaning an "x" in front of the number i.e. "x300", meaning that the variables should be matched now.

fneum:
What are the column labels for the .csv? Are the objective values the same? Looking mostly good, but could you investigate whether the differences are degeneracies?

pz-max:
Hi @fneum, Thanks for that note. Now all solutions with column index:
- [objective.csv](https://github.com/PyPSA/PyPSA/files/7561668/objective.csv)
- [constraint_dual.csv](https://github.com/PyPSA/PyPSA/files/7561666/constraint_dual.csv)
- [solution_variables.csv](https://github.com/PyPSA/PyPSA/files/7561669/solution_variables.csv)

I checked the outcomes of the results for 3 runs (degeneracy check I assume?). We can observe always the same difference/results. The differences of the unique results are given here:
image
image

Overall the differences are not large. Turning off the crossover, didn't change anything (would need to check the C code if anything was triggered). What are your thoughts? Should we find out how to reduce the little differences or is it already ok?

fneum:
The comparison looks almost all very good! The only thing that bothers me is the 1250 difference from variable x270 I presume. Could you check what variable this is and what cost is associated with it? If there is no cost associated with this variable it's fine.

pypsa/linopt.py
```diff
@@ -601,6 +601,175 @@ def set_int_index(ser):
...
+    you could write a dictionary replacing some of the default values or adding new
+    options:
+    ```
+    solving:
+        solver:
+            name: highs,
+            method: ipm,
+            parallel: "on",
+            <option_name>: <value>,
+    ```
```
fneum, line 673: This explanation should not refer to PyPSA-Eur. Write how you can use the `solver_options` keyword.

```py
solver_options = {
    "method": "ipm",
    ...
}
```

pypsa/linopt.py
```diff
@@ -601,6 +601,175 @@ def set_int_index(ser):
...
+    Note, the <option_name> and <value> must be equivalent to the name convention
+    of HiGHS. Some function exist that are not documented, check their GitHub file:
+    \HighsOptions.h
```
fneum, line 676: Can you provide full link?

pypsa/linopt.py
```diff
@@ -601,6 +601,175 @@ def set_int_index(ser):
...
+    constraints_dual : series
+    objective : float
+    """
+    options_fn = "highs_options.txt"
+    default_dict = {
+        "method": "ipm",
+        "primal_feasibility_tolerance": 1e-04,
+        "dual_feasibility_tolerance": 1e-05,
+        "ipm_optimality_tolerance": 1e-6,
+        # "ipm_iteration_limit": 100,
```
fneum, line 694: Remove if commented out.

pypsa/linopt.py
```diff
@@ -601,6 +601,175 @@ def set_int_index(ser):
...
+    # read out solution file (.sol)
+    f = open(solution_fn, "rb")
+    trimed_sol_fn = re.sub(rb'\*\*\s+', b'', f.read())
+    f.close()
+    
+    sol = pd.read_csv(io.BytesIO(trimed_sol_fn), header=[1], sep=r'\s+')
+    row_no = sol[sol["Index"] == 'Rows'].index[0]
+    sol = sol.drop(row_no+1)  # Removes header line after "Rows"
+    sol_rows = sol[(sol.index > row_no)]
+    sol_cols = sol[(sol.index < row_no)].set_index("Name").pipe(set_int_index)
```
fneum, line 764: Wouldn't it be possible to use `set_int_index` as well for `sol_rows`. Then you wouldn't need the `reset_index(drop=True) + 1` below.

fneum:
I think these are my final comments, after that, good to go in.

doc/release_notes.rst
```diff
@@ -16,6 +16,7 @@ Upcoming Release
 
 PyPSA 0.18.1 (15th October 2021)
 ================================
+* Add "highs" solver.
```
fneum, line 19: Let's add a link.

pz-max:
Hi @fneum,
https://github.com/PyPSA/PyPSA/pull/308#issuecomment-974151807. Attached the lp file:
[lpfile_ac_dc_meshed_example.txt](https://github.com/PyPSA/PyPSA/files/7574902/lpfile_ac_dc_meshed_example.txt)

Attached the .lp file for the above solutions. I analyzed here the pypsa ac_dc example mesh. From the .lp file as well as from .solution_primal_variable file above, we can see that there is no value assigned to **x270**. I guess this means x270 = 0 (never read out variables, hopefully I interpret that correctly).

pypsa/linopt.py
pz-max: This does not work. Below is the dataframe of `sol_rows`. While sol_cols have variables with a "x" i.e. x300, `sol_rows` doesn't have that. Therefore, set_int_index does not work (could be modified).

image

Linopy will be added soon, I think it's probably not useful to modify the set_int_index function?

pypsa/linopt.py
pz-max: done

pypsa/linopt.py
pz-max: done in https://github.com/PyPSA/PyPSA/pull/308/commits/9becfe64b6a12c6d1f9640d923b431192b2073d4

pypsa/linopt.py
fneum: That sounds absolutely reasonable. Thought the duals may have a "y" prefix.

fneum:
> Attached the .lp file for the above solutions. I analyzed here the pypsa ac_dc example mesh. From the .lp file as well as from .solution_primal_variable file above, we can see that there is no value assigned to x270. I guess this means x270 = 0 (never read out variables, hopefully I interpret that correctly).

The variable x270 does show up in both LP file and solution. For gurobi, CBC it's 50, whereas for highs it's 1200. That's curious. What variable is it? Investment variable of a generator or link?

pz-max:
Hi @fneum, I was looking for it. How/where can I check this out?

pz-max:
> Hi @fneum, I was looking for it. How/where can I check this out?

**Update.** When `network_lopf` is run with `keep_references=True` as described [here](https://pypsa.readthedocs.io/en/latest/api_reference.html?highlight=network_lopf#pypsa.linopf.network_lopf) then the variables/constraints can be checked. We found out that in the example above the problematic variable x=270 is related to the state_of_charge variable. 

image

While Gurobi, Cbc and Glpk are solving towards the same variable (x270=100), only HiGHs slightly deviate (x270=1250). This suggests that something is going wrong in the HiGHS solver. The reasons are unclear, therefore a **logger.warning** will be added. Nevertheless, all implemented solvers are very accurate regarding the objective solution (objective=14337113. €) with no deviation before the comma.

fneum:
Congratulations on this very nice contribution! Merging whenever you are ready.

fneum (approved):

PR #315, opened 2021-11-25
Component index name followup

FabianHofmann:

* The snapshot levels of a multi-indexed snapshot were renamed to ['period', 'timestep'], the name of the index was set to 'snapshot'. This makes the snapshot name coherent for single and multi-indexed snapshots.

* A new convenience function `Network.get_committable_i` was added. This returns an index containing all committable assets of component `c`. In case that component `c` does not support committable assets, it returns an empty dataframe.    

* This also fixes the naming of the test functions io test_io.py. Before, two functions were skipped in the test due to repeated function names. 

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -11,6 +11,9 @@ Upcoming Release
 
 * Hierarchical Agglomerative Clustering (HAC) was introduced as new spatial clustering method [`#289 <https://github.com/PyPSA/PyPSA/pull/289>`_].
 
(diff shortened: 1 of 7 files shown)


FabianHofmann:
@lisazeyen could you have a small look? This is mostly affecting the multi-inv optimization due to renaming of the snapshot levels. This is necessary for the linopy integration, which need coherent index names.

lisazeyen (approved):
Looks all good to me!

PR #316, opened 2021-11-28
Minor hac clustering fixes

nmartensen:
Thank you for adding the new HAC clustering method!

Noticed two minor issues:
- The docstring of the busmap_by_hac function probably should not mention k-means in the description of the return value.
- Hardcoded branch_components instead of using the kwarg.


fneum:
Yes, both fixes look correct. @martacki do you agree? Thanks @nmartensen for spotting this!

martacki:
Yup, thanks for spotting!

martacki (approved):

PR #323, opened 2021-12-10
Fix some line parameter calculation issues in network clustering

YifanLi86:
**Purpose**

Observed and fix a few issues when testing network clustering module using BES model so that the resulting network can run lopf right after clustering has been performed.

**What the code is doing**

Update the network components attributes for lines so that r, x will not be missing in aggregatelines function.
As length is an optional input when type is not set, populate the orginal network with length information if it is not there so that length factors can be properly cauculated
Testing for input branches with 0 length but non-zero paramters, which could exist under certain conditions. Also make it forgiving to avoid 0-dividing when calculating parameters aggregatelines.
Carry forward global constraints after clustering.

**Testing**

Manual testing

**Time estimate**

10-20 minutes. Should be fairly straightforward.

### DIFF
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -180,14 +180,8 @@ def aggregatelines(network, buses, interlines, line_length_factor=1.0):
         v_nom_s = buses.loc[list(l.name),'v_nom'].max()
 
         voltage_factor = (np.asarray(network.buses.loc[l.bus0,'v_nom'])/v_nom_s)**2
-        
-        # Could be lines (not transformer) that is 0 distance before and after clustering, but also with non-0 original r, x, g, b values, due to GIS resolution or other reasons.
-        if np.asarray(l['length']).min() == 0 and np.asarray(l['x']).prod() != 0:
-            length_factor = 1
-        else:
-            length_factor = (length_s/l['length'])
-        
-        # length_factor = (length_s/l['length'])
+        non_zero_len = l.length != 0
+        length_factor = (length_s[non_zero_len]/l.length[non_zero_len]).reindex_like(l, fill_value=1)
 
         data = dict(
             r=1./(voltage_factor/(length_factor * l['r'])).sum(),
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -255,8 +255,7 @@ def get_clustering_from_busmap(network, busmap, with_time=True, line_length_fact
     io.import_components_from_dataframe(network_c, lines, "Line")
 
     # Carry forward global constraints to clustered network.
-    if global_constraints:
-        network_c.global_constraints = network.global_constraints
+    network_c.global_constraints = network.global_constraints
     
     if with_time:
         network_c.set_snapshots(network.snapshots)
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -147,12 +147,20 @@ def aggregatebuses(network, busmap, custom_strategies=dict()):
 
 def aggregatelines(network, buses, interlines, line_length_factor=1.0):
 
+    # Make sure network component contents start with "Input" for r and x to avoid missing data in the clustered network before call aggregatelines.
+    network.components["Line"]["attrs"].at['x','status']='Input (required)'
+    network.components["Line"]["attrs"].at['r','status']='Input (required)'
+  
     #make sure all lines have same bus ordering
     positive_order = interlines.bus0_s < interlines.bus1_s
     interlines_p = interlines[positive_order]
     interlines_n = interlines[~ positive_order].rename(columns={"bus0_s":"bus1_s", "bus1_s":"bus0_s"})
     interlines_c = pd.concat((interlines_p,interlines_n), sort=False)
 
+    # line length is an optional input when type is not set so original network may or may not contain this information. If not, line length factor below won't work.
+    if network.lines.length.sum() == 0:
+        interlines_c['length']=haversine_pts(network.buses.loc[interlines_c.bus0, ['x', 'y']], network.buses.loc[interlines_c.bus1, ['x', 'y']])
+    
     attrs = network.components["Line"]["attrs"]
     columns = set(attrs.index[attrs.static & attrs.status.str.startswith('Input')]).difference(('name', 'bus0', 'bus1'))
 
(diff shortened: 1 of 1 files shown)


pypsa/networkclustering.py
```diff
@@ -225,7 +240,7 @@ def get_buses_linemap_and_lines(network, busmap, line_length_factor=1.0, bus_str
 Clustering = namedtuple('Clustering', ['network', 'busmap', 'linemap',
                                        'linemap_positive', 'linemap_negative'])
 
-def get_clustering_from_busmap(network, busmap, with_time=True, line_length_factor=1.0,
+def get_clustering_from_busmap(network, busmap, with_time=True, global_constraints=True, line_length_factor=1.0,
```
FabianHofmann, line 243:
```suggestion
def get_clustering_from_busmap(network, busmap, with_time=True, line_length_factor=1.0,
```

pypsa/networkclustering.py
```diff
@@ -239,6 +254,10 @@ def get_clustering_from_busmap(network, busmap, with_time=True, line_length_fact
     io.import_components_from_dataframe(network_c, buses, "Bus")
     io.import_components_from_dataframe(network_c, lines, "Line")
 
+    # Carry forward global constraints to clustered network.
+    if global_constraints:
+        network_c.global_constraints = network.global_constraints
+
```
FabianHofmann, line 260:
```suggestion
    # Carry forward global constraints to clustered network.
    network_c.global_constraints = network.global_constraints
    
```
If think it make more sense to carry along global constraint in any cases.

pypsa/networkclustering.py
```diff
@@ -147,12 +147,20 @@ def aggregatebuses(network, busmap, custom_strategies=dict()):
 
 def aggregatelines(network, buses, interlines, line_length_factor=1.0):
 
+    # Make sure network component contents start with "Input" for r and x to avoid missing data in the clustered network before call aggregatelines.
+    network.components["Line"]["attrs"].at['x','status']='Input (required)'
+    network.components["Line"]["attrs"].at['r','status']='Input (required)'
+
```
FabianHofmann, line 153: could you point out why this is necessary?

pypsa/networkclustering.py
```diff
@@ -147,12 +147,20 @@ def aggregatebuses(network, busmap, custom_strategies=dict()):
...
     #make sure all lines have same bus ordering
     positive_order = interlines.bus0_s < interlines.bus1_s
     interlines_p = interlines[positive_order]
     interlines_n = interlines[~ positive_order].rename(columns={"bus0_s":"bus1_s", "bus1_s":"bus0_s"})
     interlines_c = pd.concat((interlines_p,interlines_n), sort=False)
 
+    # line length is an optional input when type is not set so original network may or may not contain this information. If not, line length factor below won't work.
+    if network.lines.length.sum() == 0:
+        interlines_c['length']=haversine_pts(network.buses.loc[interlines_c.bus0, ['x', 'y']], network.buses.loc[interlines_c.bus1, ['x', 'y']])
+
```
FabianHofmann, line 163: I don't think we should have this here. Now we are adding information in the cluster routing which was not there before. I rather suggest we move that feature to `n.calculate_dependent_values()`. What do you think @martacki?

FabianHofmann (changes requested):
@YifanLi86 thanks a lot for the contribution! Some open points

pypsa/networkclustering.py
YifanLi86: It looks like in the network, line component attribute text for r and x do not start with "Input", so line 165 "columns = set(attrs.index[attrs.static & attrs.status.str.startswith('Input')]).difference(('name', 'bus0', 'bus1'))" could not select r and x? Maybe this is already been fixed in newer version of Pypsa? I guess this may not be the best location to fix here anyways...

pypsa/networkclustering.py
```diff
@@ -172,7 +180,14 @@ def aggregatelinegroup(l):
...
         voltage_factor = (np.asarray(network.buses.loc[l.bus0,'v_nom'])/v_nom_s)**2
-        length_factor = (length_s/l['length'])
+        
+        # Could be lines (not transformer) that is 0 distance before and after clustering, but also with non-0 original r, x, g, b values, due to GIS resolution or other reasons.
+        if np.asarray(l['length']).min() == 0 and np.asarray(l['x']).prod() != 0:
+            length_factor = 1
+        else:
+            length_factor = (length_s/l['length'])
+        
+        # length_factor = (length_s/l['length'])
```
FabianHofmann, line 190: I am still wondering about this one. Shouldn't we make sure that only where the length is 0, the length_factor is set to 1? Like the following
```suggestion
    length_factor = (length_s/l['length']).where(l['length'] != 0, 1)
```

pypsa/networkclustering.py
FabianHofmann: ah wow, I didn't even know this was used here. Okay, but it actually seems that on the current master it is there.

pypsa/networkclustering.py
YifanLi86: Much more elegant! I was trying to rule out the possibility that there are branches with both 0 length and 0 parameters in original data. In this case, shall we somehow report out the error? That was my original thought. Agree that it is not the best way to let it give warnings on "dividing 0" to give error messages, maybe you would consider add a logic for data sanity check here?

Also, does l['length'] != 0 means every element in l['length'] should not be zero? I am not exactly sure but if so then all good. Just trying to not miss any 0 entries.

FabianHofmann: You are right about the divide by zero warning! Let me make a better suggestion.

Indeed the l['length'] != 0 returns a series where every element is checked. So with the suggested change, we fill up ones everywhere where the length of a line is zero.

pypsa/networkclustering.py
FabianHofmann, line 153:
```suggestion
```
I think we can delete this one

pypsa/networkclustering.py
```diff
@@ -153,6 +153,10 @@ def aggregatelines(network, buses, interlines, line_length_factor=1.0):
     interlines_n = interlines[~ positive_order].rename(columns={"bus0_s":"bus1_s", "bus1_s":"bus0_s"})
     interlines_c = pd.concat((interlines_p,interlines_n), sort=False)
 
+    # line length is an optional input when type is not set so original network may or may not contain this information. If not, line length factor below won't work.
+    if network.lines.length.sum() == 0:
+        interlines_c['length']=haversine_pts(network.buses.loc[interlines_c.bus0, ['x', 'y']], network.buses.loc[interlines_c.bus1, ['x', 'y']])
+
```
FabianHofmann, line 159:
```suggestion
```
Should be tackled in other PR, see #337

FabianHofmann (approved):

FabianHofmann:
Alright, that looks good, thanks @YifanLi86 for your contribution!

PR #329, opened 2021-12-19
Minor plotting

nmartensen:
The main change in this PR is to allow setting `color_geomap=False` in network.plot() to disable drawing of geographical features altogether. The two other commits slightly rearrange the logic in network.plot() and add an assertion.

The motivation for this PR is that so far there was no easy way to disable drawing coastlines and borders when using geographical plots. With this PR applied, you can now use cartopy to draw a map as you like, then pass the Axes to network.plot() to add the buses and branches.

### DIFF
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -137,11 +137,11 @@ def plot(n, margin=None, ax=None, geomap=True, projection=None,
     if boundaries is None and margin:
         boundaries = sum(zip(*compute_bbox_with_margins(margin, x, y)), ())
 
-    if geomap:
-        if not cartopy_present:
-            logger.warning("Cartopy needs to be installed to use `geomap=True`.")
-            geomap = False
+    if geomap and not cartopy_present:
+        logger.warning("Cartopy needs to be installed to use `geomap=True`.")
+        geomap = False
 
+    if geomap:
         if projection is None:
             projection = get_projection_from_crs(n.srid)
 
--- a/pypsa/plot.py
+++ b/pypsa/plot.py
@@ -386,15 +386,18 @@ def draw_map_cartopy(ax, geomap=True, color_geomap=None):
     assert resolution in ['10m', '50m', '110m'], (
             "Resolution has to be one of '10m', '50m', '110m'")
 
-    if color_geomap is None:
-        color_geomap = {'ocean': 'w', 'land': 'w'}
-    elif color_geomap and not isinstance(color_geomap, dict):
+    if not color_geomap:
+        color_geomap = {}
+    elif not isinstance(color_geomap, dict):
         color_geomap = {'ocean': 'lightblue', 'land': 'whitesmoke'}
 
-    ax.add_feature(cartopy.feature.LAND.with_scale(resolution),
-                    facecolor=color_geomap['land'])
-    ax.add_feature(cartopy.feature.OCEAN.with_scale(resolution),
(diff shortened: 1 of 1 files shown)


fneum:
Yes, that's currently a bit inflexible. I have a suggestion for a slightly different implementation such that `color_geomap` is still only used for the fill colours of ocean and land.

How about adding `plot(color_geomap={}, geomap_features=['borders', 'coastlines'])` to the function arguments that is then passed to a modified version of `draw_map_cartopy`?:

```py
def draw_map_cartopy(n, x, y, ax, geomap=True, color_geomap={}, geomap_features=[]):

    resolution = '50m' if isinstance(geomap, bool) else geomap
    assert resolution in ['10m', '50m', '110m'], (
            "Resolution has to be one of '10m', '50m', '110m'")
    axis_transformation = get_projection_from_crs(n.srid)

    if color_geomap and not isinstance(color_geomap, dict):
        color_geomap = {'ocean': 'lightblue', 'land': 'whitesmoke'}

    if 'land' in color_geomap.keys():
        ax.add_feature(cartopy.feature.LAND.with_scale(resolution),
                        facecolor=color_geomap['land'])
    
    if 'ocean' in color_geomap.keys():
        ax.add_feature(cartopy.feature.OCEAN.with_scale(resolution),
                        facecolor=color_geomap['ocean'])

    if 'coastlines' in geomap_features:
        ax.coastlines(linewidth=0.4, zorder=2, resolution=resolution)
    
    if 'borders' in geomap_features:
        border = cartopy.feature.BORDERS.with_scale(resolution)
        ax.add_feature(border, linewidth=0.3)

    return axis_transformation
```

(only a sketch, may have not thought this through 100%)

nmartensen:
Adding another function argument certainly works, too. I didn't do this because `color_geomap` is already documented to take a a dict or a boolean, and then taking True but not False seemed a bit unusual.

While we are modifying `draw_map_cartopy` &ndash; Is this function part of the public API and must stay compatible with previous releases, or can we move the `axis_transformation = get_projection_from_crs(n.srid)` out into network.plot() and drop the unused function arguments? I can add another commit to this PR if that is acceptable.

nmartensen:
PR updated to illustrate the proposal from the previous comment.

FabianHofmann:
@nmartensen I think that is a sensible change. Perhaps we could keep the extension suggested by @fneum for later. @nmartensen could you make a further small change such that the following works: 
* if `color_geomap` is `True` - geomap colors are set to default. 
* if `color_geomap` is False - no geomap colors are drawn

nmartensen:
>     * if `color_geomap` is `True` - geomap colors are set to default

This is old behaviour, not changed in this PR.

>     * if `color_geomap` is False - no geomap colors are drawn

This is the main functional change by this PR. The first draft implemented this as an early return in `draw_map_cartopy()`; in the second iteration the code no longer calls that function if `color_geomap` is `False`.

nmartensen:
@FabianHofmann I may have misunderstood. I'm not sure if you missed that detail in the proposed code change or what change precisely you are asking for. Could you please clarify? Thanks : )

FabianHofmann:
@nmartensen no you are totally right, my comment was not thought through. Seems that I was confused by the fact that `color_geomap` can be a bool, None or a dict.

FabianHofmann (approved):

FabianHofmann:
Thanks @nmartensen for the PR

PR #339, opened 2022-01-13
networkclustering: remove sklean.cluster.kmeans 1.0 deprecation warning

martacki:
sorry for opening a new PR so shortly before the new release, but when checking functionality of other PRs regarding network clustering, this deprecation warning just drove me crazy (appears for every country when applying the clustering in pypsa-eur). we can also include it in a later release, if more convenient for you.
the deprecation is new in the 1.0 release of sklearn, see: https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html.

it's a small fix. because we don't need the features `.x` and `.y` "feature names", taking the `.values` as darray is good enough and removes the deprecation warning which massively spams the terminal.

in the same go I removed a small bug in the default settings of HAC which were inconsistent with another function.

### DIFF
--- a/pypsa/networkclustering.py
+++ b/pypsa/networkclustering.py
@@ -470,8 +470,8 @@ def busmap_by_kmeans(network, bus_weightings, n_clusters, buses_i=None, ** kwarg
 
     kmeans.fit(points)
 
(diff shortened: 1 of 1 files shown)


FabianHofmann (approved):
That looks good. Good that you did that, annoying warning also drive me crazy...

PR #343, opened 2022-01-21
Plot: add legend

FabianHofmann:

A legend function which draws a legend according to the sizes in the network plot was added. In the docstring there is a runnable example. Please test it. 

I could not figure out why the scaling had to be correct manually, there is a hard coded number `56` which is used to determine the ratio of the default plotting and the scaling of the figure. The current setting is stable against changes in the projection/margin/dpi/figsize.

**A breaking change**
In case no boundaries or no margin is passed, the plot function now uses `autoscale_view` instead of `autoscale`. This makes the circle sizes stable for all different parameter settings. Before the circles had different sizes depending on whether boundaries (or margin) were defined or not. 

## Checklist


FabianHofmann:
A included the projected area factor. There is a problem with rescaling when passing no margin and no bounds. Could you run your tests with an explicit margins or boundaries? For me, everything (even different projections) seem fine then.

fneum:
Yes, so as soon as I add boundaries, it works well. I'd feel uncomfortable merging as long as providing explicit boundaries is necessary. I usually don't. Do you know what the cause is? Can we calculate the boundaries per default ourselves to circumvent this issue?

PR #357, opened 2022-02-04
pypsa/plot.py: do not plot buses with zero size

nmartensen:
This is a small optimisation to speed up plotting of large networks if only some nodes need to be plotted. Instead of adding a lot of patches with zero radius, just skip the corresponding buses.

In a network with a few thousand nodes of which I need to plot only some, this reduces the plotting time from 4-5 seconds to a fraction of a second.


fneum:
Yup, good idea!

PR #375, opened 2022-03-20
apply pre-commit formats

fneum:


doc/index.rst
```diff
@@ -37,6 +37,10 @@ PyPSA: Python for Power System Analysis
     :target: https://gitter.im/PyPSA/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge
     :alt: Chat on Gitter
 
+.. image:: https://results.pre-commit.ci/badge/github/PyPSA/PyPSA/master.svg
```
euronion, line 40: `black` badge missing here

.pre-commit-config.yaml
```diff
@@ -24,13 +24,13 @@ repos:
...
-    args: ['--ignore-regex="\b[A-Z]+\b"'] # Ignore capital case words, e.g. country codes
-    types_or: [python, rst, markdown]
-    files: ^(scripts|doc)/
+# - repo: https://github.com/codespell-project/codespell
+#   rev: v2.1.0
+#   hooks:
+#   - id: codespell
+#     args: ['--ignore-regex="\b[A-Z]+\b"'] # Ignore capital case words, e.g. country codes
+#     types_or: [python, rst, markdown]
+#     files: ^(scripts|doc)/
```
euronion, line 33: Why exclude codespell?

euronion (changes requested):
I'll not check the automatic changes and trust that they are fine ;-)

.pre-commit-config.yaml
fneum: It suggested some corrections for URLs which I couldn't ignore.

FabianHofmann:
But now GH will think that @fneum did all the changes right? Initial idea was to let pre-commit CI do the large parts. Or does that not work for a reason?

fneum:
Ah, yes. Happy to do it another way. I thought someone had to apply it first.

PR #377, opened 2022-03-24
networkclustering: greedy modularity maximization

martacki:
proposing a new clustering method based on electrical parameters of the transmission grid

## Checklist


pypsa/networkclustering.py
```diff
@@ -898,6 +898,97 @@ def hac_clustering(
...
+    """
+
+    if buses_i is None:
+        buses_i = network.buses.index
+
+    network.calculate_dependent_values()
+
+    lines = network.lines.query("bus0 in @buses_i and bus1 in @buses_i")
+    lines = (
+        lines.loc[:, ["bus0", "bus1"]]
```
FabianHofmann, line 936:
```suggestion
        lines[["bus0", "bus1"]]
```

pypsa/networkclustering.py
```diff
@@ -898,6 +898,97 @@ def hac_clustering(
...
+        .assign(weight=network.lines.s_nom / abs(lines.r + 1j * lines.x))
```
FabianHofmann, line 937:
```suggestion
        .assign(weight=lines.s_nom / abs(lines.r + 1j * lines.x))
```

pypsa/networkclustering.py
```diff
@@ -898,6 +898,97 @@ def hac_clustering(
...
+    )
+
+    G = nx.Graph()
+    G.add_nodes_from(buses_i)
+    G.add_edges_from((u, v, dict(weight=w)) for (u, v), w in lines.itertuples())
+
+    communities = nx.community.greedy_modularity_communities(
+        G, best_n=n_clusters, cutoff=n_clusters, weight="weight"
+    )
+    busmap = pd.Series(buses_i, buses_i)
```
FabianHofmann, line 948: Is communities a dict?

FabianHofmann (commented):
some comments, looks good in general

pypsa/networkclustering.py
martacki: should be a list: https://github.com/networkx/networkx/blob/main/networkx/algorithms/community/modularity_max.py#L294

pypsa/networkclustering.py
martacki: should stay `lines`, because `network.lines` still contains all lines, while `lines` contains only the filtered ones (depending on `buses_i`)

martacki: ah, but that's what you suggested. sorry!

fneum:
The dependency `networkx>=2.8` drops Python 3.7 support, but I think we should keep 3.7 working (https://endoflife.date/python).

Preferred alternative: Make the version requirement an optional dependency of the function. In most cases `networkx==2.8` will be installed automatically in new environments.

You could add to the start of the `greedy_modularity_clustering` function:

```py
from distutils.version import LooseVersion
_nx_version = LooseVersion(networkx.__version__)
if _nx_version < "2.8":
    raise NotImplementedError(
        "The function requires 'networkx>=2.8',"
        f"but version 'networkx={_nx_version}' is installed."
    )
```

If we go ahead with adopting only the newest `networkx` version, we should clear out the following segment in a simultaneous PR.

https://github.com/PyPSA/PyPSA/blob/245109666ede7105f381c976291a106af53c2e2f/pypsa/descriptors.py#L35-L93

coroa:
`distutils` is really old. Please avoid! Use the python3 stdlib packaging module instead: https://packaging.pypa.io/en/latest/version.html

```python
from packaging.version import Version, parse 

if parse(nx.__version__) >= Version("1.12"):
    ...
```

martacki:
shouldn't nx version `2.6` be compatible with python `3.7`?  https://networkx.org/documentation/stable/release/release_2.6.html
anyway i'll try removing my nx version changes to `1.10`, then

fneum (approved):
Looks great!

martacki:
Thanks. But I don't understand the redthedocs failing check relating to ... mamba?

fneum:
It's not for you to worry :) see #415

martacki:
> It's not for you to worry :) see #415

Ah, oversaw that one.

PR #399, opened 2022-05-13
Feature solver options handler

Cellophil:
I updated the solver options - as far as I understand the functions in linopt are for pyomo=False only. Cbc and glpk expected a string although checking for empty dict. I removed some checks for None type, since now an empty dict would be returned by the handler.

I could not thoroughly check all cases since I only have cbc on this computer! I did also not push a test specific to cbc - how would tests be set up for specific solvers since not everyone has them?

Best, Andreas

### DIFF
--- a/pypsa/linopt.py
+++ b/pypsa/linopt.py
@@ -37,6 +37,7 @@ import os
 import re
 import subprocess
 from importlib.util import find_spec
+from tkinter import Y
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


Cellophil:
I removed a tkinker import - remaining error seems to be connected to this, unconnected to the issue. Git does not recognize that change, don't know why. I can add it back in?

FabianHofmann:
@Cellophil thanks for the PR. The problem I see (and this is likely the reason for the implementation done in this way), is that CBC and GLKP would allow non-keyword parameters, which we cannot write into a dictionary. I would rather prefer a check, that if `solver_options` is a dict, it is converted keyword arguments for the command line, for

1. CBC `" ".join([f"-{k} v" for k, v in solver_options.items()])`
2. GLPK `" ".join([f"-{k} v" for k, v in solver_options.items()])`

These can be two-liners just at the before appending the `solver_options` string to the command.

PR #411, opened 2022-06-07
Add new column fom_cost to all components with capital costs

### LINKED ISSUE #371: Split capital cost into FOM and investment costs
## Split capital cost in FOM + investment cost
Instead of combining fixed operationand maintenance (FOM) plus annualised investment costs in one attribute, currently called `capital_cost`, split them into 2 attributes and change the objective function accordingly.

This would make it possible to
 (i) store information on FOM and annualised investment costs directly in the pypsa network and
 (ii) also to better calculate the costs of, for example, an early shutdown of a power plant before the end of its lifetime. In the event of an early shutdown, the annualised investment costs would still have to be paid, but not the FOM.

Suggested steps would be:

- [ ] replace attribute `capital_cost` for all components by two new attributes `investment_cost` and `FOM`
- [ ] change objective function to consider  `investment_cost` from all build assets and `FOM` from all active assets

Another option that breaks less with the existing implementation of the costs would be to add only the `FOM `and keep the current definition of `capital_cost`. In case of early phase-out of a power plant, the `FOM` could be subtracted from the `capital_cost` to only consider the annualised investment costs.

@coroa did you include this already into pypsa?

lisazeyen:
Closes #371 .

This PR should add an additional column to all components with capital costs. The capital costs are still including FOM + annualised investment costs. The information about the FOM `fom_cost` can be only used to calculated savings compared to the full capital costs in case of early retirement.

## Checklist


pypsa/component_attrs/generators.csv
Alex-Neve, line 1: Please consider writing "Fixed Operation and Maintenance costs" as description of fom_cost, at least to facilitate searches in the file and in the documentation.

The early retirement could also stem from a constraint on the max GHG emissions, in which case the savings not only stem from FOM costs decrease but also from VOM costs and fuel costs decrease.

pypsa/component_attrs/lines.csv
Alex-Neve, line 1: See my comment for generators.csv: please consider using the description "Fixed Operation and Maintenance costs" to be more explicit and facilitate searches for this parameter.

Alex-Neve:
For each component CSV file, please consider using the description "Fixed Operation and Maintenance costs. Applies to an asset that is active, whether it is ON or OFF. Required to calculate total cost of electricity and enable decommissioning." to be more explicit and facilitate searches for this parameter.

The fixed O&M cost calculation should also be implemented downstream (spatial.py, linopf.py, opf.py, abstract.py, global_constraints.py, optimize.py, pf.py and statistics.py). Do you want to implement it? 
If not I could try to create a separate PR to do it, following yours.

The doc/release_notes.rst should also be updated to indicate the changes.

FabianHofmann:
closing in favor of #1507

PR #416, opened 2022-06-29
test: skip pypower test if import fails

FabianHofmann:
pypower is not importable with newest numpy version. skip test if import fails


fneum (changes requested):
paver needs to move into try: in PF test

fneum (approved):

PR #446, opened 2022-09-11
add redispatch example

### LINKED ISSUE #335: Add redispatch example
We have had repeated questions about how one could model redispatch with PyPSA, and I think this calls for an example:

I think there might be a way based on two network objects, where one distinguishes between copper-plating and a redispatch LOPF with full network resolution:

To model the market in a single bidding zone, one could build another network object from the full network (with grid) that connects all generators, loads and storage units to a single node (i.e. no grid constraints). One could actually also have multiple bidding zones represented by nodes with NTCs or FBMC-like constraints for cross-border flows.

One can then transfer the dispatch decisions back to the full model that includes the grid. One way to do this is to subtract the generator dispatch from the load and remove the generators from the network object. Or, if one needs to keep the generators, set p_min_pu=p_max_pu for each snapshot, which fixes the dispatch obtained from the single-bidding zone market (for LOPF one cannot use p_set for that, that's only for PF).

One can then formulate the redispatch problem, by simply adding the P+ and P- generators which do not have a setpoint yet but an associated cost (and capacity?). Setting the generator p_max_pu=0 and p_min_pu=-1 is no problem, we use this frequently.

The example could be based on the scigrid example for a single snapshot.

fneum:
Closes #335.

Adds an example using SciGRID network outlining how one can model redispatch with PyPSA.

## Checklist

- ~[ ] Unit tests for new features were added (if applicable).~
- ~[ ] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).~

### DIFF
--- a/doc/examples-lopf.rst
+++ b/doc/examples-lopf.rst
@@ -13,6 +13,7 @@ This section provides examples related to network optimization with PyPSA. Note 
     examples/lopf_with_pyomo_False.ipynb
     examples/capacity-constraint-per-bus.ipynb
     examples/multi-investment-optimisation.ipynb
+    examples/scrigrid-redispatch.ipynb
 
 
 .. toctree::
--- a/doc/examples-lopf.rst
+++ b/doc/examples-lopf.rst
@@ -13,7 +13,6 @@ This section provides examples related to network optimization with PyPSA. Note 
     examples/lopf_with_pyomo_False.ipynb
     examples/capacity-constraint-per-bus.ipynb
     examples/multi-investment-optimisation.ipynb
-    examples/scrigrid-redispatch.ipynb
 
 
 .. toctree::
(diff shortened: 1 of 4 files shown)


FabianHofmann (commented):
Looks good/ I do not see the example from [RTD built](https://pypsa--446.org.readthedocs.build/en/446/examples-lopf.html#) though, might be visible as soon as merged.

fneum:
> Looks good, I do not see the example from [RTD built](https://pypsa--446.org.readthedocs.build/en/446/examples-lopf.html#) though, might be visible as soon as merged.

Just a stupid typo....

fneum:
https://pypsa.readthedocs.io/en/latest/examples/scigrid-redispatch.html

PR #465, opened 2022-10-21
linopf: Allow empty lhs in nodal balance for zero rhs

coroa:

For multiperiod investment optimisation, it is easy to trigger the `Empty LHS in nodal balance constraint.` exception, since `Bus`es do not have a `build_year`. For example if one models an extendable battery with two links, which have a `build_year` in the future and a bus, then the bus will necessarily be isolated and lead to an all-zero row in the nodal balance constraint matrix part for that bus, which can only be mitigated by add an always present other component to this bus, like a `p_nom=0` non-extendable generator.

In this PR, we instead build a mask that filters out these rows completely.

Implemented for both `linopf` and `linopy`.

## Checklist

### DIFF
--- a/pypsa/linopf.py
+++ b/pypsa/linopf.py
@@ -459,8 +459,6 @@ def define_nodal_balance_constraints(n, sns):
         .reindex(columns=n.buses.index, fill_value="")
     )
 
-    if (lhs == "").any().any():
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):
thanks @coroa

PR #502, opened 2022-11-21
add python 3.11 to CI

fneum:


FabianHofmann:
pytable is not yet available for python3.11, see https://github.com/PyTables/PyTables/issues/976

Alternatively, we could use 
```
pip install git+https://github.com/PyTables/PyTables.git@v3.7.0
```

PR #521, opened 2022-12-13
Add max_relative_growth as a carrier based global constraint

coroa:
Extends the linopy implementation of the carrier's growth limit by an
additional linear term, so that one can allow an expansion growth by
1.5 times what was installed in the investment period before.

This is added directly to the max growth limit constraint so that it adds additional allowed capacity.

One difficult communications perspective is that this is added as additional capacity allowance to
the already existing max growth limit.

$$
G_c(\Pi) - \nu G_c(\Pi - 1) <= \mu
$$

where 

$$
G_c(\Pi) = \sum_{g \in \mathcal G(c, \Pi)} G_g
$$

with the set of generators with carrier $c$ that became active in inv period $\Pi$ called: $\mathcal G(c, \Pi)$.

@lisazeyen I do not see a systematic performance decrease with this.

Closes # (if applicable).

## Checklist


FabianHofmann:
Something is going on with the groupby functionality here https://github.com/PyPSA/PyPSA/actions/runs/3687372274/jobs/6240867131#step:10:279

Already saw this in #365  

I think it has to do with the new xarray version. I will investigate later.

coroa:
> Something is going on with the groupby functionality here https://github.com/PyPSA/PyPSA/actions/runs/3687372274/jobs/6240867131#step:10:279
> 
> Already saw this in #519
> 
> I think it has to do with the new xarray version. I will investigate later.

Note that this does not include any tests as of yet, so my code changes are not even triggered, so the `linopy` errors are unrelated.

FabianHofmann (approved):

FabianHofmann:
@coroa I added a test, could you add a release note. Then we are good to go.

pypsa/linopf.py
```diff
@@ -739,6 +739,13 @@ def define_growth_limit(n, sns, c, attr):
     ext_i = get_extendable_i(n, c)
     if "carrier" not in n.df(c) or n.df(c).empty:
         return
+
+    if n.carriers.max_relative_growth.notnull().any():
+        logger.warning(
+            "Max relative growth is not implemented for the native pypsa optimization framework. "
+            "Use the linopy framework with `n.optimize` instead."
+        )
```
coroa, line 747: Sorry, seeing this message only now!

@FabianHofmann 

This will always trigger, since the default value for `max_relative_growth = 0`.

This should use
```
if (n.carrier.max_relative_growth > 0).any():
    ...
```

FabianHofmann: Ah thanks, will fix it

PR #557, opened 2023-02-10
optimize: speed up large model creation

FabianHofmann:
Speeds up model creation for spatially large models. 

Example: 181 sector coupled model with 24 snapshots. Creation time reduced from 80 sec to 3 sec.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
 
 # Make docstrings PEP 257 compliant
 - repo: https://github.com/PyCQA/docformatter
-  rev: v1.6.0.rc1
+  rev: v1.5.1
   hooks:
   - id: docformatter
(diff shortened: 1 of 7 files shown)


fneum (commented):
Neat!

Should `skip_isolated_buses=True` be the default?

Should we also pass this in `linopf` and `pyomo` versions?

FabianHofmann:
I didn't want to change the default behavior, that's why I left it. We could add a warning that in future versions this will be changed. Yes, forwarding seems to be reasonable.

fneum (approved):

fneum:
But something is going wrong with tests. Is this related to this PR?

FabianHofmann:
It threw error with the linopf code. Can also introduce later if needed.

fneum:
:+1:

PR #563, opened 2023-02-20
Fix IO bug

lisazeyen:
Closes # (if applicable).

## Change IO function to avoid renaming of the index
Currently, when exporting a network the index is rename, e.g. for Links index name before export is "`Link`", index after export is "`links_i`". This causes an error in the nodal balance constraint if the network is solved with linopy (`n.optimize)`, if one for example first exports the presolved network and then runs the optimisation in the same script.

In general, export functions should not change the index names or structure of the network.

## Checklist

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -11,7 +11,6 @@ dependencies:
 - pytables
 - matplotlib
 - networkx>=1.10
-- linopy
 - pyomo>=5.7.0
 - cartopy>=0.16
 - glpk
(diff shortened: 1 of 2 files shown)


fneum (commented):
@coroa's counter-proposal does not need to copy data. Let's prioritise that approach.

FabianHofmann:
@lisazeyen thanks anyway for reporting and narrowing the bug down

PR #564, opened 2023-02-20
io: Avoid changing axis names on export

coroa:
Alternative proposal for #563 .

### DIFF
--- a/pypsa/io.py
+++ b/pypsa/io.py
@@ -350,11 +350,8 @@ if has_xarray:
             # therefore we have to work-around. Note that this does not copy data
             index = df.index.copy()
             index.name = "snapshots"
-            df = pd.DataFrame(
-                df,
-                index=index,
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):

pypsa/io.py
```diff
@@ -328,24 +328,31 @@ def save_meta(self, meta):
             self.ds.attrs["meta"] = json.dumps(meta)
 
         def save_snapshots(self, snapshots):
-            snapshots.index.name = "snapshots"
+            snapshots = snapshots.rename_axis(index="snapshots")
```
coroa, line 331: Ai, this can be a multi-index, too, can it not? @FabianHofmann

FabianHofmann: Indeed it can be, but I think this should still be fine. Pandas multiindex have a name and level names. Can test it later though.

PR #566, opened 2023-02-21
[fix] io: Avoid changing axis names on export

coroa:
Follow-up to #564 to fix shortcoming highlighted in https://github.com/PyPSA/PyPSA/pull/564#discussion_r1112805430 .


coroa:
Don't merge yet. Sorry.

PR #569, opened 2023-02-22
io: add functionality to read .nc and .h5 files from URL

fneum:
NetCDF (``.nc``) and HDF5 (``.h5``) network files can now be read directly from URL:

 ```py
url = "https://github.com/PyPSA/PyPSA/raw/master/examples/scigrid-de/scigrid-with-load-gen-trafos.nc"
n  = pypsa.Network(url)
```

I moved the `_data_dir` from `pypsa.examples` to `pypsa.io`. Does that make sense?

Future additions could address caching of downloads similar to the `pypsa.examples`. (e.g. with `update=True` argument or similar). For now, files are downloaded again with every call.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -7,7 +7,7 @@ Upcoming Release
 
 .. warning:: The features listed below are not released yet, but will be part of the next release! To use the features already you have to install the ``master`` branch, e.g. ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
 
-* new feature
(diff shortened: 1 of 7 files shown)


FabianHofmann (approved):
Awesome!

FabianHofmann:
> I moved the `_data_dir` from `pypsa.examples` to `pypsa.io`. Does that make sense?

Yes! 

> Future additions could address caching of downloads similar to the `pypsa.examples`. (e.g. with `update=True` argument or similar). For now, files are downloaded again with every call.

:+1:

PR #576, opened 2023-03-08
Fix preservation of non-aggregated time-dependent variables

koen-vg:
When aggregating generators, the current version of the code erroneously only preserves time-dependent variables of generators when some generators with time-dependent variables had been aggregated (the `if not df_agg.empty` statement right about this commit diff). In this case, time-dependent variables of non-aggregated generators are simply forgotten.

The problem arose in the context of building some pypsa-eur network with two clustering steps (for example with wildcards simpl=200 and clusters=100); if the network only has renewable generators which aren't aggregated in the second clustering step, their capacity factors  (`p_max_pu`) would be forgotten. This PR fixes that issue.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -19,6 +19,8 @@ Upcoming Release
   attributes had to carry identical values for components that were to be
   merged.
 
+* Fix a bug where time-dependant generator variables could be forgotten during aggregation in a particular case.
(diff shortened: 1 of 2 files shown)


fneum (approved):
Great catch! Many thanks for providing the fix!

PR #591, opened 2023-03-31
update examples to linopy framework

FabianHofmann:


doc/examples-lopf.rst
```diff
@@ -3,20 +3,16 @@ Power System Optimization
 #########################
 
 
-This section provides examples related to network optimization with PyPSA. Note that PyPSA allows for optimising networks in three ways, first using the ``pyomo`` interface and second using the in-home implementation (which is way faster and memory-efficient). The third one is the new Linopy based implementation, which was inspired by the in-house optimization code. We are *slowly* moving towards using the linopy implementation only.
+This section contains examples of network optimization with PyPSA. Note that PyPSA allows you to optimize networks in three ways: based on Linopy, based on Pyomo, and based on custom code. PyPSA's default optimization framework is based on Linopy. In earlier versions, PyPSA used the ``pyomo`` interface and an in-house implementation, both available through the `lopf` function. The latter are still available, but will be deprecated in the future. The following examples all use the Linopy-based implementation.
```
fneum, line 6:
```suggestion
This section contains examples of network optimization with PyPSA. PyPSA allows you to optimize networks in three ways: with `Linopy <https://linopy.readthedocs.io>`_, with `Pyomo <http://www.pyomo.org/>`_, and based on `custom code <https://github.com/PyPSA/PyPSA/blob/v0.22.1/pypsa/linopt.py>`_. The new default optimization framework is based on Linopy. In earlier versions, PyPSA used the ``pyomo`` interface and an in-house implementation, both available through the ``n.lopf()`` function. The previous options are still available but will be deprecated in the future. The following examples all use the Linopy-based implementation.
```

doc/examples-lopf.rst
```diff
@@ -3,20 +3,16 @@ Power System Optimization
...
+
 
 .. toctree::
     :maxdepth: 1
-    :caption: No Pyomo
 
+    examples/optimization-with-linopy.ipynb
     examples/lopf_with_pyomo_False.ipynb
```
fneum, line 13: This is removed, right?

fneum (changes requested):
Only a few small comments:

- Meshed AC-DC and the sector-coupling examples (https://pypsa--591.org.readthedocs.build/en/591/examples-sector_coupling.html) and other examples (https://pypsa--591.org.readthedocs.build/en/591/examples-other.html) still use n.lopf() and could also be switched to n.optimize()
- scigrid sclopf should be network.optimize.security_constrained_optimization rather than network.optimize.security_constraint_optimization?

examples/notebooks/replace-generator-storage-units-with-store.ipynb
```diff
@@ -37,56 +37,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "def replace_gen(network, gen_to_replace):\n",
-    "    \"\"\"Replace the generator gen_to_replace with a bus for the energy\n",
```
fneum, line 41: Why remove the example of replacing a generator with links and stores?

fneum (changes requested):
Thanks! What was the reason to remove the generator to link/store combination example?

PR #594, opened 2023-04-03
Update optimization section to linopy framework

FabianHofmann:


FabianHofmann:
@lisazeyen could you have a small look? Hope I got all parts where we were referring to the pyomo implementation

PR #602, opened 2023-04-13
bugfixes for native implementation of unit commitment for links

nworbmot:
Incorrect attribute was taken and we need to avoid applying lower and upper bounds to committable links twice.

Fixes minor bugs from https://github.com/PyPSA/PyPSA/pull/582.


fneum (approved):
Superb! Thanks for tracing down this bug!

PR #604, opened 2023-04-14
couple improvements

Cellophil:


Cellophil:
I have the suspicion that small impedances cause numerical problems in the PTDF calculation - can someone confirm or check? Possibly a warning could be put in consistency check if the impedances are too far apart.
Apart from that I converted PTDF and BODF into DataFrames, and replaced the sparse solve by sparse inverse. I don't know why sparse solve was used, maybe there's a good reason, but inverse seems to be the dedicated function to do it.

fneum:
Do you have a small example where you see numerical problems in the PTDF calculation, that could help us reproduce the behaviour?

Regarding the PR: I think using the scipy inversion function makes a lot of sense; also turning PTDF and BODF into DataFrames. But the tests fail because the code is not yet fully adapted to take DataFrames rather than numpy arrays for these.

Cellophil:
Hi, yes I realized, I'll debug. There's a sum in the bodf calculation that adds a numpy matrix and a scipy sparse, that fails at the moment. Maybe the possibility of keeping Ptdf and bodf sparse was a reason for returning pure numpy in the first place...

Cellophil:
I had a case where the B matrix had entries of 1e9 - the solver reported a singular matrix. Scaling down actually "solved" it but adapting the impedance spread is maybe better. I'll extract an example.
Maybe an optional sparse return format for ptdf and bodf could be an unstacked version of a Dataframe ie multi index pandas Series with cutoff?

Cellophil:
Updates on my side:

- tbh, I'm not so sure about the numerical issue anymore, possibly I had a hickup there
- the various matrices calculated in sub_network rely on using the same bus and branch ordering in a number of places - probably in order to use scipy sparse matrices. Some functions use for loops a lot, maybe there's a speedup to be gained (incidence matrix for instance)
- An equivalent of sparse scipy matrices could be set up in the form of Pandas Series with multiindexes. It's an extra step in the middle and probably a bit slower than the blank handling of numpy and scipy, but it would keep track of the index.
- I'll try to set up a working version that accepts the result at the end and attaches the correct indexes. For now.

Cellophil:
This is probably on me but I have problems with pytest - pandapower and it's numba dependency I think. Sorry for the failed checks, hope it starting to take form.

Cellophil:
Hi @fneum ,

I replaced the usage of BODF and PTDF is a dozen places; and I feel it really makes sense to continue with pandas instead of the annonymous numpy. Some further things should be cleaned up, I patched it in some places...
What I couldn't really check is p_branch_shift and p_bus_shift, I don't know whether that's covered by the test cases. And what it does.

Best,
Andreas

Cellophil:
@fneum One test fails concerning tkinter and some plotting library - it passes on my machine.

pypsa/pf.py
```diff
@@ -1320,6 +1336,23 @@ def calculate_Y(sub_network, skip_pre=False):
     )
 
 
+def get_network_ptdf(network):
+    # convenience function
+
```
fneum, line 1341: Maybe add a small docstring that this gives PTDF for whole network (not per subnetwork).

pypsa/pf.py
```diff
@@ -1187,7 +1190,9 @@ def calculate_B_H(sub_network, skip_pre=False):
             for c in sub_network.iterate_components(network.passive_branch_components)
         ]
     )
-    sub_network.p_branch_shift = np.multiply(-b, phase_shift, where=b != np.inf)
+    sub_network.p_branch_shift = (
+        -b * phase_shift
+    )  # np.multiply(-b, phase_shift, where=b != np.inf)
```
fneum, line 1195: Could you briefly explain why `where=b != np.inf` is no longer needed?

And remove old code.

pypsa/pf.py
```diff
@@ -1156,24 +1157,26 @@ def calculate_B_H(sub_network, skip_pre=False):
...
+    # old: b = np.divide(1.0, z, out=np.full_like(z, np.inf), where=z != 0)
 
-    if np.isnan(b).any():
+    if b.isna().any():
         logger.warning(
             "Warning! Some series impedances are zero - this will cause a singularity in LPF!"
         )
-    b_diag = csr_matrix((b, (r_[: len(b)], r_[: len(b)])))
+
+    # old: b_diag = csr_matrix((b, (np.r_[: len(b)], np.r_[: len(b)])))
```
fneum, line 1170:
```suggestion
```

pypsa/pf.py
```diff
@@ -1156,24 +1157,26 @@ def calculate_B_H(sub_network, skip_pre=False):
...
-        [
-            (c.df.loc[c.ind, attribute]).values
-            for c in sub_network.iterate_components(network.passive_branch_components)
-        ]
-    )
+    z = sub_network.branches()[attribute]
+    b = 1.0 / z
     # susceptances
-    b = np.divide(1.0, z, out=np.full_like(z, np.inf), where=z != 0)
+    # old: b = np.divide(1.0, z, out=np.full_like(z, np.inf), where=z != 0)
```
fneum, line 1163:
```suggestion
```

pypsa/pf.py
```diff
@@ -1205,17 +1210,24 @@ def calculate_PTDF(sub_network, skip_pre=False):
...
         calculate_B_H(sub_network)
 
     # calculate inverse of B with slack removed
 
     n_pvpq = len(sub_network.pvpqs)
-    index = np.r_[:n_pvpq]
-
-    I = csc_matrix((np.ones(n_pvpq), (index, index)))
+    # I = eye(n_pvpq)
+    # B_inverse = spsolve(csc_matrix(sub_network.B[1:, 1:]), I)
```
fneum, line 1228:
```suggestion
```

pypsa/pf.py
```diff
@@ -24,10 +24,11 @@
 from numpy import ones, r_
 from numpy.linalg import norm
 from pandas.api.types import is_list_like
-from scipy.sparse import csc_matrix, csr_matrix, dok_matrix
+from scipy.sparse import csc_matrix, csr_matrix, diags, dok_matrix, eye
```
fneum, line 27:
```suggestion
from scipy.sparse import csc_matrix, csr_matrix, diags, dok_matrix
```

pypsa/pf.py
```diff
@@ -1156,24 +1157,26 @@ def calculate_B_H(sub_network, skip_pre=False):
 
     # following leans heavily on pypower.makeBdc
 
-    z = np.concatenate(
-        [
-            (c.df.loc[c.ind, attribute]).values
-            for c in sub_network.iterate_components(network.passive_branch_components)
-        ]
-    )
+    z = sub_network.branches()[attribute]
```
fneum, line 1160: Oh, this looked very complicated beforehand.

fneum (commented):
Sorry for taking a long time to get to this! This looks very good!

I had a few small questions. Maybe you can answer them.

Have you tested that the code changes yield same PTDF as before?

The one test failure is fine.

Final thing: Would you add a release note outlining the changes? I.e. BODF and PTDF are now dataframes and new utility function to build PTDF for whole network.

Cellophil:
Hi @fneum , great I'll check it out next week - agreed code needs a little polishing. With ptdf index, I'll check it once again that it's correct.

FabianHofmann:
Hey @Cellophil, gently pinging you. Do you think it is feasible for you to resolve the remaining issues?

Cellophil:
Hi, sry for the delay, I took parental leave. Time is unfortunately a bit limited atm, but I'll check things this week. Best, Andreas

FabianHofmann:
Hey Andreas, no worries, thanks for letting us know. The young ones have full priority :) We can also try to take it from here.

PR #618, opened 2023-05-09
global_constraint: add operational limit constraint

FabianHofmann:
Analogous to `primary_energy_limit` but directly working on carrier output. This will help to replace some stores in pypsa-eur (sector-coupled) by light-weight constrained generators.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -33,6 +33,8 @@ Upcoming Release
 
 * Fix an issue appeared when processing networks which were reduced to a set of isolated nodes in course of clustering. Previously, an empty ``Line`` component has lead to problems when processing empty lines-related dataframes. That has been fixed by introducing special treatment in case a lines dataframe is empty.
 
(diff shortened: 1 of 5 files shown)


fneum:
Can you explain a little more about how it works? I don't yet understand why it is better than a Store.

FabianHofmann:
The reason is that modelling stores is quite extensive. You have intertemporal constraints with auxiliary variables (soc, charge, discharge). If you limit the total energy output of one single generator (or a generator carrier fleet) instead of setting up a store with an initial soc which depletes over time, you save a lot of complexity. So (reducing the generators) the constraint is just saying 

```math
\sum_{G \in \Theta, t} g_{G, t} <= limit
```
where $\Theta$ is the set of generators which belong to the carrier set in the global constraint.

fneum (approved):
Looks good! Nicely tested, so it was easy to review.

fneum:
At some point, we need to update the documentation of the GlobalConstraint component:

https://pypsa.readthedocs.io/en/latest/components.html#global-constraints

tintin305:
I don't see any place that caters for selecting investment_periods. Will this still need to be introduced as a parameter?

PR #628, opened 2023-05-10
move networkclustering module to clustering.spatial

fneum:
Up for discussion!

The motivation here is that it will create a nice structure to add a temporal clustering module based on `tsam`:

```py
from pypsa.clustering import spatial, temporal
```


fneum:
@lisazeyen approved off-list

FabianHofmann:
This was a bit of brute force. Now the clustering.py lost track of all the changes and commits. @fneum did you use `git mv` when copying the file?

fneum:
Oh right, I didn't think of that... Sorry!

FabianHofmann:
No worries :) we lost contributions, but I think it's okay. As far as I understand, the PR in #632 would add the reversion on top, not revert the initial commit (i.e. the tracking would be missing). One could do a force push on the master, but people who pulled in the meantime have to do a `git reset --hard origin master` then :dancers:

PR #632, opened 2023-05-10
Revert "move networkclustering module to clustering.spatial"

FabianHofmann:
Reverts PyPSA/PyPSA#628


FabianHofmann:
No, that does not do it. Never mind.

fneum:
That would revert the PR, doesn't it?

PR #635, opened 2023-05-15
Assign duals

lisazeyen:
Closes # (if applicable).
This PR should assign all dual variables to the pypsa network in case the name of the constraint follows the syntax ("`component-something`").
- if attribute "something" is not part of the `n.pnl(c)` it is added
- if component is "`GlobalConstraint`" and the attribute is not time-dependent, the dual variable is added to `n.global_constraints[name, "mu"]`
- further this PR adds back the keyword arguement `keep_shadowprices` from nomopyomo, which allows to specify a list of component names from which the dual variables should be saved

I think it needs some careful checking since I also modified the function `set_from_frame()`

## Checklist


FabianHofmann:
Hey @lisazeyen, looks good :) However, do you think we really need the `keep_shadowprices` argument? I think it can be misleading, as the shadow prices are all there (in `n.model`), they are not just not all added to the model. I find the default behaviour that you have implemented now, totally fine as a standard behaviour which can not be altered.  

Also, could you make sure that the `mu` of the global constraint is only added if `name` is in the index of `n.global_constraints`?

lisazeyen:
> Hey @lisazeyen, looks good :) However, do you think we really need the `keep_shadowprices` argument? I think it can be misleading, as the shadow prices are all there (in `n.model`), they are not just not all added to the model. I find the default behaviour that you have implemented now, totally fine as a standard behaviour which can not be altered.
> 
I am fine with removing the keep_shadowprices argument as well, I just thought it would be nice for consistency with nomopyomo.

> Also, could you make sure that the `mu` of the global constraint is only added if `name` is in the index of `n.global_constraints`?

Mhm, actually I want to add them to the dataframe even if they the constraint is not there yet. Let's say I have a custom constraint which applies for multiple components, then I would like to save the shadow prices in the global constraints dataframe.

 If I first have to add the constraint to the network it is getting messy because there are constraints like the CO2 emissions, where adding the constraint with `n.add("GlobalConstraint", ...) ` enforces the constraint and then there would be my custom constraints, which I would have to formulate with `n.model.add_constraints(..) `and add with `n.add("GlobalConstraint"...) `but the latter would have no impact on the optimisation.

fneum (approved):
LGTM.

PR #642, opened 2023-05-25
Add deprecation warning to networkclustering module

FabianHofmann:
Shows a deprecation warning when import `networkclustering` instead of `clustering.spatial`


fneum:
Is it ok that it shows on every import of PyPSA?

```py
import pypsa
```

PR #645, opened 2023-06-07
`assign_duals`: add storage consistency equations

fneum:
- removed `e_set` and `s_set` because they are not used
- renamed storage consistency equation from `energy-balance` to `energy_balance` for consistency with other constraints
- added `mu_energy_balance` to `n.stores_t` and `n.storage_units_t`
- track missed unassigned dual variables (those that are not matched but also don't cause an error)


nworbmot (approved):
I tested and it works well!

PR #648, opened 2023-06-16
add time-resolved energy balance to statistics module

### LINKED ISSUE #578: Improve n.statistics() defaults
Avoid confusion in df aggregation methods. Instead of defining a single default time aggregation method in class `StatisticsAccessor`, use defaults from each metric function.

p-glaum:
- closes #578 

- enhances n.statistics by
    - allowing to retrieve time series (e.g. with `n.statistics.supply(aggregate_time=False)`)
    - adding `n.statistics.energy_balances()`

## Checklist


p-glaum:
Noticed that something is wrong. I will try to fix it first.

p-glaum:
> Noticed that something is wrong. I will try to fix it first.

Fixed the bug and PR is no ready for review

pypsa/statistics.py
```diff
@@ -125,9 +144,9 @@ def __init__(self, network):
     def __call__(
         self,
         comps=None,
-        aggregate_time="mean",
```
fneum, line 128: Does this need to be removed?

fneum (commented):
Superb! I tested it with the hydrogen network .nc files. This will save a lot of time!

One additional feature request and one documentation improvement:

- It would be nice to also have an option `aggregate_buses` that you can disable/enable. In this way, you could have energy balances per bus.
- The docstrings should explain a little in which settings the snapshot weightings are applied and when they are not (e.g. if I say `aggregate_time=None`, are the values in MWh or MW?)

pypsa/statistics.py
p-glaum: I thought it is better because the call method is applied to all statistics methods. And as @Irieo suggested in his issue #578, this is confusing because one would assume that e.g. for curtailment you are not interested in the mean curtailment but in the total curtailment.

p-glaum:
> Superb! I tested it with the hydrogen network .nc files. This will save a lot of time!
> 
> One additional feature request and one documentation improvement:
> 
>     * It would be nice to also have an option `aggregate_buses` that you can disable/enable. In this way, you could have energy balances per bus.

I try to add this feature

>     * The docstrings should explain a little in which settings the snapshot weightings are applied and when they are not (e.g. if I say `aggregate_time=None`, are the values in MWh or MW?) With time aggregation it will be in energy, right?

You are right. When we aggregate time we get MWh and if we get the time series we will have MW. I will add this in the documentation

p-glaum:
Hi @FabianHofmann,

yes, I removed them intentionally. I was speaking to @fneum and we thought that we do not need it because the positive values of the energy balance are the supply and the negative the withdrawal. But maybe this is not so intuitive and I can add them back. However, I would change both functions to make them more generic and not only consider bus0 and bus1

euronion:
Comment from the sidelines: Please try to keep backward compatibility. If you change the behaviour of existing functions or deprecate them, do so with > 1 version heads-up.

FabianHofmann:
I strongly agree with @euronion, I would not remove them just like that. Also, I think it's useful to have a function that are not netting out. Think about stores and storage units. This supply quite a lot over time but their total balance is quite low... yes, including the other buses for multi links might be more generic, and rather corrects things. Thanks @p-glaum

fneum:
Excellent! I also tested the energy balance per bus just now and can confirm that it works. Supply and Withdrawal functions are kept in this revision. LGTM

PR #649, opened 2023-06-20
Add option to aggregate carriers in specific buses

LukasFrankenQ:

This PR adds a feature to the function `clustering.spatial.get_clustering_from_busmap`, and `aggregategenerators`.
They now would accept the kwarg `aggregate_generator_buses` and `aggregate_buses`, respectively. These are a list of buses (of the origin network), where carriers are aggregated, irrespective of the argument passed for `carriers`.

The motivation here is that this greatly facilitates a feature (for which I will also make a PR) in PyPSA-Eur, that allows generators to be aggregated in a subset of regions. In my case for instance, I would like to model all gas and nuclear generators individually in GB, but I don't wish to model ~100 plants in France and Germany. To me at least, this appears like a features that could easily be useful for studies based on PyPSA-Eur, but currently requires hacking of some quite deep functionalities of `PyPSA`, which might turn off some users. I think this PR would unlock this functionality for non-hackers.

My understanding of `clustering.spatial` is quite partial, so please let me know if the proposed changes will not work as I intend, I am more than happy to amend.

All the best,
Lukas

## Checklist

- [X] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [X] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.

### DIFF
--- a/pypsa/clustering/spatial.py
+++ b/pypsa/clustering/spatial.py
@@ -57,12 +57,22 @@ def _make_consense(component, attr):
 
 
 def aggregategenerators(
-    network, busmap, with_time=True, carriers=None, custom_strategies=dict()
+    network,
+    busmap,
+    with_time=True,
+    carriers=None,
+    custom_strategies=dict(),
+    buses=None,
 ):
     if carriers is None:
         carriers = network.generators.carrier.unique()
-
+    
     gens_agg_b = network.generators.carrier.isin(carriers)
+
+    if buses is not None:
+        # add all generators in buses that should be clustered
+        gens_agg_b += network.generators.bus.isin(buses)
+
     attrs = network.components["Generator"]["attrs"]
     generators = network.generators.loc[gens_agg_b].assign(
         bus=lambda df: df.bus.map(busmap)
@@ -385,6 +395,7 @@ def get_clustering_from_busmap(
     bus_strategies=dict(),
     one_port_strategies=dict(),
     generator_strategies=dict(),
+    aggregate_generator_buses=None,
 ):
     buses, linemap, linemap_p, linemap_n, lines, lines_t = get_buses_linemap_and_lines(
         network, busmap, line_length_factor, bus_strategies, with_time
@@ -415,6 +426,7 @@ def get_clustering_from_busmap(
             with_time=with_time,
             carriers=aggregate_generators_carriers,
             custom_strategies=generator_strategies,
+            aggregate_buses=aggregate_generator_buses,
         )
         io.import_components_from_dataframe(network_c, generators, "Generator")
         if with_time:
--- a/pypsa/clustering/spatial.py
+++ b/pypsa/clustering/spatial.py
@@ -395,7 +395,7 @@ def get_clustering_from_busmap(
     bus_strategies=dict(),
     one_port_strategies=dict(),
     generator_strategies=dict(),
-    aggregate_generator_buses=None,
+    aggregate_generators_buses=None,
 ):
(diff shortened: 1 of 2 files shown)


p-glaum:
Great @LukasFrankenQ! I like that feature because I just thought of implementing something similar for my offshore modelling. I am mainly interested in the North Sea and for this region I don't want to aggregate the offshore generators. But for other regions it is fine.

LukasFrankenQ:
Thanks @p-glaum, yeah, I though it could be nice. I am however unsure if passing a list of buses from the source network is the best way to do it, it would feel more intuitive to pass a list of buses from the target network... Happy about any feedback here :)
Also, I created a mess keeping the fork updated and trying to squash... If there is interest in this feature I would be happy to clean things up before the merge :)

pypsa/clustering/spatial.py
```diff
@@ -57,12 +57,22 @@ def consense(x):
 
 
 def aggregategenerators(
-    network, busmap, with_time=True, carriers=None, custom_strategies=dict()
+    network,
+    busmap,
+    with_time=True,
+    carriers=None,
+    custom_strategies=dict(),
+    aggregate_buses=None,
```
FabianHofmann, line 65: I'd vote for naming it `buses`, to align with the `carriers` argument.

pypsa/clustering/spatial.py
```diff
@@ -57,12 +57,22 @@ def consense(x):
...
 ):
     if carriers is None:
         carriers = network.generators.carrier.unique()
 
     gens_agg_b = network.generators.carrier.isin(carriers)
+
+    if aggregate_buses is not None:
+        # add all generators in buses that should be clustered
+        gens_agg_b += network.generators.bus.isin(aggregate_buses)
```
FabianHofmann, line 74: Shouldn't it be `|=` since it is boolean series?

pypsa/clustering/spatial.py
```diff
@@ -385,6 +395,7 @@ def get_clustering_from_busmap(
     bus_strategies=dict(),
     one_port_strategies=dict(),
     generator_strategies=dict(),
+    aggregate_generator_buses=None,
```
FabianHofmann, line 398:
```suggestion
    aggregate_generators_buses=None,
```
to align with the other arguments

pypsa/clustering/spatial.py
```diff
@@ -415,6 +426,7 @@ def get_clustering_from_busmap(
             with_time=with_time,
             carriers=aggregate_generators_carriers,
             custom_strategies=generator_strategies,
+            aggregate_buses=aggregate_generator_buses,
```
FabianHofmann, line 429:
```suggestion
            aggregate_buses=aggregate_generators_buses,
```

FabianHofmann (changes requested):
Thanks @LukasFrankenQ, just a few comments

pypsa/clustering/spatial.py
LukasFrankenQ: Good point!

pypsa/clustering/spatial.py
LukasFrankenQ: They seem to work the same way, I find `+=` a bit more intuitive, and found myself looking up `|=` once upon a time but I am happy about either.

LukasFrankenQ:
Alright, thanks for the review @FabianHofmann, all suggestions are worked in, and commits are nice and squashed! Let me know if further changes are desired!

FabianHofmann (approved):
Great, thanks @LukasFrankenQ

PR #655, opened 2023-06-27
add quadratic marginal cost for stores and storage units

fneum:


coroa:
What's the plan for the implementation in the objective function? is there already linopy support for a quadratic objective?

fneum:
It's already done! Both in linopy and PyPSA.

coroa:
Ah very, good!

https://github.com/PyPSA/PyPSA/blob/6ef116fe73d19062610647479cf50c09d165f1ee/pypsa/optimization/optimize.py#L129-L140

Do you know by any chance whether non-diagonal quadratic objective function parameters are also supported by linopy now, in the spirit of:

v A v where A is a quadratic matrix and v the vector of variables?

FabianHofmann:
Yes they are :)

PR #372, opened 2022-03-15
Discrete investment

coroa:
Alternative to #371.

## Changes proposed in this Pull Request

Extends lowmem lopf for generators by 3 new attributes:

- `buildable`
- `retirable`
- `fom_cost`

`buildable` allows a fixed capacity power plant to be built as proposed or not. `retirable` allows a fixed capacity power plant to be retired before the actual lifetime, salvaging fom_cost * years of early retirement.

This is an untested draft.

## Checklist

- [ ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ ] Unit tests for new features were added (if applicable).
- [ ] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #666, opened 2023-07-04
Add modularity to components

### LINKED ISSUE #658: Modular investment capacity

It would be interesting to add the modularity in the investment capacity: rather than having a pure LP formulation, enabling the possibility to install a "number" of components.

I reflected on this topic and I see two options, detailed below:
1. (my favourite solution) add a new integer variable n_nom that describes the number of units installed and add a constraint "p_nom (variable) == p_nom (parameter)  * n_nom (variable)", where the latter p_nom shall be the optional parameter set during the "add/madd" function.
2. revise the formulation of the algoritm from p_nom to n_nom. This may be slightly more efficient from a mathematical point of view, but a breaking change. So I wouldn't recommend this option

This feature may be interesting to have and especially option 1 may be easy to implement.
The needed elements may be:
1. add a new `define_integer` function to add integer-valued function, accounting for the LP format to add integer variables, hence using "General" section https://www.ibm.com/docs/en/icos/12.8.0.0?topic=representation-mip-features-in-lp-file-format but also adding a constraint to specify that they are non-negative
2. Add an appropriate constraint to set: "p_nom (variable) == p_nom (parameter)  * n_nom (variable)".
   To avoid adding additional parameters and since the parameter p_nom generally loses value when using extendable, we may use that same parameter to specify the size of each base unit

This simple implementation is consistent with the current PyPSA structure. A follow-up of this could be a revision of the unit commitment to address multi-units of the same type.
This can be done by extending to integer the variable status and simple adjustments.

With @enricogiglio, we could work on this, but comments are welcome to steer the work in a direction that is acceptable.

To ease the review process, the first PR shall not address the revision of the unit commitment with multi-units

enricogiglio:
Closes #658 

This feature aims to make the already extendable technologies also modular by introducing an integer variable n_opt, where n represents the number of installed devices (e.g., generators, storages). The optimization process will adjust the optimal capacities to ensure that p_nom_opt = p_nom * n_opt.

Please note that we are currently working on the code with @davide-f, so it is still under development and should not be considered final yet :)

## Checklist


pypsa/linopf.py
```diff
@@ -257,6 +261,74 @@ def define_unit_commitment_status_variables(n, sns, c):
     define_binaries(n, (sns, com_i), c, "status", mask=active)
 
 
+def define_generator_multiple_device(n):
+    c = "Generator"
+    p_nom_i = n.generators.index[n.generators.p_nom > 0]
+    ext_i = get_extendable_i(n, c)
+    if (ext_i.intersection(p_nom_i)).empty:
+        return
```
davide-f, line 269: the calculation of the indices may be done in a clear way probably.
I'd be prone adding a new parameter like "modular" with value True/False (default False) that specifies whether this functionality shall be implemented. However, this is not mandatory and we may use as flag the combination as "extendable" (that would be required anyway) and a non-zero p_nom.
In the following, I'll refer to the possible reformulation of `ext_i.intersection(p_nom_i)` as `mod_i`

regardless, on the choice, the indices mod_i may be calculated using query:
`n.df(c).query("extendable and (p_nom > 0)").index` or `n.df(c).query("extendable and modular").index`

@FabianHofmann or @fneum May you have preferences?
Personally, I'd be prone towards an additional option rather than overloading p_nom and extendable, but not sure what is your opinion.

pypsa/linopf.py
```diff
@@ -257,6 +261,74 @@ def define_unit_commitment_status_variables(n, sns, c):
...
+    ##serve? a mio parere no, quindi silenzio
+    # active = get_activity_mask(n, c, sns)[ext_i.intersection(p_nom_i)] if n._multi_invest else None
+    var = define_integers(n, (ext_i.intersection(p_nom_i)), "Generator", "n_opt")
+    define_boundary_for_multiple_device(n, c, "p_nom", np.array(var))
+    lhs = linexpr(
+        (
+            n.generators.p_nom[ext_i.intersection(p_nom_i)],
```
davide-f, line 276: This can also be restyled to be available for any component using n.df(c)

pypsa/linopf.py
```diff
@@ -1300,8 +1377,13 @@ def map_solution(c, attr):
             n.solutions.at[(c, attr), "pnl"] = False
             sol = variables.map(variables_sol)
             if predefined:
-                non_ext = n.df(c)[attr]
-                n.df(c)[attr + "_opt"] = sol.reindex(non_ext.index).fillna(non_ext)
+                if not (
+                    attr == "n_opt"
+                ):  # if inserito perchè la seguente riga serve solo per far diventare quello che in pypsa viene ottimizzato come nominal optimal, quindi entravamo in conflitto
```
davide-f, line 1382: At first glance, I've the feeling that the need for this if is no more necessary after making the change on variables.csv, but needs testing

pypsa/linopf.py
```diff
@@ -257,6 +261,74 @@ def define_unit_commitment_status_variables(n, sns, c):
...
+    n : pypsa.Network
+    c : str
+        network component of which the nominal capacity should be defined
+    attr : str
+        name of the variable, e.g. 'p_nom'
+    """
+    p_nom_i = n.generators.index[n.generators.p_nom > 0]
+    ext_i = get_extendable_i(n, c)
+    if (ext_i.intersection(p_nom_i)).empty:
+        return
```
davide-f, line 300: revise mod_i as discussed above

pypsa/linopf.py
```diff
@@ -257,6 +261,74 @@ def define_unit_commitment_status_variables(n, sns, c):
...
+    if not (
+        (
+            n.df(c)[attr][ext_i.intersection(p_nom_i)]
+            < n.df(c)[attr + "_max"][ext_i.intersection(p_nom_i)]
+        )
+    ).all():
+        return
```
davide-f, line 307: If you are checking that p_nom is lower than p_nom_max, that is not necessary.
I wouldn't add such constraint: the code would work and should solve anyway, though this generator won't be installed

davide-f (commented):
Hello @enricogiglio, as discussed I'm adding the comments :)
I think we are definitely going in the right direction and we are very close! :)

pypsa/variables.csv
```diff
@@ -4,6 +4,7 @@ Generator,status,False,False,True
 Generator,start_up,False,False,True
 Generator,shut_down,False,False,True
 Generator,p_nom,False,True,False
+Generator,n_opt,False,True,False
```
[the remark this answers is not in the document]
enricogiglio: Done :)

pypsa/linopf.py
enricogiglio: I'll replace it with: mod_i=n.df(c).query('p_nom_extendable and (p_nom>0)').index

If @FabianHofmann and @fneum have different preferences, we can change them accordingly :)

examples/ac-dc-meshed/ac-dc-data/generators.csv
```diff
@@ -1,7 +1,7 @@
...
-Norway Wind,Norway,2184.3747960912,1,0.09,100,True,100,wind
-Norway Gas,Norway,158.2512497168,0.3568363832,5.8928445406,20000,True,0,gas
-Frankfurt Wind,Frankfurt,2129.4561224763,1,0.1,110,True,100,wind
-Frankfurt Gas,Frankfurt,102.6769530076,0.3516658529,4.0863219899,80000,True,0,gas
+Manchester Wind,Manchester,2793.6516029328,1,0.11,0,True,100,wind
+Manchester Gas,Manchester,196.6151679691,0.3500264336,4.5323676307,0,True,0,gas
+Norway Wind,Norway,2184.3747960912,1,0.09,0,True,100,wind
+Norway Gas,Norway,158.2512497168,0.3568363832,5.8928445406,0,True,0,gas
+Frankfurt Wind,Frankfurt,2129.4561224763,1,0.1,0,True,100,wind
+Frankfurt Gas,Frankfurt,102.6769530076,0.3516658529,4.0863219899,0,True,0,gas
```
davide-f, line 7: For Fabians, as later described, the new functionality changes the behavior of what happens when p_nom_extendable is true.
Before when p_nom_extendable was true, the info on p_nom is basically ignored, however this is not the case in this PR: when that occurs, the extendability is enabled but using modularity.
To preserve the old behavior, we need to put p_nom = 0.

To preserve fully the old behavior, we would need a new argument es modularity=True/False, however that was not regarded a priority and we may desire not to add a new option, and the current proposed approach could make sense.

pypsa/linopf.py
```diff
@@ -257,6 +258,41 @@ def define_unit_commitment_status_variables(n, sns, c):
     define_binaries(n, (sns, com_i), c, "status", mask=active)
 
 
+def define_modular_nominal(n, c, attr, attr_nom):
```
davide-f, line 261: @enricogiglio , we need to add a test to actually test this functionality.
Currently, the feature is not tested

pypsa/linopf.py
```diff
@@ -1194,6 +1233,7 @@ def prepare_lopf(
     # consider only state_of_charge_set for the moment
     define_fixed_variable_constraints(n, snapshots, "StorageUnit", "state_of_charge")
     define_fixed_variable_constraints(n, snapshots, "Store", "e")
+    define_modular_nominal(n, c="Generator", attr="n_opt", attr_nom="p_nom")
```
davide-f, line 1236: For Fabians, this is easy to generalize to all components and it is a next step, however, we would like to have a preliminary check to avoid overhead in the next stages

pypsa/linopf.py
```diff
@@ -1300,8 +1341,11 @@ def map_solution(c, attr):
             n.solutions.at[(c, attr), "pnl"] = False
             sol = variables.map(variables_sol)
             if predefined:
-                non_ext = n.df(c)[attr]
-                n.df(c)[attr + "_opt"] = sol.reindex(non_ext.index).fillna(non_ext)
+                if not (attr == "n_opt"):
+                    non_ext = n.df(c)[attr]
+                    n.df(c)[attr + "_opt"] = sol.reindex(non_ext.index).fillna(non_ext)
+                else:
+                    n.df(c).n_opt = sol
```
davide-f, line 1348: This is delicate as it is clearly an ad-hoc fix and relatively difficult to generalize.
Currently, it may be accepted.
Alternatives may involve adding a new column in variables.csv to generalize and handle this exeptional case.

The main difference is the following: this if is on only for p_nom and e_nom that need the ending "*_opt", however with n_opt that is no more the case.
Do you have recommendations? Shall we keep it as is for the moment?

pypsa/optimization/optimize.py
```diff
@@ -330,6 +330,8 @@ def assign_solution(n):
             else:
                 set_from_frame(n, c, attr, df)
         else:
+            if attr == "n_opt":
+                return
```
davide-f, line 334: @enricogiglio this is a refuse to remove

pypsa/variables.csv
```diff
@@ -4,6 +4,7 @@ Generator,status,False,False,True
 Generator,start_up,False,False,True
 Generator,shut_down,False,False,True
 Generator,p_nom,False,True,False
+Generator,n_opt,False,False,True
```
davide-f, line 7: We started from the Generator and as discussed other components will follow once agreed on the structure.
Do you like the "n_opt" name of the variable?

davide-f (commented):
Thanks @enricogiglio for the great work! :)
I think this PR is closing to have a good shape.

In particular, I think it would be nice to have a feedback by @FabianHofmann and/or @fneum .
During the PR, there are some architecture questions that need answers to finalize the PR.

Many thanks!

enricogiglio:
Dear @fneum and @FabianHofmann,

Together with @davide-f, we have nearly completed the initial version of our proposal for this pull request.

In addition to the architectural aspects highlighted by Davide in the previous comments, there is one final aspect primarily related to expanding modularity in linopy. We observed that when simulating in linopy, the functions we have implemented read all solutions as integers, rounding the actual values. Specifically, for the given case (simulated with linopf), these should be the expected outputs related to generators:

Generator | n.generators.p_nom_opt | n.generators.n_opt

coal          |    1000.0                           |          500

coal2        |      500.0                            |          Nan (it has no p_nom defined)

gas           |    1.0                                  |         Nan (it is not extendable)

gas2         |     189.8                             |           130

gas3         |      4309.2                          |          3192

If you simulate the same script with linopy, you will see in the 1st picture the solution file (which corresponds to those obtained with linop). However, as you can see from the 2nd picture, the values assigned as solutions are rounded to integer.

Furthermore, after some debugging, I noticed that the solutions are already registered in m.solution (3rd picture) in their rounded form (whereas the n_opt are correct, with respect to the simulation done in linopf).

In summary, the equations seem to be correct, as well as the solutions found by the solver. The problem lies in the values assigned after the solution process, as they are rounded to integer values. I would appreciate any advice on how to proceed with this problem :)

MicrosoftTeams-image (23)

MicrosoftTeams-image (24)

MicrosoftTeams-image (25)

davide-f:
@enricogiglio as discussed, this is a great job :)

@FabianHofmann and/or @fneum I believe this PR is ready to be discussed on an architecture level; I'm fully available to have a call and discuss on it.

Enrico also found an interesting bug that in my opinion is related to linopy.
As shown in the pictures above, it seems like when integer variables are added the whole dataset that stores the data are somehow converted into int format that truncated the fractional results.

fneum:
See also https://github.com/PyPSA/PyPSA/pull/372.

davide-f:
> See also #372.

Thanks Fabian for the comment, yet this PR is not a duplication because:
1. it enables integer investment capacities: the new investment p_nom_opt shall be an integer multiple of p_nom (or other parameter)
2. it extends the formulation for all components, not only Generators

The current formulation does not account for retiring the units, though I am unsure this is a wanted feature.
I see the problem of modularity for new installations, but the retiring may be more questionable
@enricogiglio, could you revise the title and the description of the text to avoid misunderstanding?

@fneum , what do you think? Happy to have a 5-min chat in case of need.
I feel that from a technical perspective, modularity is something welcome by many users and can play a big impact in various applications

pypsa/variables.csv
davide-f: we just discussed with fabian to name the variable as "n_mod"

pypsa/optimization/optimize.py
```diff
@@ -330,7 +338,10 @@ def assign_solution(n):
             else:
                 set_from_frame(n, c, attr, df)
         else:
-            n.df(c)[attr + "_opt"].update(df)
+            if attr == "n_opt":
+                n.df(c)[attr] = df
+            else:
+                n.df(c)[attr + "_opt"].update(df)
```
davide-f, line 344: Let's keep the formulation as before and have the optimal number of components named "n_mod_opt" (n_mod + _opt).
To do so, we just need to keep the original formulation

pypsa/optimization/constraints.py
```diff
@@ -624,6 +624,42 @@ def define_fixed_nominal_constraints(n, c, attr):
     n.model.add_constraints(var, "=", fix, f"{c}-{attr}_set")
 
 
+def define_modular_constraints(n, c, attr, attr_nom):
+    """
+    Sets constraints for fixing modular variables of a given component. It
+    allows to define optimal capacity of a component as multiple of the nominal
+    capacity of the single module.
+
```
davide-f, line 632: Instead of overloading "p/e_nom" let's define a new parameter "p/e_nom_mod" that is used to identify the size of the modular component.
That requires little changes in the code below and the addition of new parameters to be loaded during the add(...) function

examples/ac-dc-meshed/ac-dc-data/generators.csv
davide-f: As discussed with fabian, we shall preserve the previous p_nom behaviour.
As such, the test csv files shall be restored as in the original formulation

pypsa/component_attrs/generators.csv
```diff
@@ -34,6 +34,7 @@ weight,float,n/a,1,"Weighting of a generator. Only used for network clustering."
 p,series,MW,0,active power at bus (positive if net generation),Output
 q,series,MVar,0,reactive power (positive if net generation),Output
 p_nom_opt,float,MW,0,Optimised nominal power.,Output
+n_opt,int,n/a,0,Optimised number of generators.,Output
```
davide-f, line 37: As discussed the optimized variable shall be named "n_mod_opt" as output.
Morever, we need a new input parameter "p_nom_mod" that is the size of each module.
The logic shall be equivalent with the one we proposed: "if 0, we preserve the continuous formulation, if > 0, we implement modularity"

davide-f (commented):
Thanks @FabianHofmann for the time for the review and comments.
@enricogiglio I added some comments for moving towards the final implementation, in agreement to the discussion had with fabian :)
Let's coordinate, also in agreement with the holidays (please, take them :) )

enricogiglio:
Hello :)

The test is giving us problems with "test_multiinvest", but when we tested it offline (see fig below) everything seemed fine. Any thoughts on what might be the problem?

MicrosoftTeams-image (27)

davide-f:
@FabianHofmann Hello :)
The changes you proposed are now implemented.

It is quite surprising though that the CI is not working.
Note:
- the cI with micromamba succeeds, but the otherone fails due to some improper name in the multiinvest test
- locally, the pytest executed by @enricogiglio succeeds and we cannot reproduce the issue.

To me, the PR looks clean.

Before wasting too much time, have you already encountered this issue?

pz-max:
Some ideas on why just parts are failing?
@fneum maybe good to rerun this PR here as it was an old run

enricogiglio:
Hi @FabianHofmann :) Now it seems to work, could you review it? Do we need to change anything for the codecov problem (only -0.01%)?

enricogiglio:
Absolutely, and thank you :)

FabianHofmann:
hey @enricogiglio, I pushed some changes to have a leaner code and implementation. I would be in favor of removing the code changes in linopf/linopt as these modules are deprecated. Are you relying on them?

enricogiglio:
> hey @enricogiglio, I pushed some changes to have a leaner code and implementation. I would be in favor of removing the code changes in linopf/linopt as these modules are deprecated. Are you relying on them?

Hi @FabianHofmann ! If you plan to remove linopt and linopf in the next few weeks, we understand that it makes sense to remove the code changes. Otherwise, we would prefer to keep them. Thank you very much!

FabianHofmann:
@enricogiglio I adjusted the linopf code to the more minimalistic layout. I will merge now, thanks for the contribution!

PR #669, opened 2023-07-06
handle multilinks on the fly

### LINKED ISSUE #545: By default, have multilinks in pypsa networks
options:

- create columns and semantics on the fly
- n = pypsa.Network(multilinks=3)

fneum:
closes #545 

With this PR, one no longer needs to define `busX`, `efficiencyX`, `pX` for multilinks at network construction using override components. Now, these fields are handled automatically.

I am not 100% convinced that there is not a more clever way to do it. Maybe @FabianHofmann has suggestions for style?

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,10 @@ n.optimize()
 # plot results
 n.generators_t.p.plot()
 n.plot()
+
+# get statistics
+n.statistics()
+n.statistics.energy_balance()
 ```
 
 There are [more extensive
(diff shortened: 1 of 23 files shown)


FabianHofmann (commented):
I like the approach! But wouldn't another place for updating the component attrs suit better than `additional_linkports`? Perhaps `calculate_dependent_values` or `determine_network_topology`?

fneum:
Ok, I think this is better now, minimising the calls to update the component attributes. I added it to `n.calculate_dependent_values()` but one needs to keep it in the `n.add()` function as otherwise the bus2, p2 etc. are ignored.

FabianHofmann (approved):

PR #677, opened 2023-07-11
clustering: follow-up, renenable strategies of non-default columns

FabianHofmann:


FabianHofmann:
Appeared at a later stage in pypsa-eur

PR #679, opened 2023-07-13
Disable Compression per default

FabianHofmann:
Is up to 20 times faster for large networks (while using 3 times more space)


fneum:
Failed tests unrelated to PR.

PR #698, opened 2023-07-25
Make HiGHS default solver

fneum:
`highspy>=1.5.3` is available for all of linux, macOS and windows.

Is that enough?


fneum:
https://github.com/Pyomo/pyomo/issues/2094#issuecomment-1007572933

may have to  be "appsi_highs" for pyomo

fneum:
I think it's better to only switch to HiGHS as default solver in linopy API.

PR #706, opened 2023-08-02
linopt: Fix for the use of solver_options for cbc & glpk

lukasnacken:
Closes # (if applicable).

This pull request fixes the construction of the solver command for the solvers cbc and glpk in the module linopt. 
I received the error when testing PyPSA-Earth on a machine that does not yet have access to Gurobi. The correction is not very relevant for the future, as linopy.solvers.run_cbc will replace pypsa.linopt.run_and_read_cbc.

## Checklist


FabianHofmann:
thanks @lukasnacken, I remember we did not put it this way as there are also options which require `--` in front. But I don't mind changing it. Perhaps make a case distinction, if solver_options is a string -> old behavior, if is a dict -> new behavior.

fneum:
xref https://github.com/PyPSA/PyPSA/pull/399

lukasnacken:
Thanks for the feedback! Okay, now I understand the old code and I added the case distinction.

Should the possibility to define solver_options as a string remain a hidden feature or are you aware of several users who prefer to define solver_options as a string?
If the latter, one would have to 
1) generalise the case distinction for all solvers in linopt and linopy as suggested by @Cellophil in #399 
2) update the type hint in the doc-strings of the top-level functions

fneum:
Since linopt is going to be phased out at some point, let's keep string option formatting a "hidden feature".

Do you have some code to test your contribution @lukasnacken?

lukasnacken:
@fneum, sure, here are a few lines to test the contribution. 
```python
import pypsa

configs = [("cbc",  "-threads 4 "), 
           ("cbc",  {"threads": 4}),
           ("glpk",  " --interior")]

for name, options in configs:
    n.lopf(solver_name=name, solver_options=options)
```

I have not changed the code for the "string"-way. That's why the whitespaces are different. 

I noticed that glpk doesn't just ignore solver options that are not supplied. When glpk is used with unavailable solver options, it takes forever to solve.  Therefore, I have added a note as logger info.

fneum (approved):

PR #709, opened 2023-08-15
Add functionality to store geometries with buses, lines, and links

### LINKED ISSUE #708: GeoPandas support in PyPSA

Especially for plotting, it is helpful to carry geo-information around.
It would be nice if some of the PyPSA dataframes are geodata frames "if" the geometry is available

E.g. `n.buses` could carry optional columns "geometry" (with points or polygons)
and `n.lines` and `n.links` could carry an optional column "geometry" (with line strings)

fneum:
and transformers.

closes #708 

This PR is up for debate and definitely needs rigorous testing before being merged.

**Pro:** I think it may be useful for plotting data regarding the bus' associated regions. Based on the suggested changes, one could further extend the `n.plot` function to support choropleth maps for example and line strings for lines and links that do not follow a straight line.

**Con:** Adds `geopandas` and `pygeos` as dependencies.

I am also open to ideas on how to implement this in a more beautiful way.

TODO

- [ ] two seemingly unrelated tests mysteriously fail
- [ ] move code duplication into utility function


coroa:
Hmm .. the negative effects could be mitigated by making the geopandas availability conditional, I guess, ie. introducing `has_geopandas` for instance.

`pygeos` has been integrated into `shapely` 2 and only exists for backwards compatibility. don't think you need that as a dependency anymore.

pz-max:
Very cool! What about we store a geodataframe at `n.buses.shapes`. This would allow the `n.buses` can stay a pandas dataframe. But making it conditional as @coroa mentioned sounds also good.

davide-f:
Hello!
That's an interesting feature and my feeling is the following:
- the geometry may be extended to all components potentially. In some cases, the geometry for buses could describe the region of the substation, similarly for generators. In case of hydro storage_units, that may be interesting to store the shape of the reservoir as well. But this may be a second-level detail.
- On a usability perspective, forcing pypsa to rely mandatority on gpd may have some limitation when the users edit the (geo)dataframes by themselves. Currently, the damage is relatively limited, but modifying the dataframe where info on the geometries are more likely to create issue for beginners. Therefore, I think keeping the old behavior may be a good practice.
- Shapes could also be stored as strings into the meta part. I think that we may group the potential needs of users in having the shape integrated and then identify the best approach. If the goal is to let user have their shapes integrated, with the meta part that is already possible. I'm wondering if the additional plotting features may be an optional new package or component that automatically interacts with the meta structure. If someone may be interested into that feature, if geometries are stored in the meta or another component, then it may be easy to create geodataframes without touching the core behavior of the algorithm.

My feeling is that using geodataframes may lead to more breaking and conflics and as fabian mentioned, it requires a strong testing. If it was me, I would collect ideas from users on the new features and plottings that could be available with that change and then decide the best approach on:
a) integrate geometries using geopandas
b) integrate geometries using meta or alike
c) a hybrid approach where the core algorithm is not affected but during the plotting phase, upon request, geometries from meta or alike are retrieved to create the geodataframes needed for the features for example

The use of b and c may be a soft step to better understand the actual needs and the potential advantages and disadvantages.
Definitely feel free to disagree :)

pz-max:
Given the [pandapower example](https://github.com/e2nIEE/pandapower/blob/master/tutorials/plotting_basic.ipynb) they store geo data in e.g. `net.bus_geodata` and `net.line_geodata`. One example here: ```coords = zip(net.bus_geodata.x.loc[buses].values, net.bus_geodata.y.loc[buses].values) # tuples of all bus coords```. I think we can build something similar. We need a PyPSA utility function in for **saving and reading** the geojson as string in the netcdf as well as plenty test. More below on read and write examples below:

**Store a GeoPandas DataFrame into a NetCDF file:**
1. Converting GeoDataFrame to xarray Dataset: GeoPandas DataFrame can be converted into an xarray.Dataset which can then be easily written to a NetCDF file.

2. Storing geometry information: Geometry data in a GeoPandas DataFrame, especially polygons and lines, don't have a direct native representation in NetCDF. This means you'll have to come up with a strategy to serialize and store geometry information. A common approach is to convert geometries to Well-Known Text (WKT) format, store them as strings, and then reconvert them back when reading.

Below just an example how read and write would look like:

Install dependencies: 
```python
pip install geopandas xarray netCDF4 shapely
```

**1. Conversion of GeoDataFrame to NetCDF:**
```python
import geopandas as gpd
import xarray as xr

# Sample GeoDataFrame
gdf = gpd.read_file(gpd.datasets.get_path('nybb'))

# Convert geometries to WKT format
gdf['geometry_wkt'] = gdf['geometry'].apply(lambda geom: geom.wkt)

# Convert the GeoDataFrame to a Pandas DataFrame and then to an xarray Dataset
ds = xr.Dataset.from_dataframe(gdf.drop(columns='geometry'))

# Save to NetCDF
ds.to_netcdf('data.nc')
    Install the necessary libraries:
```

**2. Reading from NetCDF and converting to GeoDataFrame:**

```python
import geopandas as gpd
import xarray as xr
from shapely import wkt

# Load the data from NetCDF
ds = xr.open_dataset('data.nc')

# Convert the xarray Dataset to a Pandas DataFrame
df = ds.to_dataframe()

# Convert the WKT strings back to geometry objects
df['geometry'] = df['geometry_wkt'].apply(lambda geom: wkt.loads(geom))

# Convert the Pandas DataFrame to a GeoDataFrame
gdf_from_nc = gpd.GeoDataFrame(df, geometry='geometry', crs="EPSG:4326")

print(gdf_from_nc)
```

fneum:
I think the solution to store geometries separately as a `gpd.GeoSeries` has some appeal.

@pz-max the conversion from/to WKT strings is already implemented in this PR.

pz-max:
Just talked with @drifter089 about the topic. I think @fneum  implemented **Method 1**. We prefer Method 1 mapped below instead of the pandapower approach. Next steps:
1. Repair PyPSA CI
2. Finalise Fabians PR (e.g. remove pygeos, expand on documentation and tests)

**Method 1 - Geopandas per default** (when no geometry exists geopandas aims to fallback to pandas: https://github.com/geopandas/geopandas/pull/2060)
**Access example**
n.buses.x  (still accessible)
n.buses.geometry  # polygons or Points NOT both possible for now
n.buses.plot(column="v_nom")  # nice plot with voltage level coloured per region
n.lines.geometry
...
**Pro/Con**
+ Easy compatible with existing code
+ Easy access as in the same dataframe
+ Works with multiple geometries
-- Risk of different UX when dealing with geopandas instead of panda data frames
-- Maybe requires has_geopandas option / need to deal with geopandas stability

**Method 2 - Pandapower approach**
**Access example**
n.bus_geodata  # contains index of n.buses
n.line_geodata
...
**Pro/Con**
+ Easy compatible with existing code
+ Works with multiple geometries
-- Data is not at one place/ less accessible
-- Automatic updating. E.g. row deletion in n.buses_geodata required if a row in e.g. n.buses is deleted (might end up as a bigger issue than expected)

fneum:
closing in favour of #730

PR #726, opened 2023-09-09
repair CI

pz-max:
Build on PR https://github.com/PyPSA/PyPSA/pull/724 but adding pandas constraint. 
We probably have to delve into the pandas warning to use the newer pandas versions:
```
FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas.
Value '[-1.34291453e-13 -9.58966072e-14]' has dtype incompatible with int64,
please explicitly cast to a compatible dtype first.
    network.buses_t.p.loc[snapshots, sn_buses] = s_calc.real[
```
Collab with @drifter089 on this PR.

## Checklist

### DIFF
--- a/.codespell.ignore
+++ b/.codespell.ignore
@@ -1,3 +1,4 @@
 ines
 psace
 datas
+fom
--- a/environment.yaml
+++ b/environment.yaml
@@ -5,14 +5,14 @@ dependencies:
 - python>=3.9
 - numpy
 - scipy
-- pandas>=0.24.0
+- pandas>=0.24.0,<2.1
 - xarray
 - netcdf4
 - libnetcdf
 - pytables
 - matplotlib
 - networkx>=1.10
-- pyomo>=5.7.0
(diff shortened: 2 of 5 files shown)


fneum (approved):
Not really a fix since constraining pandas version, but temporary solution.

maurerle:
Is this still needed? @pz-max 
For python 3.12 the first pyomo release supported is 6.7 which is not installable with the latest pypsa version 0.26.3 - due to the pyomo version limit.

This results in conda installing an older pypsa version (0.25.1) which is unsupported in other ways..

So it would be good to have a pypsa release which supports pyomo 6.7 :blush:

fneum:
I agree. Support for Pyomo 6.7 would be good. We're happy to receive a PR that restores compatibility, but our current plan is to phase out maintenance of the `pyomo` backend in favour of `linopy` and eventually remove it completely.

PR #730, opened 2023-09-12
geopandas support for components

drifter089:
I've introduced modifications to enhance geopandas support in PyPSA. The updates are minor tweaks to the functions originally developed by @fneum, and I've also added tests for geopandas integration. A big shout-out to @pz-max  for his in-depth insights on PyPSA and pytest that guided this contribution.


test/test_components.py
```diff
@@ -15,6 +18,41 @@ def empty_network_5_buses():
...
+    assert isinstance(geo_components_network.links.geometry, geopandas.GeoSeries)
+    assert isinstance(geo_components_network.transformers.geometry, geopandas.GeoSeries)
+
+
+def test_geo_component_add():
+    # Test whether 'geo_components' can be added to the network by add function.
+    # geo_components = {"Bus", "Line", "Link", "Transformer"}
+    # Point(0,1) and 'POINT (1 4)'(wkt_string) are valid and will be converted to shapely.geometry objects.
+    # {"None", "", np.nan, "nan"} are also valid values for geometry column while adding using add function and will be added as None.
+    # note - invalid strings will also be added as None.
```
pz-max, line 41: Use docstrings as applied in other cases

pz-max: Also consider the PEP convention: https://peps.python.org/pep-0008/#documentation-strings

test/test_components.py
```diff
@@ -15,6 +18,41 @@ def empty_network_5_buses():
...
+
+    network = pypsa.Network()
+
+    # {"None", "", np.nan, "nan"}
+    network.add("Bus", "bus_1", geometry="")
```
pz-max, line 46: Here you only test the bus geometry.
It needs to work for bus/line/links/transformers etc..

Think about also how to test not only for POINT coordinates but also for LINESTRING and POLYGONS (MULTI LINESTRING AND MULTI POLYGON should be forbidden -> see here the[ geometric objects](https://geopandas.org/en/stable/docs/user_guide/data_structures.html#geoseries) of geopandas )

test/test_components.py
```diff
@@ -15,6 +18,41 @@ def empty_network_5_buses():
     return network
 
 
+def test_geo_components_gpd_df(geo_components_network):
+    # Test whether 'geo_components' are 'geopandas.DataFrame' and conatin a 'geopandas.GeoSeries' in the 'geometry' column.
+    # geo_components = {"Bus", "Line", "Link", "Transformer"}
```
pz-max, line 23: change to docstring with pep convention

test/conftest.py
```diff
@@ -45,6 +45,40 @@ def scipy_network():
...
+@pytest.fixture(scope="function")
+def geo_components_network():
+    # Constructing a basic network using all 'geo_components'.
+    # geo_components = {"Bus", "Line", "Link", "Transformer"}
+    network = pypsa.Network()
+    num_components = 5
+
+    for i in range(num_components):
+        network.add("Bus", f"bus_{i}")
+
```
pz-max, line 57: Can you not avoid this and use of of the other networks e.g. ac_dc_network or scipy_network?

pypsa/io.py
```diff
@@ -901,6 +922,10 @@ def import_components_from_dataframe(network, dataframe, cls_name):
         logger.error("Error, new components for {} are not unique".format(cls_name))
         return
 
+    if cls_name in network.geo_components:
+        new_df["geometry"] = new_df["geometry"].apply(clean_geometry)
+        new_df = gpd.GeoDataFrame(new_df, geometry="geometry", crs=network.srid)
```
pz-max, line 927: just out of curiousity. What is `network.srid` ??

pz-max (changes requested):

pz-max:
Extends: https://github.com/PyPSA/PyPSA/pull/709

drifter089:
After executing all the tests, only the test in test_lopf_multiinvest.py is failing.
Screenshot 2023-09-19 194120

drifter089:
When I manually run the test with api = 'linopy', the gpd version produces different values for n.lines.s_nom_opt.
image

drifter089:
the data variable values are different in both cases in the solved model. I am trying to find out how the Linopy model being sent to solve differs, or if Linopy is producing different results.

Screenshot 2023-09-21 234535
Screenshot 2023-09-21 234508

test/test_components.py
drifter089: why should MULTI LINESTRING AND MULTI POLYGON should be forbidden?

test/conftest.py
drifter089: Implementing this to streamline testing and eliminate the need to repeatedly add geo components. This will also simplify the addition of future Geopandas-related tests.

pypsa/io.py
drifter089: image

drifter089:
I've identified a problem with the Linopy Model when working with a geopandas dataframe. When executing df.length, geopandas returns NaN values, which leads to incorrect constraints in the Linopy model.
Screenshot 2023-09-30 203446
Screenshot 2023-09-30 203436

To address this, we should explore alternative methods for retrieving dataframe indexes. Additionally, we need to verify if the reindex function is causing similar issues in other scenarios.

pz-max:
We are currently restructuring/redesigning the code.
**GOAL:** Make this backwards compatible. Meaning people without geometry won't be bothered by having to deal with GeoPandas

We try to consider these two statements:
- "When adding geometry data to PyPSA, it's ok to make it explicitly by requiring the user to convert the previous pandas DataFrame to Geopandas GeoDataFrame"
- "If all geometries are NONE, drop geometry columns. This will convert Geopandas GeoDataFrame to Pandas DataFrame"

drifter089:
In order to make this backwards compatible or not cause any breaking changes, we can:

1. Add a separate `GeoNetwork` function.
2. Or add a parameter `Geo = True/False` in the existing `network` function.

For instance:
```python
n = pypsa(path, Geo=True)

Figuring out how `io` functions and component attributes in the `csv` will look like in this case.

**Comment by drifter089** [comment]
**Keeping backwards compatibility** is crucial, ensuring no breaking changes. After considering various approaches, I believe the best one is to introduce a `enable_geometries` function as a method of the `Network` class.

```python
n = pypsa.Network()
n.enable_geometries()
n.revert_geometries()

**Comment by drifter089** [comment]
-  This function abstracts away the current functionality of converting into geopandas df and defining geo_components coded in _build_dataframes.
- By default Network will have geo_components as {} and no geopandas df.
- The add method can be modified to account for adding geometry values only when geo_components is not null, so can be any other method of the network class.
- This approach eliminates the need to modify the component attribute CSV. For instance, in the case of a bus, x and y parameters are already present, which might lead to confusion.
- It offers the flexibility to convert geopandas df back to pandas df without information loss.
- We can gradually introduce support for geometries in existing methods and incorporate new methods with extra functionality.

Please share your thoughts.

**Comment by pz-max** [comment]
I currently don't understand why the existing version is not backwards compatible...
I am now downloading the code locally to test everything and eventually create a PR at your repo.

```
git clone https://github.com/pz-max/PyPSA.git
cd PyPSA
git remote add drifter089 https://github.com/drifter089/PyPSA
git fetch drifter089
git checkout -b geopypsa drifter089/akshat_dev
```

File: pypsa/components.py [hunk 6]
```diff
@@ -391,6 +403,10 @@ def _build_dataframes(self):
                 columns=static_dtypes.index,
             )
 
+            if component in self.geo_components:
```
drifter089, line 406: @pz-max  this is the code isn't backward compatible because we create dataframes for bus links when initializing a network, and here we convert them to GeoPandas dataframes.

drifter089: We can abstract this into a separate method, allowing us to easily convert between standard dataframes and GeoPandas as needed.

pz-max: Got it! :+1:

pz-max:
> Please share your thoughts.

I now understand your approach and very much like it. 

I am looking forward to read the **test cases** with:
n.enable_geometries()
n.revert_geometries()

Btw I just remembered that we noticed with Fabian that crs information needs to be stored e.g. in n.meta. Because these are not automatically dragged along in the WKT string. --- But let's address that io stuff later

FabianHofmann:
Hey @pz-max, thanks for tuning me ;) As you know I like the idea. There, is some things to consider: 

- Storing more than one geometry per component: In some cases, it is necessary to store multiple geometries for a single component. For example, in pypsa-eur, we have both onshore and offshore shapes for single buses.
- Storing non-component related geometries: In other cases, there are important geometries which are not directly associated with single components, like in pypsa-eur the shape of whole Europe which we want to store.  
- A non-automatic process with enforces the user to use a flag like `n.enable_geometries()` feels a bit ponderous. But that's just a minor point.  
- Type stability: Optionally using geopandas dataframe instead of pandas dataframes can lead to unexpected behavior in some cases. Examples are :`n.buses.plot` leads to different figures, then with a geopandas dataframe; `n.buses.area` is different from `n.buses["area"]` if you have a column "area" in the buses dataframe. (@fneum knows that quite well ;) ). I have the concern that there are some functionalities, that we don't test which won't work, like clustering, pf etc.   

Let me know what you think.

pz-max:
Hi @FabianHofmann , great points!

Let me just summarise for others what we are talking about. From UX experience, the question is whether we should store GIS information directly at the components Dataframe, e.g. `n.buses` would also have geometry information per row, **or** store all GIS information in a separate dataframe `n.geometries` which contains all geometries from line to bus to EU shape etc.

Discussing your points:
1. Storing more geometries per component is possible either with "multi polygons" or with storing multiple geometry columns "geometry", "geometry_2". The latter "geometry_2" of course, needs some extra attention for the plotting scripts if other shapes are available and should be generally avoided if we can do so.
2. Agree. For storing any other geometries like continental shapes, having something like e.g. "n.geometries" is absolutely useful in any case.
3. Hm, with `n.enable_geometries()` any future IO will be automatically geopandas. We possibly could also make it `pypsa.network(geomtry=True)` or similar.
4. Regarding type stability.  Plots will in any case require geomtries checks. E.g. if geometries exist also use them. We already recognised some issues with columns like areas / length etc in the PyPSA jungle and indeed some parts need to be renamed at some locations. Pytests that continuously check stability can help spot issues before they occur (and they already worked well for this purpose).

In conclusion.
- I think we need something like `n.geometries` anyway given your perspective
- As of the above, it will be either a **pure** `n.geometries` solution or a **hybrid** solution where additional `n.buses` can carry geometry information
- For any of the solutions, we also need to deal with storing crs data somewhere
- We should meet together and discuss opportunities and challenges for each approach (& then definitely decide on which approach we should test first -- we can also test one method first & pivot to the other approach if necessary). I don't have hard feelings about this but just see UX benefits for the **hybrid** approach (no filtering required to access geometry of a particular line/bus/transformer object) & @drifter089 sees technical stability benefits when data is duplicated e.g. `n.buses` would have an index with geomtries in `n.geometries`. 

When should we meet up ideally? Ping on Discord about the details in PyPSA meets Earth Server

fneum:
superseded by #759

PR #734, opened 2023-09-14
address all-zero time-dependent value removal bug (closes #722)

### LINKED ISSUE #722: Export bug when time-dependent variable is all 0.
## Checklist

- [x] I am using the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch.

## Describe the Bug

When a time-dependent variable, in this case n.links_t.p2 is all 0., the whole table is dropped upon export, here is the line of code where it happens:

https://github.com/PyPSA/PyPSA/blob/a0027f50a24744251e58ea6577f446bfdc90f1f6/pypsa/io.py#L504

I.e. if all columns equal the default, the whole table is dropped. Which leads to results disappearing.

## MWE

```py
!wget https://model.energy/networks/c6c1d08a805dbbeae283ce07a1195934.nc
import pypsa
n = pypsa.Network("c6c1d08a805dbbeae283ce07a1195934.nc")
(n.links_t.p2 != n.components["Link"]["attrs"].at["p2","default"]).any()
n.links_t.p2["hydrogen_electrolyser"] = 0.
n.export_to_netcdf("c6-test.nc")
n = pypsa.Network("c6-test.nc")
n.links_t.p2
n.links_t.p2["hydrogen_compressor"] = 0.
n.export_to_netcdf("c6c-test.nc")
n = pypsa.Network("c6c-test.nc")
n.links_t.p2
```

## Solution

Should not actually drop anything, i.e. save the file exactly as the user expects, then let the compression algorithm worry about saving disk space.

fneum:
Still have to check how much it affects file size.

### DIFF
--- a/pypsa/io.py
+++ b/pypsa/io.py
@@ -501,11 +501,8 @@ def _export_to_exporter(network, exporter, basename, export_standard_types=False
                 else:
                     col_export = pnl[attr].columns[(pnl[attr] != default).any()]
 
-            if len(col_export) > 0:
-                df = pnl[attr].reset_index()[col_export]
-                exporter.save_series(list_name, attr, df)
-            else:
-                exporter.remove_series(list_name, attr)
+            df = pnl[attr].reset_index()[col_export]
+            exporter.save_series(list_name, attr, df)
 
         exported_components.append(list_name)
 


coroa:
Fixes #722

coroa:
I agree this makes sense for the netcdf representation, where there is compression (which is now enabled by default), but there are other exporters like to csv where this has a big impact (not only in storage space, but also in file space, which incidentally is also where this branch was originally implemented and included into the exporter/importer infrastructure).

fneum:
Xref https://github.com/PyPSA/whobs-server/commit/dbc6726e600dce7dbd62f7b7f0ac2871df0990c5

lkstrp:
I think I fixed that in #1128. This can be achieved in #1128 by adding dynamic attributes, which are generally the same as static attributes, or only io. This could also be a simple option. I need to revive that PR anyway, so I'll move the discussion there

FabianHofmann:
I would close this as this is already tracked as an issue linked to the data validation work of @lkstrp

PR #755, opened 2023-10-13
Add "Shape" component

### LINKED ISSUE #708: GeoPandas support in PyPSA

FabianHofmann:
closes #708
closes #759 
closes #766  

## Checklist


FabianHofmann:
Hey @drifter089, this is an alternative approach to #730, which is less invasive and purely additive (yet full backwards-compatible). Could you have a look whether this works for you? We could optionally extend the consistency_check function, by checking that 

1. the `component` column contains valid component names
2. the `idx` columns contains is a subset of the respective component 

We should also think about deprecating the `srid` attribute.

pz-max:
@FabianHofmann just a flag. We are currently trying something new out in the previous PR.
In this solution the PyPSA .nc container remains a pandas dataframe per default and only if someone sets single time `n.enable_geometries` it will become a geopandas dataframe. One can also drop the geopandas default by e.g. `n.disable_geometries`

pypsa/component_attrs/shapes.csv
```diff
@@ -0,0 +1,6 @@
+"attribute","type","unit","default","description","status"
+"name","string","n/a","n/a","Unique name","Input (required)"
+"geometry","geometry","n/a","n/a","Geometry","Input (required)"
+"component","string","n/a","n/a","Component","Input (required)"
+"idx","string","n/a","n/a","Name of the component. Should be in component's index","Input (required)"
+"type","string","n/a","n/a","Type of geometry","Input (required)"
```
fneum, line 6: Are really all required, or are some optional?

environment.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
 - networkx>=1.10
 - pyomo>=5.7.0,<6.6.2
 - cartopy>=0.16
+- geopandas>=0.9
```
fneum, line 17: Is there any downside to having `geopandas` as compulsory dependency?

pypsa/components.py
```diff
@@ -387,10 +390,17 @@ def _build_dataframes(self):
...
 
-            df = pd.DataFrame(
-                {k: pd.Series(dtype=d) for k, d in static_dtypes.items()},
-                columns=static_dtypes.index,
-            )
+            if component == "Shape":
+                df = gpd.GeoDataFrame(
+                    {k: gpd.GeoSeries(dtype=d) for k, d in static_dtypes.items()},
+                    columns=static_dtypes.index,
+                    crs="epsg:4326",
```
fneum, line 397: This should probably be taken from `n.srid` or the new `n.crs`, rather than implicitly assuming a projection. Then the default of these should be "EPSG:4326".

pypsa/components.py
```diff
@@ -465,6 +475,13 @@ def meta(self, new):
             raise TypeError(f"Meta must be a dictionary, received a {type(new)}")
         self._meta = new
 
+    @property
+    def crs(self):
+        """
+        Coordinate reference system of the network.
+        """
+        return self.shapes.crs
```
fneum, line 483: This overlaps with `n.srid`, which is essentially the same. What do we do with this? Deprecate `n.srid` and map `n.crs` onto it in the mean time?

I am wondering what I have to do to change the projection of my `n.shapes`.

Probably, `n.shapes = n.shapes.to_crs("3035")` would work.

Maybe, a utility function to do the projection of `n.shapes` automatically if you say one of:

```py
n.crs = 3035

n.crs = "EPSG:3035"

crs = ccrs.EqualEarth()
n.crs = crs

crs = ccrs.EqualEarth()
n.crs = crs.proj4_init
```

pypsa/io.py
```diff
@@ -870,8 +877,14 @@ def import_components_from_dataframe(network, dataframe, cls_name):
         else:
             if static_attrs.at[k, "type"] == "string":
                 dataframe[k] = dataframe[k].replace({np.nan: ""})
-
-            dataframe[k] = dataframe[k].astype(static_attrs.at[k, "typ"])
+            if dataframe[k].dtype != static_attrs.at[k, "typ"]:
+                if static_attrs.at[k, "type"] == "geometry":
+                    crs = network.meta.pop("_crs", None)
```
fneum, line 882: Why do we need to go through `n.meta` to retrieve the CRS? Couldn't we just store `n.crs` with the exported files?

setup.py
```diff
@@ -31,6 +31,7 @@
         "pyomo>=5.7,<6.6.2",
         "linopy>=0.2.1",
         "matplotlib",
+        "geopandas",
```
fneum, line 34:
```suggestion
        "geopandas>=0.9",
```
add version constraint here as well?

environment_dev.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
 - networkx>=1.10
 - pyomo>=5.7.0,<6.6.2
 - cartopy>=0.16
+- geopandas>=0.9
```
fneum, line 19: Should `shapely` be added here for the development environment, since it is used in unit tests?

fneum (commented):
This looks very nice and minimally invasive. I had a few comments about the

- potential drawbacks of geopandas dependency
- the relation of `n.srid` and `n.crs`
- a potential setter function for `n.crs` which automatically reprojects `n.shapes`

fneum:
One thing from #759, which I am not sure one needs, could also be suitable for this PR.

```py
def clean_geometry(geom):
    null_types = {"None", "", np.nan, "nan"}
    if geom in null_types:
        return None
    if isinstance(geom, shapely.Geometry):
        return geom
    try:
        return wkt.loads(geom)
    except:
        logger.warning(f"{geom} is invalid string adding geometry value as null.")
        return None

File: pypsa/io.py [hunk 8]
```diff
@@ -22,9 +22,11 @@
 from pathlib import Path
 from urllib.request import urlretrieve
 
+import geopandas as gpd
 import numpy as np
 import pandas as pd
 import validators
+from pyproj import CRS
```
fneum, line 29: add `pyproj` to dependencies?

fneum:
One more thing that should be checked:

- [ ] what happens if you try to import a network with `n.shapes` with an older PyPSA version.

drifter089:
Hey @FabianHofmann and @fneum ,
I've implemented the consistency check and addressed the feedback from the comments. It would be appreciated if you could review and suggest any further changes or improvements on #766.
tests are failing for now working to fix that

environment.yaml
FabianHofmann: not sure, but since we are already having cartopy, I'd say that's a minor addition

environment_dev.yaml
FabianHofmann: already included in geopandas dependencies

pypsa/components.py
FabianHofmann: One problem I see at the moment, is that srid is actually referring to the coordinates of n.buses. So, in the current implementation, crs is referring to the projection of the geometries only. I am updating the docstrings accordingly. For later (in another PR), I would propose to make a smooth transition to replace srid & crs by crs only which then refers to geometries and coordinates. When setting the crs, the coordinates of the buses should then be updated as well. Does this make sense?

pypsa/io.py
FabianHofmann: already in geopandas

pypsa/io.py
FabianHofmann: The `crs` attribute is actually a network property which need some data reference, I found this the cleanest solution (alternatively, one could create an intermediate n._crs object, which then also has to be deleted after loading (if it is there))

pypsa/components.py
fneum: Sounds good! The `@setter` function for reprojection would still be nice.

pypsa/io.py
fneum: This should be possible to do analogous to `n.srid`, right?

FabianHofmann: Not really, since `n.crs` is a `property` which does not have a setter.

FabianHofmann:
> One more thing that should be checked:
> 
> * [x]  what happens if you try to import a network with `n.shapes` with an older PyPSA version.

tested, works without problems (does just not load the shapes)

FabianHofmann:
Okay, I think we are there. There is a `crs` setter now, a consistency check and nearly full test-coverage.

FabianHofmann:
@drifter089 I took the freedom to copy some things from your PR and modify it. I added you as a co-author.

pypsa/io.py
fneum: Ok, if it's difficult, I am fine with this!

pypsa/components.py
FabianHofmann: done

fneum:
Yes, agreed. Functionality is in good shape (no pun intended).

The final step would be to make the new component show up in the documentation:

https://pypsa--755.org.readthedocs.build/en/755/components.html

And the update of `n.srid` and `n.crs` network attributes:

https://pypsa.readthedocs.io/en/latest/components.html#network

drifter089:
Hey,
There's a minor issue: the `crs` isn't being updated after reading. It remains in the `meta`.
Screenshot 2023-10-25 170101

pypsa/io.py
```diff
@@ -870,8 +877,15 @@ def import_components_from_dataframe(network, dataframe, cls_name):
         else:
             if static_attrs.at[k, "type"] == "string":
                 dataframe[k] = dataframe[k].replace({np.nan: ""})
-
-            dataframe[k] = dataframe[k].astype(static_attrs.at[k, "typ"])
+            if dataframe[k].dtype != static_attrs.at[k, "typ"]:
+                if static_attrs.at[k, "type"] == "geometry":
+                    crs = network.meta.pop("_crs", None)
```
drifter089, line 882: this doesn't get executed when we import a new network with an empty shape component.

FabianHofmann: Exactly! thanks for raising. Perhaps we should instead go for n._crs object ...

drifter089: yes, that makes more sense as in the current scenario when we do this 
image
instead of assert {"test":"test"} == {"test" : "test"}
we are getting { "test" : "test , _crs : "....." } == { "test" : "test , _crs : "....." } 
when it is getting passed which might be an issue in some other cases

FabianHofmann: Okay, the more I am thinking about it, the more complicate it seems. I would say for now, we leave it imperfect and don't support tracking of CRS when shape is empty (it is really an edge case). I will remove the CRS setter for now. We should tackle the whole SRID/CRS topic in another PR.

FabianHofmann:
> Yes, agreed. Functionality is in good shape (no pun intended).
> 
> The final step would be to make the new component show up in the documentation:
> 
> https://pypsa--755.org.readthedocs.build/en/755/components.html
> 
> And the updated description of `n.srid` and `n.crs` network attributes:
> 
> https://pypsa.readthedocs.io/en/latest/components.html#network

Thanks for reviewing @fneum , I've tackled the points.

fneum (approved):

PR #759, opened 2023-10-18
enable and disable geometries

drifter089:
- **Idea**: We have our standard pandas dataframe with a column named `geometry` containing either `None` or string values.
- When utilizing the geometries, we convert these dataframes into geopandas dataframes, providing access to a rich set of functions for geometry-related operations.
- The `disable_geometry` function reverts the dataframe back to a pandas dataframe, where `shapely.Geometry` is translated into WKT strings.
- The `add` function has been modified to insert values as `shapely.Geometry` when geo is enabled, and as a "WKT string" when geo is disabled.
- By default, when writing these to any format, we disable geometries, resulting in a pandas dataframe containing string values.
- `plot` function (and others) can be modified to yield enhanced results when geometry is enabled.


FabianHofmann:
Hey @drifter089, really appreciate your motivation :) One thing that one also needs to be aware of, is that the testing is not complete with a flag and two modes of network data storage. Strictly spoken, all tests that we have, should run for the first mode without geometries and the second mode with geometries. Let's discuss it tomorrow but I see that the downsides and risks of this approach outweigh its convenience.

fneum:
Hey @drifter089, this PR also looks nice!

I have reviewed both your PR and #755, and my vote is in favour of #755.

I found this to be less invasive, as it only requires one new component to be a GeoDataFrame, and it allows to set multiple geometries for a single component (e.g. offshore and onshore shapes for a bus).

pz-max:
Yep, we agreed to close this in favour of #755

drifter089:
deeply appreciate your feedback @FabianHofmann  @fneum.
Thank you for pointing it out.

PR #760, opened 2023-10-19
Add capacity additions and retirements to statistics

coroa:
Adds a new `when` argument to only `optimal_capacity` add the moment, which allows switching from selecting power plants, which are "active" in a given investment period to only show them at their "build_year" or at their "retirement" year.

Not super happy with the names so fishing for suggestions and also comments on which other variables would benefit from such a scope.

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -14,6 +14,7 @@ __copyright__ = (
 
 import logging
 from functools import reduce, wraps
+from typing import Literal
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


pypsa/statistics.py
```diff
@@ -235,31 +262,41 @@ def aggregate_components(
...
+                # TODO uses 9999 as the sentinel value for after modelling period
+                investment_periods = pd.Index(np.r_[n.investment_periods, 9999])
+                if "build_year" not in n.df(c):
+                    period = investment_periods[0 if when == "build_year" else -1]
+                else:
+                    year = n.df(c)["build_year"]
+                    if when == "retirement":
+                        year += n.df(c)["lifetime"]
+
+                    period = investment_periods[
```
lisazeyen, line 286: I haven't tested it but is it necessary to define the `period` or can we just groupby `year`?

pypsa/statistics.py
```diff
@@ -235,31 +262,41 @@ def aggregate_components(
...
 
         df = func(n, c)
-        if isinstance(n.snapshots, pd.MultiIndex) and not isinstance(df, pd.DataFrame):
-            # for static values we have to iterate over periods and concat
-            per_period = {}
-            for p in n.investment_periods:
-                per_period[p] = df[n.get_active_assets(c, p).loc[df.index]]
-            df = pd.concat(per_period, axis=1)
+
+        if isinstance(n.snapshots, pd.MultiIndex):
```
lisazeyen, line 269: maybe this functionality is also useful for running myopic code where you don't have a multi-index snapshot but still the information about build year and lifetime

lisazeyen:
Great! I only have some minor comments. I can't come up with better names, sorry! I think they one suggested are fine!

FabianHofmann:
@coroa do you think this is still needed? I think the feature is nice, so if you don't mind I would create another pr starting from the up-to-date code basis

PR #761, opened 2023-10-19
Fix globalconstraints in a multiperiod context

coroa:
And some pandas deprecations.

### DIFF
--- a/pypsa/optimization/global_constraints.py
+++ b/pypsa/optimization/global_constraints.py
@@ -453,8 +453,8 @@ def define_transmission_expansion_cost_limit(n, sns):
     def substr(s):
         return re.sub("[\\[\\]\\(\\)]", "", s)
 
-    lhs = []
     for name, glc in glcs.iterrows():
+        lhs = []
         car = [substr(c.strip()) for c in glc.carrier_attribute.split(",")]
         period = glc.investment_period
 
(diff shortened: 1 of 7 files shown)


pypsa/optimization/global_constraints.py
```diff
@@ -389,12 +391,8 @@ def define_transmission_volume_expansion_limit(n, sns):
     m = n.model
     glcs = n.global_constraints.query("type == 'transmission_volume_expansion_limit'")
 
-    def substr(s):
-        return re.sub("[\\[\\]\\(\\)]", "", s)
-
-    lhs = []
     for name, glc in glcs.iterrows():
-        car = [substr(c.strip()) for c in glc.carrier_attribute.split(",")]
```
fneum, line 397: I think `car` is still needed. It's used a few lines down in a query string.

fneum (changes requested):
LGTM, almost.

One comment and a release note, then we can merge.

pypsa/optimization/global_constraints.py
coroa: Ah, yes, damn. Never trust VS Code blindly. The inquery string usage was lost on it.

coroa:
Thanks for the review. Comments should be addressed. Please feel free to squash!

coroa:
> Ah, yes, damn. Never trust VS Code blindly. The query string usage was lost on it.

PR #766, opened 2023-10-24
Clean shapes

drifter089:
I'm attempting to integrate `clean_geometries`. 
I've updated the `n.crs` to modify the projection and recreated the GeoDataFrame with these projections. when saving and retrieving.


FabianHofmann:
@drifter089 could you make this PR against #755 ?

drifter089:
Yes sure @FabianHofmann 
l will also get the tests to run. And some minor adjustments then do it.

PR #769, opened 2023-10-26
adding save_crs functions

drifter089:
Now, I'm storing `n.crs` directly instead of within `meta`.
I have hardcoded 4326 as CRS across the codebase, with the intention of deprecating `srid` usage in the future.


FabianHofmann:
Hey @drifter089, thanks for your PR. I like most of the concepts here. Some points for improving

- add a function `n.to_crs()` for changing the projection (similar to `geopandas.GeoDataFrame.to_crs`), this should also change the coordinates in `n.buses`, as well as srid
- let the `crs` setter overwrite the current CRS in the network (similar to `geopandas.GeoDataFrame.crs == ...`)
- convert `srid` to a property which uses `crs` in the background 


Here are some functions that should give a good start in `components.py`:


```python

    @property
    def crs(self):
        """
        Coordinate reference system of the network's geometries (n.shapes).
        """
        return self._crs

    @crs.setter
    def crs(self, new):
        """
        Set the coordinate reference system of the network's geometries (n.shapes).
        """
        self.shapes.crs = new
        self._crs = self.shapes.crs

    def to_crs(self, new):
        """
        Convert the network's geometries and bus coordinates to a new coordinate reference system.
        """
        current = self.crs
        self.shapes.to_crs(new, inplace=True)
        self._crs = self.shapes.crs
        self.buses["x"], self.buses["y"] = pyproj.transform(
            current, self.crs, self.buses["x"], self.buses["y"]
        )

    @property
    def srid(self):
        """
        Spatial reference system identifier of the network's geometries (n.shapes).
        """
        return self.crs.to_epsg()

    @srid.setter
    def srid(self, new):
        """
        Set the spatial reference system identifier of the network's geometries (n.shapes).
        """
        self.crs = pyproj.CRS.from_epsg(new)
```

It would require to remove the srid attribute definition in line 176 by 

```python
    _crs = pyproj.CRS.from_epsg(4326)
```

What do you think @drifter089 ?

FabianHofmann:
here is the update test function:

```python

def test_shape_reprojection(ac_dc_network_shapes):
    n = ac_dc_network_shapes

    with pytest.warns(UserWarning):
        area_before = n.shapes.geometry.area.sum()
    x, y = n.buses.x.values, n.buses.y.values

    n.to_crs("epsg:3035")

    assert n.shapes.crs == "epsg:3035"
    assert n.crs == "epsg:3035"
    assert area_before != n.shapes.geometry.area.sum()
    assert not np.allclose(x, n.buses.x.values)
    assert not np.allclose(y, n.buses.y.values)
```

drifter089:
@FabianHofmann ,
Thank you for the review this way's more intuitive and it also uses _crs and SRID. I like it! Gonna make changes along these lines.

drifter089:
@FabianHofmann I have made changes according to the suggestions please review it. 😄

FabianHofmann:
@drifter089 I would take the freedom to alter the functions in components.py a bit, are you fine with that?

FabianHofmann (approved):

FabianHofmann:
Thanks @drifter089! I reverted some things in the components.py back to the originally suggested functions. Looks all good to me

fneum (approved):

PR #780, opened 2023-11-09
Fix rolling horizon bug in unit commitment

### LINKED ISSUE #771: Rolling horizon unit commitment not using end states of previous periods
## Checklist

- I am using the current pypsa v0.25.1

## Describe the Bug

We are running unit commitment over one year periods with a rolling horizon of 365 hours. Running code that is similar to the rolling unit commitment example given in PyPSA documentation example for rolling unit commitment: https://pypsa.readthedocs.io/en/latest/examples/unit-commitment.html 

overlap = 2
for i in range(sets_of_snapshots):
    nu.optimize(nu.snapshots[i * len(p_set) : (i + 1) * len(p_set) + overlap])

We would expect for expect that for each period/value of i, values such as up_time_before for the generators would be calculated using the generator status values from the previous periods. But instead, it seems that for each period, optimize uses the up_time_before initial values that were set in the generators file. 

Note: This is indicated in the PyPSA documentation:

"If snapshots falls in the middle of network.snapshots, then PyPSA assumes the statuses for hours before snapshots have been set by previous simulations, and reads back the previous up-time by examining the previous statuses. "
https://pypsa.readthedocs.io/en/latest/optimal_power_flow.html

p-glaum:
Closes #771 

## Checklist


p-glaum:
Many thanks @sophie-pelland who tested this PR in #780

sophie-pelland:
Hi @p-glaum,

I did an initial test of your solution this morning, and at first I thought it was working, but I had forgotten to comment out my own fix (outside PyPSA). I just re-tested, commenting out my fix. The behavior is definitely more realistic, but there remains an issue: I think the up_time_before and down_time_before are not correctly calculated: I found a case where a unit with a min_down_time of 24 hours was down for only 3 hours...

p-glaum:
@sophie-pelland I tried to replicate your bug, but I could not. When I fixed min_down_time to e.g. 24, the generator stayed off for 24 snapshots, even though it would like to operate if I relax the min_down_time. Could you provide me a minimal example of your test?

sophie-pelland:
@p-glaum Unfortunately, I am not allowed to share the inputs I used :( I had also set a min_down_time of 24, and got a case where down time was only 3 hours.

p-glaum:
Since I could not reproduce the bug @sophie-pelland mentioned, I guess we can already merge this PR because it fixes some parts of the bug.

p-glaum:
As soon as we find a minimal example, which reproduces the bug, I will open another PR

FabianHofmann (approved):

FabianHofmann:
sounds good, thanks @p-glaum for looking into it!

PR #781, opened 2023-11-09
Add network function

### LINKED ISSUE #506: Idea? - implement __add__ for networks
Hi,

Occasionally we had the situation that there's subnetworks that could be defined separately, maybe we are also interested in solving those subnetwork separatly. Then we want to combine them into one big network.
Would it be an idea to implement the class method __add__ such that network1+network2 makes sense and gives you the combined network? Could make some workflows quite tidy. I guess there's a ton of things to consider there, overlapping buses for instance to have them connected after the merger. So at this stage just an idea.

Best,
Andreas

jessLryan:
Closes #506  

## Checklist

- [yes] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [yes] Unit tests for new features were added (if applicable).
- [N/A] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [yes] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [yes] I consent to the release of this PR's code under the MIT license.


FabianHofmann (commented):
before reviewing, I was thinking about a better naming. Perhaps `import_from_network` would be better aligning with the other import function names?

jessLryan:
> before reviewing, I was thinking about a better naming. Perhaps `import_from_network` would be better aligning with the other import function names?

how about "add_components_from_network"? Could also change the current add function in components.py to "add_component" so that the distinction is clear?

FabianHofmann:
> > before reviewing, I was thinking about a better naming. Perhaps `import_from_network` would be better aligning with the other import function names?
> 
> how about "add_components_from_network"? 

sounds good

> Could also change the current add function in components.py to "add_component" so that the distinction is clear?

which function do you mean exactly? I would not touch existing ones though...

jessLryan:
> > > before reviewing, I was thinking about a better naming. Perhaps `import_from_network` would be better aligning with the other import function names?
> > 
> > 
> > how about "add_components_from_network"?
> 
> sounds good
> 
okay great I'll do that :-)
> > Could also change the current add function in components.py to "add_component" so that the distinction is clear?
> 
> which function do you mean exactly? I would not touch existing ones though...

fair point

jessLryan:
@FabianHofmann can this now be merged?

FabianHofmann:
Hey @jessLryan, I will review as soon as I have time. This will likely take some time.

jessLryan:
@FabianHofmann I've come across a bug: we need to change the method so that buses are added first (otherwise new components e.g. lines, which are attached to the new buses will fail to be added due to a check at Line 914 of io.py). I'll update and commit

jessLryan:
> @FabianHofmann I've come across a bug: we need to change the method so that buses are added first (otherwise new components e.g. lines, which are attached to the new buses will fail to be added due to a check at Line 914 of io.py). I'll update and commit

now fixed

PR #793, opened 2023-12-02
Fix stand_by_cost implementation in optimization

### LINKED ISSUE #792: Stand_by_cost is not implemented correctly
## Checklist

- [x] I am using the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch or the latest [release](https://github.com/pypsa/pypsa/releases). Please indicate:

## Describe the Bug

**The standby cost for generators and links is not implemented correctly and subsequently its not present in the the objective function.**

The issue is caused by the way that the feature is implemented. While all the other parts of the objective function are appended together in a list and then merged together, the standby cost is added to the objective function directly.  **The problem is that this is done before the other parts of the objective function are merged and therefore its gets overwritten**

## Possible solutions

As far as I see, this issue can be solved in two ways:
-either the method of adding the standby cost to the objective function is harmonized with the rest of the code in
https://github.com/PyPSA/PyPSA/blob/f94aeeef03db0f6e52a9a273096831d97f1b676e/pypsa/optimization/optimize.py#L161

-or if there is a specific reason for this feature to be written in this manner, it can just be moved down after the construction of the objective function in 
https://github.com/PyPSA/PyPSA/blob/f94aeeef03db0f6e52a9a273096831d97f1b676e/pypsa/optimization/optimize.py#L199

cfvescovo:
Closes #792.

Uses the same approach applied when defining the other parts of the objective function (e.g. marginal cost) for stand_by_cost

## Checklist

### DIFF
--- a/test/test_stand_by_cost.py
+++ b/test/test_stand_by_cost.py
@@ -2,13 +2,15 @@
 import numpy as np
 import pandas as pd
 import pytest
-from conftest import SUPPORTED_APIS, optimize
+from conftest import optimize
 from numpy.testing import assert_array_almost_equal as equal
 
 import pypsa
 
+SUPPORTED_APIS_STAND_BY_COST = ["linopy", "native"]
 
-@pytest.mark.parametrize("api", SUPPORTED_APIS)
+
+@pytest.mark.parametrize("api", SUPPORTED_APIS_STAND_BY_COST)
 def test_stand_by_cost(api):
     """
     This test is based on https://pypsa.readthedocs.io/en/latest/examples/unit-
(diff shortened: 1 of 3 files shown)


test/test_stand_by_cost.py
```diff
@@ -2,13 +2,15 @@
...
 from numpy.testing import assert_array_almost_equal as equal
 
 import pypsa
 
+SUPPORTED_APIS_STAND_BY_COST = ["linopy", "native"]
 
-@pytest.mark.parametrize("api", SUPPORTED_APIS)
+
+@pytest.mark.parametrize("api", SUPPORTED_APIS_STAND_BY_COST)
 def test_stand_by_cost(api):
```
davide-f, line 14: Hello and thanks for taking care of the PR :)
The functionality seems ok to me.

As little comment, I'd revise the exclusion of pyomo as follows.
Rather then creating a new support app variable, in this test function, I'd add an if close like:

If api=="pyomo":
    Return

The effect should be the same but I think it is less impactful.

Thanks for taking care of it :D

fneum (approved):

PR #798, opened 2023-12-07
Sclopt patch, allows multiple subnetworks and pre-existing constraints

TomAdamsHET:
Closes # (if applicable).

Allows naming of constraints for separate subnetworks, and allows existing constraints to carry through

## Checklist

### DIFF
--- a/pypsa/optimization/abstract.py
+++ b/pypsa/optimization/abstract.py
@@ -213,7 +213,7 @@ def optimize_security_constrained(
         )
 
     # Allow existing constraints to carry through
-    if 'model' in dir(n):
+    if "model" in dir(n):
         m = n.model
     else:
         m = n.optimize.create_model(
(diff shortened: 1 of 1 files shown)


pypsa/optimization/abstract.py
```diff
@@ -212,11 +212,15 @@ def optimize_security_constrained(
...
-    )
+    # Allow existing constraints to carry through
+    if "model" in dir(n):
+        m = n.model
+    else:
+        m = n.optimize.create_model(
+            snapshots=snapshots,
+            multi_investment_periods=multi_investment_periods,
+            **model_kwargs,
+        )
```
fneum, line 223: It's a nice idea, but I would not handle it in this way by default. The risk is that components added between the first model creation and this function call may be ignored:

```py
n.optimize()
n.add("Generator", "gen", ....)
n.optimize.optimize_security_constrained()
```

I would also replace:

```py
if "model" in dir(n):
```

with

```py
get(n, "model")
```

But I would consider adding this as an optional keyword argument of the function.

fneum:
possibly related to #946

PR #804, opened 2023-12-12
remove `n_mod` output attribute

fneum:
because:

- it only contains duplicate information (can calculate via `p_nom_opt / p_nom_mod`)
- the `int` output caused problems


FabianHofmann (approved):
Yes, I like this leaner implementation.

PR #818, opened 2024-01-23
bugfix: sign error in line capacity constraint with losses

fneum:
This is a relatively minor bug as the sign error only affects the line capacity limits in one direction and only to the level of the transmission losses. Only extendable lines are affected.

The pyomo version is not affected by this. Bug must have been introduced in translation to nomopyomo and then carried over into the linopy backend.

The constraint should mirror: https://arxiv.org/pdf/2008.11510.pdf equation 32

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -7,6 +7,9 @@ Upcoming Release
 
 .. warning:: The features listed below are not released yet, but will be part of the next release! To use the features already you have to install the ``master`` branch, e.g. ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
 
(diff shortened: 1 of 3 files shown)


Irieo (commented):
This fixes the bug and aligns code with the equation as intended.

the stylistic concern is that expression structure of the two constraints `f"{c}-fix-{attr}-lower"` (and -upper) is different to `f"{c}-ext-{attr}-lower"` (and -upper). The latter brings all terms to LHS, the former keeps line capacity at the RHS. NB this is the same constraint, written for non-extendable and extendable subsets of model components. Although minor, this can lead to plus/minus-type of bugs and hinder pattern recognition for quick debugging. 
if the stylistic concern is less of a problem, let's merge.

PR #841, opened 2024-02-19
merge instead of concat

lisazeyen:
Closes # (if applicable).

This PR proposes a formulation which consumes less memory when calling the function `get_switchable_as_dense` especially with large dataframes, concating the static and varying dataframes consumes a lot of memory.
## Checklist


FabianHofmann:
@lisazeyen you think the `del` calls are necessary? garbage collection of python should be taking care of it.

FabianHofmann:
@lisazeyen I looked up the functionality of `pd.merge`, and I have a semi-good gut feeling about using it. Could you perhaps try wether we could use something like this?

```python

    static = df.loc[fixed_i, attr]
    varying = pnl[attr].loc[snapshots, varying_i]
    return varying.assign(**static).reindex(columns=index)
```

and if that is perhaps memory efficient?

lisazeyen:
> @lisazeyen I looked up the functionality of `pd.merge`, and I have a semi-good gut feeling about using it. Could you perhaps try wether we could use something like this?
> 
> ```python
>     static = df.loc[fixed_i, attr]
>     varying = pnl[attr].loc[snapshots, varying_i]
>     return varying.assign(**static).reindex(columns=index)
> ```
> 
> and if that is perhaps memory efficient?


 this would work as well, what are your suspicions about `pd.merge`?

FabianHofmann:
my suspicion is that I don't understand it fully  :D 

my question would be why left_index and right_index? 
and why join=inner? this should only lead to returning the overlapping columns or? 

`pd.merge` is rather used for more complex operations where you have partially overlapping and distinct indexes and columns. But here the operation is quite clear: just add a couple of new columns with constant values. since this is a very basic function I would like to make it as clear as possible.

PR #855, opened 2024-03-26
weighted active extendable transmission costs for multihorizon constraint

AdrienneRTE:
The constraint on the total transmission cost for a multi-horizon optimization didn't included the weights of relative investment periods. 
Each active period of a line/link is weighted by their respective investment_period_weightings.

The weighting applies only if the constraint's period is nan and the snapshots are multi-indexed. Otherwise, the weight is set to 1. 

## Checklist
I tested the constraint with and without a specific period and it worked fine.

## Open question
- Should a constraint set on a specific investment_period be weighted ? For the sake of simplicity, I assumed that it shouldn't be.


pypsa/optimization/global_constraints.py
```diff
@@ -477,13 +481,27 @@ def substr(s):
             )
 
             if not isnan(period):
-                ext_i = ext_i[n.get_active_assets(c, period)[ext_i]].rename(ext_i.name)
+                ext_i = ext_i[n.get_active_assets(c, period)[ext_i]].rename(
+                    ext_i.name
+                )  # or extendable only ?
```
FabianHofmann, line 486: what do you mean here?

FabianHofmann (commented):
@AdrienneRTE this looks good. As you said, I would leave the single period selection unweighted.

pypsa/optimization/global_constraints.py
AdrienneRTE: That's some scrap of a question I asked myself but solved since, and forgot to remove ^^'

fneum:
fixed and merged via 26720f0

PR #860, opened 2024-04-08
Port consideration in statistics module

FabianHofmann:
* The statistics module introduces a new keyword argument `at_port` to all functions. This allows considering the port of a component when calculating statistics. Depending on the function, the default of `at_port` is set to `True` or `False`, for example for the dispatch all ports are considered.

* The statistics module now supports an optional `port` argument in `groupby` functions. This allows to group statistics while considering the port of a component.

* The `statistics.revenue` function introduces a new keyword argument `kind` to optionally calculate the revenue based on the `input` commodity or the `output` commodity of a component.

## Checklist

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -14,6 +14,7 @@ __copyright__ = (
 
 import logging
 from functools import reduce, wraps
+from inspect import signature
 
 import numpy as np
 import pandas as pd
@@ -41,11 +42,20 @@ def get_bus_and_carrier(n, c, port="", nice_names=True):
     """
     Get the buses and nice carrier names for a component.
     """
-    if port == "":
-        bus = "bus0" if "bus" not in n.df(c) else "bus"
-    else:
-        bus = f"bus{port}"
+    bus = f"bus{port}"
     return [n.df(c)[bus].rename("bus"), get_carrier(n, c, nice_names=nice_names)]
+
+
+def get_bus_unit_and_carrier(n, c, port="", nice_names=True):
+    """
+    Get the buses and nice carrier names for a component.
+    """
+    bus = f"bus{port}"
+    return [
+        n.df(c)[bus].rename("bus"),
+        n.df(c)[bus].map(n.buses.unit).rename("unit"),
+        get_carrier(n, c, nice_names=nice_names),
+    ]
 
 
 def get_name_bus_and_carrier(n, c, port="", nice_names=True):
@@ -119,6 +129,11 @@ def get_ports(n, c):
     return [col[3:] for col in n.df(c) if col.startswith("bus")]
 
 
+def get_stacked_ports(n, c):
+    ports = get_ports(n, c)
+    return pd.concat([n.df(c)[f"bus{p}"][lambda ds: ds != ""] for p in ports])
+
+
 def port_mask(n, c, port="", bus_carrier=None):
     """
     Get a mask of components which are optionally connected to a bus with a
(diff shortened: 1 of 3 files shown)


pypsa/statistics.py
```diff
@@ -119,6 +129,11 @@ def get_ports(n, c):
     return [col[3:] for col in n.df(c) if col.startswith("bus")]
 
 
+def get_stacked_ports(n, c):
```
p-glaum, line 132: Why do we need this function, because currently it is not used.

pypsa/statistics.py
```diff
@@ -190,6 +204,48 @@ def get_transmission_carriers(n, bus_carrier=None):
...
+        if "port" in signature(groupby).parameters:
+            by = groupby(n, c, port=port, nice_names=nice_names)
+        else:
+            by = groupby(n, c, nice_names=nice_names)
+    elif isinstance(groupby, list):
+        by = [n.df(c)[key] for key in groupby]
+    elif isinstance(groupby, str):
+        by = n.df(c)[groupby]
+    elif isinstance(groupby, dict):
+        # TODO: fix kwargs behavior. what is this used for again?
```
p-glaum, line 239: I think this was formerly used for the energy balance, where we wanted group by the index levels so we needed something like {"level":["bus", "carrier", "bus_carrier"]}

p-glaum: I think we can now remove it

pypsa/statistics.py
```diff
@@ -210,19 +266,52 @@ def aggregate_timeseries(df, weights, agg="sum"):
...
+    """
+    Filter the DataFrame for components which are connected to a bus with
+    carrier `bus_carrier`.
+    """
+    if bus_carrier is None:
+        return df
+
+    ports = n.df(c).loc[df.index, f"bus{port}"]
+    port_carriers = ports.map(n.buses.carrier)
+    if isinstance(bus_carrier, str):
```
p-glaum, line 291: I think here it would be nice to have somehing like I did before, that we fist check if the `string` is in n.buses.carrier.unique() and if not we look if the bus_carriers contain the `string`

pypsa/statistics.py
```diff
@@ -210,19 +266,52 @@ def aggregate_timeseries(df, weights, agg="sum"):
...
+        return df[port_carriers == bus_carrier]
```
p-glaum, line 292: something like:

```suggestion
    if isinstance(bus_carrier, str):
        if bus_carrier in n.buses.carrier.unique():
            return df[port_carriers == bus_carrier]
        else:
            return df[bus_carrier in port_carriers]
```

pypsa/statistics.py
```diff
@@ -41,13 +42,22 @@ def get_bus_and_carrier(n, c, port="", nice_names=True):
...
     """
-    if port == "":
-        bus = "bus0" if "bus" not in n.df(c) else "bus"
-    else:
-        bus = f"bus{port}"
+    bus = f"bus{port}"
     return [n.df(c)[bus].rename("bus"), get_carrier(n, c, nice_names=nice_names)]
 
 
+def get_bus_unit_and_carrier(n, c, port="", nice_names=True):
```
p-glaum, line 49: should add it to Groupers

pypsa/statistics.py
```diff
@@ -898,48 +980,32 @@ def energy_balance(
             Note that for {'mean', 'sum'} the time series are aggregated to MWh
```
p-glaum, line 980: Maybe make the energy balance compatible with attributes of other statistic functions and make the attribute aggregate_bus setting the grouper to the corresponding one.

pypsa/statistics.py
```diff
@@ -898,48 +980,32 @@ def energy_balance(
...
-            return aggregate_timeseries(p, weights, agg=aggregate_time)
-
-        groupby = ["carrier", "bus_carrier"]
-        if not aggregate_bus:
-            groupby.append("bus")
-
-        df = aggregate_components(
-            n,
-            func,
+        if aggregate_bus:
```
p-glaum, line 983: maybe also add the other attributes from dispatch like `groupby`to the energy_balance and add warning that with `aggregate_bus=True`, this attribute will be ignored

PR #851, opened 2024-03-12
Improve statistics

p-glaum:
Closes # (if applicable).

## Changes proposed in this Pull Request

Does the aggregation to the wanted grouper format while having the most details. Therefore, one can retrieve the supply for the correct location and can group it for the countries. This is still compatible with the former grouper functions and should be completely backwards compatible.
Furthermore, now one can similarly to `supply` and `withdrawal` split the revenue into `input_cost` and `output_revenue`

## Checklist

- [ ] Improve documentation and examples
- [ ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ ] Unit tests for new features were added (if applicable).
- [x] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #868, opened 2024-04-16
Improve statistics 2.0

p-glaum:
This combines the features from #860 and #851 to have a final version with all the new features while providing backwards compatibility. The only change which is not backward compatible is the default grouper from all `energy_balance` related features like `supply`, `withdrawal`, `dispatch`. But we think it is better to have the new grouper which retains the information on the bus carrier instead of summing different bus carriers (e.g. electricity and heat energy). 

I validated the PR against the current statistics module, and it provides the same values for the combination of statistics methods and groupers.

Closes PR #860 

## Checklist

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -129,11 +129,6 @@ def get_ports(n, c):
     return [col[3:] for col in n.df(c) if col.startswith("bus")]
 
 
-def get_stacked_ports(n, c):
-    ports = get_ports(n, c)
-    return pd.concat([n.df(c)[f"bus{p}"][lambda ds: ds != ""] for p in ports])
-
-
 def port_mask(n, c, port="", bus_carrier=None):
     """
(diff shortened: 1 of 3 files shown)


pypsa/statistics.py
```diff
@@ -119,6 +129,11 @@ def get_ports(n, c):
     return [col[3:] for col in n.df(c) if col.startswith("bus")]
 
 
+def get_stacked_ports(n, c):
+    ports = get_ports(n, c)
+    return pd.concat([n.df(c)[f"bus{p}"][lambda ds: ds != ""] for p in ports])
+
+
```
FabianHofmann, line 136: I think we can remove this, right? Was an intermediate alternative route from me.

p-glaum: will do so.

FabianHofmann (approved):
great work @p-glaum

PR #870, opened 2024-04-17
Save shadow price for CO2Limit global constraint as positive

martavp:
Closes # (if applicable).

The dual of the CO2Limit global constraint was saved as negative due to[ the co2_emissions defined as -1](https://github.com/PyPSA/pypsa-eur/blob/a0436372febd158155fa469ce74a4cb69199c300/scripts/prepare_sector_network.py#L601). This PR forces the shadow CO2 price to be saved as a positive number. 

## Checklist

### DIFF
--- a/pypsa/optimization/optimize.py
+++ b/pypsa/optimization/optimize.py
@@ -417,7 +417,10 @@ def assign_duals(n, assign_all_duals=False):
         elif (c == "GlobalConstraint") and (
             (assign_all_duals or attr in n.df(c).index)
         ):
-            n.df(c).loc[attr, "mu"] = dual
+            if attr == 'CO2Limit':
+                n.df(c).loc[attr, "mu"] = dual*n.carriers.co2_emissions["co2"]
+            else:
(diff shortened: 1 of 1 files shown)


FabianHofmann:
@martavp I don't know if that is the right spot. people might name their constraints differently, "co2_limit", "co2limit", "co2-constraints' .... 
Like this, we are only covering one specific case. Would it not be better to fix that in pypsa-eur explicitely? Then we also leave the functionality and its logic generic without special cases.

martavp:
Thanks for the suggestion @FabianHofmann 

I know the solution is a bit hacky but at least self-explanatory. 

Would you suggest modifying the network in PyPSA-Eur after the optimize function was run?

Wouldn't potential people using different names for their co2Limite-related constraint also expect to get a positive value for the dual?

fneum:
I will close this suggestion as I think the shadow prices / duals should have consistent orientation. It's more a question of convention. In PyPSA and most solvers: Negative dual = reducing cost when constraint is relaxed in minimisation problem.

PR #879, opened 2024-04-30
optimize: tackle future warning in df.update

FabianHofmann:
& test_consistency_check: avoid future warning

... these are quite tiny fixes

### DIFF
--- a/pypsa/optimization/optimize.py
+++ b/pypsa/optimization/optimize.py
@@ -354,7 +354,8 @@ def assign_solution(n):
             else:
                 set_from_frame(n, c, attr, df)
         elif attr != "n_mod":
-            n.df(c)[attr + "_opt"].update(df)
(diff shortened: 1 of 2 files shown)


pypsa/optimization/optimize.py
```diff
@@ -354,7 +354,8 @@ def assign_solution(n):
             else:
                 set_from_frame(n, c, attr, df)
         elif attr != "n_mod":
-            n.df(c)[attr + "_opt"].update(df)
+            idx = df.index.intersection(n.df(c).index)
```
p-glaum, line 357: Everything looks fine, I just don't quite understand this line, but it has more to do with the fact that I haven't used the modularity feature yet.

p-glaum (approved):

pypsa/optimization/optimize.py
fneum: This is fine.

Issue #877, opened 2024-04-26
`define_operational_limit` modifies store carrier


koen-vg:
## Checklist

- [x] I am using the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch or the latest [release](https://github.com/pypsa/pypsa/releases). Please indicate:

## Describe the Bug

````
>>> n = pypsa.Network()
>>> n.add("Bus", name="foo", carrier="A")
>>> n.add("Store", name="bar", bus="foo", carrier="B")
>>> n.add("Generator", name="baz", bus="foo", marginal_cost=1)
>>> n.add("GlobalConstraint", name="lim", type="operational_limit")
>>> n.stores.carrier
Store
bar    B
Name: carrier, dtype: object
>>> pypsa.optimization.optimize.create_model(n)
Linopy LP model
[...]
>>> n.stores.carrier
Store
bar    A
Name: carrier, dtype: object
````

One can argue whether stores should always have the carrier of their bus, but if that's the case, then pypsa shouldn't allow stores having a different carrier from the bus they are attached to in the first place. As it stands, I think this can reasonably be classified as a bug.

I found this while working with PyPSA-Eur, where it turns out that stores with the "H2 Store" carrier are modified to take the "H2" carrier, among others.

The offending code is found at https://github.com/PyPSA/PyPSA/blob/master/pypsa/optimization/global_constraints.py#L367:
````python
        # stores
        n.stores["carrier"] = n.stores.bus.map(n.buses.carrier)
        stores = n.stores.query("carrier == @glc.carrier_attribute and not e_cyclic")
````
I would propose modifying this to something like
````python
        # stores
        bus_carrier = n.stores.bus.map(n.buses.carrier)
        stores = n.stores.query("@bus_carrier == @glc.carrier_attribute and not e_cyclic")
````

I'd be happy to submit a PR for this; just opening the issue to open up for discussion about potential other approaches, and so others might find this.

koen-vg:
(Note that there are identical instances of the same bus, if considered a bug, in pyps-eur under the definition of co2 constraints; I'd also submit a PR for those)

koen-vg:
[...]

PR #880, opened 2024-05-02
Bugfix: don't  modify store carrier attribute when defining operational limits

koen-vg:
Closes https://github.com/PyPSA/PyPSA/issues/877

Fairly self-explanatory (see also the issue).

Before merging this PR, I think it's worth discussing if this is even the behaviour that's desired? As it stands, stores are treated differently from storage units and generators in this constraint, and I'm not sure if that's very helpful -- it seems to me like this mostly has the potential to cause confusion.

Maybe it would be useful to bring up a concrete example where a store plays a role in a global operational constraint in a meaningful way, so it's possible to evaluate the behaviour of this constraint. Subsequently, such an example could also be added to the documentation under https://pypsa.readthedocs.io/en/latest/optimal_power_flow.html#operational-limit (which should be improved anyway).

Off the top of my head, I can actually only imagine scenarios where you want to use the carrier of the store, and _not_ the bus it's attached to, i.e. the "opposite" of the current behaviour. For instance, say you have stores at different nodes representing gas reserves, and you want to introduce a global constraint on the use of gas reserves. You might have buses with carrier `gas`, and stores with carrier `gas reserve`; you would want to add a global constraint with `carrier_attribute == 'gas reserve'`, but with the code as it stands, this doesn't work since the carrier of the stores is mapped to `gas` before being compared to `carrier_attribute`.

So an alternative would be to replace
````python
n.stores["carrier"] = n.stores.bus.map(n.buses.carrier)
stores = n.stores.query("carrier == @glc.carrier_attribute and not e_cyclic")
````
by simply
````python
stores = n.stores.query(cond)
````
(Using the same condition as for storage units)

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -25,6 +25,8 @@ Release Notes
 
 * Updated environment_doc.yml to include the latest required pip dependencies for the documentation environment.
 
+* Bugfix: calling `create_model` or `optimize` when a global operational limit is defined will no longer set the carrier attribute of stores to the carrier of the bus they are attached to.
+
 PyPSA 0.27.1 (22nd March 2024)
 =================================
 
--- a/pypsa/optimization/global_constraints.py
+++ b/pypsa/optimization/global_constraints.py
@@ -364,8 +364,10 @@ def define_operational_limit(n, sns):
             rhs -= sus.state_of_charge_initial.sum()
 
         # stores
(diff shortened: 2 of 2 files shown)


fneum (approved):
I think the way you fixed this is suitable. Carrier modification in the optimisation is a no go.

PR #885, opened 2024-05-13
Update optimal capacity statistics

p-glaum:
With this PR, capacities for a bus_carrier are now provided with negative and positive values, depending on whether we have a production capacity or a consumption capacity. For instance, when we have H2 Electrolysis, it produces H2 and consumes electricity. Hence, the H2 capacity is positive, and the electricity capacity is negative.
For the default, when we do not select a bus_carrier, the absolute value of the component attribute is returned. 

## Checklist


pypsa/statistics.py
```diff
@@ -559,10 +560,14 @@ def optimal_capacity(
 
         if storage:
             comps = ("Store", "StorageUnit")
+        if bus_carrier:
+            at_port = True
```
FabianHofmann, line 564: This is a hard override of the keyword argument, which may lead to unexpected behaviour. We can instead put the default of `at_port` to None and only override if None. 
```python
    if bus_carrier and at_port is None:
            at_port = True
```

FabianHofmann: Or perhaps even better: not overloading the `at_port` argument: Instead create a new keyword argument `absolute` / `absolute_capacities` which is None and aligns to `bus_carrier` per default.
```python
if bus_carrier and absolute_capacities is None:
    absolute_capacities = True
```

p-glaum: The reason I used `at_port=True` is that it is more similar to how we had it before introducing the `at_port` parameter. At that time, we always looped over all ports to check if any bus had the provided `bus_carrier`. With the current implementation in the master, when selecting `bus_carrier="electricity"`, for example, you do not get links that have the `bus_carrier` "electricity" at port 2, e.g. CHP in PyPSA-Eur. Hence, it looks like they do not exist in the network.

p-glaum: Your first suggestion is valid and I will adapt it.

lkstrp:
If you don't mind, you could further improve the docstrings by adding argument descriptions. See:
```ruff check pypsa/statistics.py --select D417```

This would be nice for both PRs/ all of `statistics.py`

**Review by fneum — CHANGES_REQUESTED** [review]

File: pypsa/statistics.py [hunk 2]
```diff
@@ -538,16 +538,17 @@ def optimal_capacity(
         comps=None,
         aggregate_groups="sum",
         groupby=None,
-        at_port=False,
+        at_port=None,
```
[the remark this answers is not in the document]
p-glaum: I changed this, because None is also leading to the same result as False. However, if someone deliberately wants to set at_port to False, even if it at first sight makes no sense when bus_carrier is set, then at_port=False is still kept and not replaces by True.

pypsa/statistics.py
```diff
@@ -247,7 +247,7 @@ def filter_bus_carrier(n, c, port, bus_carrier, df):
         if bus_carrier in n.buses.carrier.unique():
             return df[port_carriers == bus_carrier]
         else:
-            return df[bus_carrier in port_carriers]
+            return df[port_carriers.str.contains(bus_carrier).fillna(False)]
```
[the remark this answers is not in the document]
p-glaum: I did it that one can e.g. also enter "heat" as bus_carrier and then the network checks if this is a unique bus_carrier and if not, it looks for all bus_carriers which contains the given bus_carrier like "urban central heat", "rural heat" or "urban decentral heat".

fneum (approved):

Issue #899, opened 2024-05-22
Use ruff


lkstrp:
[Ruff](https://docs.astral.sh/ruff/) is a very fast linter and formatter. It combines some of the tools we use right now and offers many more. We could simplify the existing linting and give all developers more direct feedback/ enforce more lint rules when using it, via `pre-commit` + uniform and better IDE/ CLI feedback.

Advantages:
- Better CLI/ IDE integration than `pre-commit`. Like `black` but with linting.
- Allows for dev feedback on things which the linter can't solve (like bare excepts, docstrings etc.)
- Many more rules could be added
- Fast
- Fewer dependencies to track

We could add it in multiple steps:
- [x] (1) Apply Basic Ruff settings (done in #900)
- [x] (2) Move existing linters to ruff (we can just keep the the stuff Ruff can't do) (done in #901)
- [x] (3) Let pre commit fail when manual fixes are needed (done in #901)
- [ ] (4) Add more rules
- [ ] Add to developer contribution documentation

Nothing of this is urgent. I just added a PR for the first step, so people can play around and are not buried in warnings.

### Tools:
In use right now (`pre-commit`).

Formatter:
- black
    - Ruff equivalent: yes

Linter:
- isort
    - Ruff equivalent: yes
- absolufy-imports
    - Ruff equivalent: yes
- codespell
    - Ruff equivalent:  I don't think so
- docformatter
    - Ruff equivalent: similar rules exist
- jupyter remove output
    - Ruff equivalent:  I don't think so
- pretty yaml
    - Ruff equivalent:  I don't think so
- end of file fixer/ trailing whitespaces, mixed line ending etc.
    - Ruff equivalent:  I think so

Static type checker
- mypy (not in use yet but planned)
    - Ruff equivalent:  Not implemented yet but in dev

And tons of more stuff can be set up: [Rules - Ruff](https://docs.astral.sh/ruff/rules/)

PR #900, opened 2024-05-22
Apply basic ruff settings

lkstrp:
See #899 

Only basic Ruff style settings are applied in this PR, to test it out and to use it next to the existing structure.

## Checklist

### DIFF
--- a/examples/multi-decade-example.py
+++ b/examples/multi-decade-example.py
@@ -7,6 +7,7 @@ example script to illustrate new features of the multi-decade investment
 
 @author: Lisa
 """
+
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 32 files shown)


fneum (approved):

FabianHofmann:
@lkstrp we could also add it to pre-commit

lkstrp:
@FabianHofmann 
Agree, that would be step 2 in here #899

We just shouldn't run ruff and the equivalent at the same time, because they sometimes differ slightly and then format against each other. I'll check how much we can move at some point

FabianHofmann:
Ah, I have not seen this. Sounds great

PR #901, opened 2024-05-22
Add more rules to ruff and apply them

lkstrp:
See #899 (Step 2 and 3)

## Checklist


.pre-commit-config.yaml
```diff
@@ -3,51 +3,30 @@ repos:
...
-
-# Formatting with "black" coding style
-- repo: https://github.com/psf/black
-  rev: 24.4.2
-  hooks:
-  # Format Python files
-  - id: black
-  # Format Jupyter Python notebooks
-  - id: black-jupyter
-
```
FabianHofmann, line 50: that means ruff replaces docformatter, balck and isort?

FabianHofmann (approved):
Looks good! Feel free to merge :)

.pre-commit-config.yaml
lkstrp: Yes. For the Docstrings it covers mostly the same on what docformatter did and more (see [here](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w)). I adjusted the rules in `pyproject.toml` and we can add some more in future (e.g. once we have docstrings everywhere, so it fails if you wanna merge something without docstring)

FabianHofmann: great

PR #902, opened 2024-05-22
Configure individual line colors in add_legend_line

yerbol-akhmetov:
Good day. Here I propose a small but useful change to `add_legend_lines` function in `plot.py` script. In particular, the functionality to choose individual line colors in legend is being added. Currently, `patch_kw` can be used to configure line parameters as a whole, not individual lines. So it is not possible to give individual line colors, which is useful when plotting lines and links with different colors (instead of defining another instance) or when you want to give different color for different lines. The modification helps to add color as a list similar to sizes. However, giving the color is optional. So it is compatible with old code. 
image

## Checklist


martacki (approved):
Looks good! I also tested, and the suggestion works as intended! Thanks @yerbol-akhmetov

PR #907, opened 2024-05-28
Auto release & Improve CI

lkstrp:
Some changes to the CI and release workflow:
- Moved `setup.py` content to `pyproject.toml`
- No manual version bumping needed anymore (just use a tag)
	- Use [setuptools scm](https://setuptools-scm.readthedocs.io/en/latest/) for version extraction
	- `CITATION.cff` is manually updated via deploy action, because `setuptools_scm` can't handle that
- Deploy action
	- Only runs for tagged version now (instead of starting and aborting afterwards) -> cleaner job history
	- Also creates GitHub releases
		- I added a link to the revised release notes in the documentation, we could probably drop the auto generated release notes here completely.
	- Also updates `CITATION.cff`, see above
- Renamed ci action (test)
	- Since only tests are run here and all workflows are ci

Using `setuptools_scm` has some additional changes:
- All commits after the release `0.28.0` do not have the version `0.28.0` anymore, but get a dev version for each commit after the release like `pypsa-0.28.1.dev18+ged71d587`
	- Which means there are some changes for the internal behaviour of the version for network objects and loading examples from GitHub. In linopy this was probably not noticed because there is no methodology based on the version.
	- I adjusted it so far, but this could be improved in the future


pypsa/examples.py
```diff
@@ -23,13 +24,36 @@
...
+
+
 def _repo_url(master=False):
     url = "https://github.com/PyPSA/PyPSA/raw/"
     if master:
         return f"{url}master/"
-    from pypsa import __version__  # avoid cyclic imports
-
-    return f"{url}v{__version__}/"
+    from pypsa import release_version  # avoid cyclic imports
```
FabianHofmann, line 44: Makes sense to have release version defined. The decrementing feels a bit hacky but should be okay :)

FabianHofmann (approved):
Looks great! Feel free to merge :)

pypsa/examples.py
lkstrp: Yeah I would change the decrement at some point. Either just going for the examples in the master branch, or going for examples for each version. But then they need to be updated in every version which is not happening right now.

PR #912, opened 2024-06-04
`plots.py` overhaul

### LINKED ISSUE #302: Plotting: split `flow` into `line_flow`, `link_flow`, `transformer_flow` to avoid multiindex
possibly fix line width, line flow only varies arrow size

### LINKED ISSUE #889: warning suppression for notebooks
I would suggest to remove some of the warnings in the notebooks, for readability. I use this function to suppress some of the info and warnings:
import warnings 
import logging

def suppress_warnings_and_info():
    warnings.filterwarnings("ignore", category=UserWarning, module="cartopy") # suppress irrelevant cartopy plotting warnings resulting from PyPSA itself
    # Configure logging to suppress INFO messages
    logging.getLogger('pypsa.pf').setLevel(logging.WARNING)
    logging.getLogger('pypsa.io').setLevel(logging.WARNING)
    logging.getLogger('gurobipy').setLevel(logging.WARNING)
    logging.getLogger('pypsa.optimization.optimize').setLevel(logging.WARNING)
    return 

the cartopy suppression may already have been implemented.

lkstrp:
Here is a general overhaul of the `plots.py` module:
- `plots.py`
	- Modularise and generalise `n.plot()` with `NetworkPlotter` class
	- Add type hints and update docstrings
	- Return plot collections as dict instead of dynamic list
		- Could be changed completely. I don't know all use cases for this, but if it is just for the legend, it would be easier if this is an argument
	- Deprecated:
		- `flow` to `line_flow`, `link_flow`, `transformer_flow`. See issue below
		- `bus_norm`, `line_norm`, `link_norm`, `transformer_norm` to `x_cmap_norm`
		- `color_geomap` to `geomap_colors`
- Issues
	- Closes #302
	- Fixes #889
- Other
	- `requests` library is not needed anymore
	- Changes in `geo.py`
		- Move geo functions from `plot.py` here
		- Add type checking and docstrings
	- Better deprecation
		- Add `utils.py` with changed kwargs deprecation decorator
		- Use `DeprecationWarning` instead of logger warnings

Todos:
- [ ] Rewrite `n.iplot()` with new structure
- [ ] Add animations
- [ ] Add option to plot cluster shapes
- [ ] Improve general plotting documentation
- [ ] Docstrings and types
- [ ] Add tests
- [ ] Could continue cleaning and simplify/ generalise things

Discuss:
- More API changes
	- I would also change `geomap` to `disable_geomap` and add a `geomap_res` argument, instead of mixing types
	- Maybe even drop `geomap` and include it into `projection`
- branch components
	- For all branch components we have many duplicated arguments, this could be simplified
	- `branch_components` could be dropped and controlled via attribute arguments
- This is mainly a clean and split up of all the things happened before in `n.plot()`
	- More simplification with generalized branch component Collections could be added in the background without API changes
	- Collection return
		- The collection return is not very intuitive
		- Right now I changed it to a dict, but we could also return a custom class here, that would also allow a lot of other functionality. But this would be another PR I guess

Let me know what you think @FabianHofmann @fneum


pypsa/plot.py
```diff
@@ -41,244 +46,232 @@
...
+            X data with added jitter.
+        y_jittered : numpy.ndarray
+            Y data with added jitter.
+        """
+        self.x = self.x + np.random.uniform(low=-jitter, high=jitter, size=len(self.x))
+        self.y = self.y + np.random.uniform(low=-jitter, high=jitter, size=len(self.y))
+
+        return self.x, self.y
+
+    def get_multiindex_busses(
```
FabianHofmann, line 266:
```suggestion
    def get_multiindex_buses(
```
and perhaps name it `draw_...`

pypsa/plot.py
```diff
@@ -298,125 +291,47 @@ def plot(
...
-        if bus_cmap is not None and c.dtype is np.dtype("float"):
-            if isinstance(bus_cmap, str):
-                bus_cmap = plt.get_cmap(bus_cmap)
-            if not bus_norm:
-                bus_norm = plt.Normalize(vmin=c.min(), vmax=c.max())
-            c = c.apply(lambda cval: bus_cmap(bus_norm(cval)))
-
-        for b_i in s.index[(s != 0) & ~s.isna()]:
-            radius = s.at[b_i] ** 0.5
+    def get_singleindex_busses(self, sizes, colors, alpha):
```
FabianHofmann, line 305:
```suggestion
    def get_singleindex_buses(self, sizes, colors, alpha):
```

pypsa/plot.py
```diff
@@ -41,244 +46,232 @@
...
+        self.ax.add_feature(
+            cartopy.feature.COASTLINE.with_scale(resolution),
+            linewidth=0.3,
+            edgecolor=geomap_colors.get("coastline", "black"),
+        )
+
+    def init_axis(self, ax, projection, geomap, geomap_colors, title):
+        # Set up plot (either cartopy or matplotlib)
+
+        transform = get_projection_from_crs(self.n.srid)
```
FabianHofmann, line 199: rather use `self.n.crs` here

pypsa/plot.py
```diff
@@ -430,151 +345,507 @@ def plot(
...
-            b_collection.set_array(np.asarray(b_nums))
-            b_collection.set_cmap(b_cmap)
-            b_collection.autoscale()
-            b_collection.set(norm=b_norm)
+        return branch_coll
 
-        ax.add_collection(b_collection)
-        b_collection.set_zorder(3)
-        branch_collections.append(b_collection)
+    def get_branch_collections(self, branch_components, geometry, branch_data):
```
FabianHofmann, line 358: and then this we can name `draw_branches`

FabianHofmann (commented):
this is a very good direction! Just looked through it and I think we are on a good track. some minor comments below. I guess the plan is to use the NetworkPlotter (which we could also name NetworkMapPlotter if that is specifying its purpose) for `n.iplot` as well?

pypsa/plot.py
```diff
@@ -430,151 +345,507 @@ def plot(
...
+    # Plot branches
+    branches = plotter.get_branch_collections(
+        branch_components,
+        geometry,
+        branch_data,
+    )
+    for branch in branches.values():
+        plotter.ax.add_collection(branch)
+
+    # Plot flows
```
FabianHofmann, line 833: this still feels a bit weird that we are defining all branch components separately to then put them together into the get_branch_collections function where we iterate over them again. I would vote for making get_branch_collection process one branch component at a time

fneum:
> I would also change `geomap` to `disable_geomap` and add a `geomap_res` argument, instead of mixing types

Fine with renaming, but maybe in order to not swap meaning of True/False consider `enable_geomap` instead of `disable_geomap`.

>       * Maybe even drop `geomap` and include it into `projection`

This should make sense. Geo-features are then automatically activated if `projection is not None or hasattr(ax, "projection")`.

> 
>     * branch components
>       
>       * For all branch components we have many duplicated arguments, this could be simplified

I think arguments for branch components should remain separately configurable. At some point we had an aggregated argument for all branch components where one had to supply a dictionary with keys for "Line", "Link", etc. or a pd.MultiIndex with the first level referring to the compontent. We preferred the flat format with duplication.

>       * `branch_components` could be dropped and controlled via attribute arguments

Yes, should not be necessary, especially if no elements are created when link_widths=0 (skipped by width threshold).

>     * This is mainly a clean and split up of all the things happened before in `n.plot()`
>       
>       * More simplification with generalized branch component Collections could be added in the background without API changes
>       * Collection return
>         
>         * The collection return is not very intuitive
>         * Right now I changed it to a dict, but we could also return a custom class here, that would also allow a lot of other functionality. But this would be another PR I guess

Would not prioritize the collections. I think few people are accessing this afterwards, but dict seems to be a good idea.

lkstrp:
Closed via #1156

PR #913, opened 2024-06-05
refactor: remove unnecessary module metadata

lkstrp:
I suggest to keep the python author/copyright metadata only in the package `__init__.py`. This is cleaner and they don't differ (only slightly in years) in any module anyway.
To become REUSE compatible (do we want to?) the licence is missing anyway. We could everything to SPDX then


fneum (approved):
Fine by me! Yes, to REUSE compatibility as a target. We have it for PyPSA-Eur already.

PR #917, opened 2024-06-06
Docs contributing

### LINKED ISSUE #899: Use ruff

lkstrp:
I started to improve the Contributing Documentation a bit: Added a `CONTRIBUTIONS.md` to get the GitHub functionality, merged the Unit Testing section, split Help & References in the toctree etc.

In general, we have a lot of similar contribution pages in the docs of all packages: PyPSA, PyPSA-eur, Atlite, linopy. Sometimes they are more or less up to date, but mostly contain similar information.  In future we could write one main page, link to it in all packages, and keep only the package-specific stuff in each package's doc. 

Closes #899


doc/contributing.rst
```diff
@@ -2,12 +2,211 @@
...
+=====
+
+**Basic workflow**
+
+#. Clone the repository: ``git clone https://github.com/PyPSA/PyPSA.git``
+#. Install with dependencies in editable mode: ``pip install -e .[dev]``
+#. Setup linter and formatter (see :ref:`linting-and-formatting`)
+#. Write your code
+#. Run the tests: ``pytest`` (see :ref:`testing`)
+#. Push your changes and create a pull request
```
FabianHofmann, line 37: most people don't have access to push a branch on the upstream. perhaps expand on the steps that the user has to fork the repo, pull from there and propose the changes to the upstream? 

there might be some other authentication trick, but AFAIK cloning with https does not allow you to push. so rather use `git clone git@github.com...`

doc/contributing.rst
```diff
@@ -2,12 +2,211 @@
...
+
+.. _code-contributions:
+
+Code
+=====
+
+**Basic workflow**
+
+#. Clone the repository: ``git clone https://github.com/PyPSA/PyPSA.git``
+#. Install with dependencies in editable mode: ``pip install -e .[dev]``
```
FabianHofmann, line 33: can we add pre-commit to the `dev` installations (https://github.com/PyPSA/PyPSA/blob/4fee8b5238cd7dab79f19f300b21377f1f72b27d/pyproject.toml#L45) and add a step to activate pre-commit ie.`pre-commit install`

FabianHofmann (changes requested):
good one! some minor comments below. I think the better the instructions how to contribute the more contributions we can expect.

doc/contributing.rst
lkstrp: Ah I didn't know, why is most of the team then doing it? Using your fork would keep the branches cleaner as well.

Via HTTPS and when you push, GitHub will prompt you to login. I think that is easier for most people then setting up ssh keys, but we can mention both.

We just need to be careful to have a good mix of information specific to our workflow and those general git/ setup related things.

I'll push some changes

FabianHofmann: You are right, we should try to avoid mixed introductions. However, for all people that have push access, I would keep it the way as we handled it (directly push branches to the upstream). It makes things a bit easier: the CI is automatically triggered, the reviewer can easily check out the branches. But for all non-internal users (to whom this section is likely addressed) the forking strategy is likely the most feasible way to go.

lkstrp: Sounds good.

Only to leave that info here: The CI is triggered also for fork branches of members, and we can also checkout pull requests via `git checkout pr/xxx` and use the issue number. See [here](https://stackoverflow.com/a/27567939/25438288)

FabianHofmann: good to know (vscode's github extension is also good for checking out pr's)

PR #944, opened 2024-06-26
Fix flake8 E203 warnings (whitespace before ':')

bje-:
Self-explanatory. `diff -w` verifies that there are no non-whitespace changes.


lkstrp:
Hey @bje-,
we use Ruff and E203 is [not supported yet](https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/). As you see, the formatter reverts those changes.

PR #946, opened 2024-06-26
optimization/abstract: fix security-constrained optimization

FabianHofmann:
@martacki could you try this one out? I don't expect it to be correct at the first shot, it might well be that we have to iterate a bit. 

BTW: It is hard to create a unit test for this implementation. To realize that we would need at best a pypsa-eur network with only a few snapshots but relatively high number of nodes (64), perhaps that is something we can include to the examples @lkstrp ?   

## Checklist


martacki (approved):
tested with a complex European network containing multiple synchronous zones, for multiple snapshots and rolling horizon windows, all tests have passed. Thank you @FabianHofmann  :star: :rocket:

martacki:
For completeness, the issue was that the current implementation did not support multiple synchronous zones. Networks with one single synchronous zone / sub-network resulted in the exact same solution as the old `network_sclopf` (which is to be removed in future releases and already does not exist in `master`). However, `N-1` stability was not secured in more complex topologies with multiple synchronous zones / sub-networks.

This PR should resolve the issue.

PR #960, opened 2024-07-03
docs: api reference overhaul

lkstrp:
Restructured the API Reference section:
- Add subpages for content with summaries based on `autosummary`.
- Creates a subpage for each method (similar to [here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.index.html))

-> Makes it much easier to browse the API references and avoids this super large one pager.

Disadvantage at the moment is that there is no automatic expansion of methods/attributes with `autosummary`. This would benefit from some custom extension or testing in the future. I've experimented a bit, but this needs more time. Is also just nice to have.


fneum (approved):
I like the separation into pages. Disadvantage is that you have to list individual methods from the module? Possibly harder to maintain; i.e. forget new methods.

lkstrp:
> I like the separation into pages. Disadvantage is that you have to list individual methods from the module? Possibly harder to maintain; i.e. forget new methods.

Yes I played around with some custom extensions which allows auto expanding. I'm just not happy yet. Will add something like this and/ or tests at some point.

PR #886, opened 2024-05-14
Add parameters to statistics

p-glaum:
Similar to matplotlib where you can set rc parameters, this feature is implemented for statistics. With this PR you can set the parameter for rounding values, showing zero values and nice names in statistics. E.g., `pypysa.statistics.set_parameters(nice_names=False)` which then calls all statistics methods with the attribute nice_names set to False as default. Furthermore, you can see available groupers by calling statistics.groupers and the set parameters by calling statistics.parameters

## Checklist

- [x] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [x] Unit tests for new features were added (if applicable).
- [x] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [x] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #961, opened 2024-07-03
refactor: `statistics.py` structure

lkstrp:
- make functions which are internal actually internal
- move grouper methods to grouper

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -50,8 +50,6 @@ def get_weightings(n, c):
         return n.snapshot_weightings["objective"]
 
 
-
-
 def _port_efficiency(n, c, port=""):
     if (port == "") or "efficiency" not in n.df(c):
         efficiency = 1
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -14,28 +14,6 @@ from pypsa.descriptors import nominal_attrs
 logger = logging.getLogger(__name__)
 
 
-def get_carrier(n, c, nice_names=True):
-    """
-    Get the nice carrier names for a component.
-    """
-    df = n.df(c)
-    fall_back = pd.Series("", index=df.index)
-    carrier_series = df.get("carrier", fall_back).rename("carrier")
-    if nice_names:
-        carrier_series = carrier_series.replace(
-            n.carriers.nice_name[lambda ds: ds != ""]
-        ).replace("", "-")
-    return carrier_series
-
-
-def get_bus_and_carrier(n, c, port="", nice_names=True):
-    """
-    Get the buses and nice carrier names for a component.
-    """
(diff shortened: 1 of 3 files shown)


FabianHofmann:
I am actually not sure, if all the `get` functions are actually "internal". I think that an import from `pypsa.statistics` is not uncommon, as for example here https://github.com/PyPSA/pypsa-eur/pull/921/files#diff-6dafdba3772c0812a6a645c461f092bdbf7e0d7447be19ec5ea4c071e33e5302

Perhaps we make this non-breaking?

euronion:
> I am actually not sure, if all the `get` functions are actually "internal". I think that an import from `pypsa.statistics` is not uncommon, as for example here https://github.com/PyPSA/pypsa-eur/pull/921/files#diff-6dafdba3772c0812a6a645c461f092bdbf7e0d7447be19ec5ea4c071e33e5302
> 
> Perhaps we make this non-breaking?

I second.

lkstrp:
You both mean the groupers? I did not make them internal.
But there is no reason to have them that often and not consistent in the API:
- via `pypsa.statistics.Groupers.get_carriers`
- via `pypsa.statistics.get_carriers`
- via `n.statistics.groupers.get_carriers`

What do you think about keeping only
- via `pypsa.statistics.groupers.get_carriers`
- via `n.statistics.groupers.get_carriers`

and deprecating `pypsa.statistics.get_carriers` instead of removing?

FabianHofmann:
> You both mean the groupers? I did not make them internal. But there is no reason to have them that often and not consistent in the API:
> 
> * via `pypsa.statistics.Groupers.get_carriers`
> * via `pypsa.statistics.get_carriers`
> * via `n.statistics.groupers.get_carriers`
> 
> What do you think about keeping only
> 
> * via `pypsa.statistics.groupers.get_carriers`
> * via `n.statistics.groupers.get_carriers`
> 
> and deprecating `pypsa.statistics.get_carriers` instead of removing?

yes, that sounds like a good middle ground. Initially, I was referring to the non-strict grouper functions like  `_get_transmission_branches` which now moved to internal, but will very likely be used as official functions. Perhaps we move them to descriptors.py and keep them as official?

lkstrp:
It would make sense too pull #885 and #886 by @p-glaum first. I think they are good to go?

p-glaum:
> It would make sense too pull #885 and #886 by @p-glaum first. I think they are good to go?

I guess #885 is ready to merge if there are no further comments and #886 still needs someone to review.

fneum:
I just reviewed #885 and #886. Both require some changes or clarifications.

fneum:
#885 and #886 are now both merged

fneum:
Revise and merge after #1075, #1044

PR #968, opened 2024-07-08
build: improve test and release ci and remove env files

lkstrp:
All dependencies are managed in `pyproject.toml`, so I would stop keeping the double structure. RTD no longer needs one either.

EDIT:
Also updates CI release and test workflow. Applies same changes as already done in linopy (https://github.com/PyPSA/linopy/pull/328) and powerplantmatching (https://github.com/PyPSA/powerplantmatching/pull/182).


fneum (changes requested):
Fine by me, but requires elimination of all uses of these files:

https://github.com/PyPSA/PyPSA/blob/f8ae11052253a6306ec5f8d7a0d6bc35828b76f7/.github/workflows/test.yml#L93

lkstrp:
@fneum 
What is the purpose of the micromamba test ci anyway? It basically just tests if the package works with `environment_dev.yaml`. Nothing more, since the package installation is done with pip.

I think this can be removed completely. Also speeds up the ci.

I also ping @euronion, I think he wrote this.

fneum:
No problem to remove from my side.

PR #971, opened 2024-07-09
statistics: disable bus_carrier warning when `at_port` is false

FabianHofmann:
## Checklist

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -920,6 +920,7 @@ class StatisticsAccessor:
             n.buses.carrier.unique().size > 1
             and groupby is None
             and bus_carrier is None
+            and at_port
         ):
             logger.warning(
                 "Network has multiple bus carriers which are aggregated together. To separate bus carriers set `bus_carrier` or use groupers like `get_carrier_and_bus_carrier`."


p-glaum:
Is it obsolete? Because for a link component, one could say that bus 0 is electricity but it could also be gas for another link.

FabianHofmann:
> Is it obsolete? Because for a link component, one could say that bus 0 is electricity but it could also be gas for another link.

But it is quite unlikely that the same link carrier has different bus-carriers at bus0. The reason why I suggest this is that currently there is no way to calculate the curtailment in a network with different bus carriers without getting the warning. However, curtailment is a quantity that is related to a technology that is independent of the bus carrier it is primarily attached to.

p-glaum:
What do you call exactly to get this warning? Because for curtailment this warning should not appear.

> > Is it obsolete? Because for a link component, one could say that bus 0 is electricity but it could also be gas for another link.
> 
> But it is quite unlikely that the same link carrier has different bus-carriers at bus0. The reason why I suggest this is that currently there is no way to calculate the curtailment in a network with different bus carriers without getting the warning. However, curtailment is a quantity that is related to a technology that is independent of the bus carrier it is primarily attached to.

What do you call exactly to get this warning? Because for curtailment this warning should not appear.

FabianHofmann:
@p-glaum sorry, my bad, it was the `market_value`. I think this might be actually be related to a different case (and the proposed change is not fully fix it). But if you have a second to think about it: 

At the moment, we give out the market value as the revenue divided by the supply. Both will either be calculated 
- at individual ports to get the MV per carrier and bus carrier or 
- at the first port only to get the MV per carrier, which leads to the warning. 

However, as far as I see, we should be able to get a market value for each technology (second case) as a total for multi-links. For generators, the market value is defined as in toms paper 

image

If we now have multiple inputs/outputs, the market value should still be normed by the operational variable, ie. p0. Then, the cost equality in equation (16) is still true for a perfect market. @nworbmot perhaps you can also have a short look, hope I made myself clear. 

So, if that is right, we would need to disable the warning as it is and add a `at_port = False` in the `supply` function in `market_value`

p-glaum:
Hi Fabian,
I see what you mean. I guess in this case we should indeed change this warning as suggested in the PR. However, I think we also need to change the way market value is calculated. Because for the default, when we do not define a bus_carrier, we should get the overall revenue considering all ports, i.e. at_port=True, while we only take for withdrawal(at_port=False), i.e. p0.
So the market value has the unit of bus_carrier 0.
In case we define a bus carrier, the MV only considers the revenue for this bus carrier and the supply/withdrawal for the bus_carrier.

In the future we could take the energy balance instead of supply/withdrawal, but currently the default output of the energy balance has the wrong inputs. Dispatch has the correct index, but you would always bet a deprecated warning.

FabianHofmann:
@p-glaum would you have an idea on how to get that in then? I guess in the market values, the supply needs to have at_port=False, except for when bus_carriers is defined? It is difficult to grasp.

pypsa/statistics.py
```diff
@@ -1237,7 +1238,7 @@ def market_value(
             bus_carrier=bus_carrier,
             nice_names=nice_names,
         )
-        df = self.revenue(**kwargs) / self.supply(**kwargs)
+        df = self.revenue(**kwargs) / self.supply(**{**kwargs, "at_port": False})
```
p-glaum, line 1241: One idea would be to use `energy_balance` here to not get 0 values for `supply` of links because p0 is always negative. 
So maybe we could change this line to :
```python
update_kwargs = {"at_port": False} if bus_carrier is None else {}
df = self.revenue(**kwargs) / self.energy_balance(**{**kwargs, **update_kwargs}).abs()
```

p-glaum: Then we get the market_value in units of bus0 by default, and only when selecting a bus_carrier do we get the market_value for this bus_carrier in units of the bus_carrier.

For example, for electrolysis, where bus0 is electricity and bus1 is hydrogen:

- Default (bus_carrier=None) market_value: Revenue considering the purchase price of electricity and the selling price of hydrogen (positive revenue), divided by the utilization of p0, i.e., electricity consumption. Hence, we get €/MWh_el.
- bus_carrier="AC" (electricity): Revenue only considering the purchase price of electricity (negative revenue), divided by electricity consumption. Hence, we again get €/MWh_el, but negative, because for electricity we usually need to pay, unless prices are negative.
- bus_carrier="H2": Revenue only considering hydrogen, which is sold (positive revenue), divided by hydrogen generation. Here we get €/MWh_H2, which is positive.

lkstrp:
Whats the status here @FabianHofmann @p-glaum ?

PR #977, opened 2024-07-11
pyproject.toml: optional dependency 'geomap' including cartopy and re…

fneum:


pyproject.toml
```diff
@@ -50,7 +50,10 @@ dev = [
     "pre-commit", 
     "ruff"
 ]
-cartopy = ["cartopy>=0.16"]
+geomap = [
```
lkstrp, line 53: #912 drops requests as an dependency, so I wouldn't rename it
```suggestion
cartopy = [
```

lkstrp (changes requested):
#912 drops requests as a needed dependency completly,

lkstrp (approved):

Issue #848, opened 2024-03-01
Remove "-ext" and "-fix" suffix in index names in index getter functions


FabianHofmann:
Currently, the functions n.get_extendable_i and similar return a pandas Series with an index named after the component and the type, i.e. "-ext" for extendable and so on. This naming was needed in previous versions of linopy (<0.2.) where dimension names needed to be distinct across non-overlapping axis. 

It would however so much more convenient to have consistent axis names -> just use the component name.

Operations like 

```python 
fix_capacity = n.df(c).p_nom[n.get_non_extendable_i(c)]
ext_capacity = n.model.variables[f"{c}-p_nom"]
capacity = fix_capacity + ext_capacity
```
would then be possible

lkstrp:
Just an update on this @fneum @FabianHofmann 

While this is theoretically easy to fix by removing the suffix, we get huge problems with multi-indexed snapshots because pandas chokes dimensions (`snapshot`) vs coords (`period`, `timestep`). So I would suggest to only resolve this in #1154 via xarray.

But that means we would deviate from the ideal case of no functional changes in #1154. We can still try to keep it separate as much as possible (e.g. put the default value changes in a different PR (see [here](https://github.com/PyPSA/PyPSA/pull/1166/files#diff-b62bc9137d827b90a7f2de1ec8090bbe793751f3a0ba60e7bccf1fc05d5cfd73)), but we cannot do it completely

PR #988, opened 2024-07-23
skip create variables for non-extendable links/generators with `p_set`

fneum:
Potential performance improvement, especially for myopic pathway planning.

The idea is to skip creating the decision variables for links and generators with a fixed dispatch profile defined by `p_set` that is added to the RHS of the energy balance constraints.

So far, the variables have been created, and an equality constraint has enforced the fixed profile.

This could be applied to VRE generators. The curtailment could be modeled with a single "curtailment" generator. (https://github.com/PyPSA/pypsa-eur/pull/1177)

```py
n.add(
    "Generator",
    "curtailment",
    bus="electricity"
    p_min_pu=-1,
    p_max_pu=0,
    marginal_cost=-0.1,
    carrier="curtailment",
    p_nom=1e6,
)
```

The first tests indicate that (unfortunately for this PR) the presolving of Gurobi is very good. The following approach that does not remove the variables from the optimization problem yields near-identical solving times:

```
n.generators_t.p_min_pu = n.generators_t.p_max_pu
```

xref https://github.com/PyPSA/pypsa-eur/pull/1056

- [ ] perhaps extend logic to other components `{StorageUnit, Store}`


FabianHofmann:
Hey @fneum, nice idea! I am fearing this make the code a bit more complicated as it already is and I am wondering whether the code changes are worth the (potential) speedup.  Generally, I have the feeling such things would be much easier when  #848 is tackled (perhaps we should prioritize this). 

As an alternative, I will think about whether it would be easy to make a presolve in linopy which replaced fixed variables by constants.

fneum:
The code changes per se are not very extensive and would be acceptable if there were a benefit.

Still, I would drop this PR either way because fixed variables are, in fact, **fully** captured by Presolve of HiGHS, Gurobi, and likely all other solvers (i.e. same rows/columns after presolve with or without this PR). It only takes a few seconds, even for larger problems.

FabianHofmann:
okay, thanks, the idea was worth the try!

PR #996, opened 2024-08-01
test: assert package version found

lkstrp:


pypsa/__init__.py
```diff
@@ -35,9 +35,9 @@
 # e.g. "0.17.0" # TODO, in the network structure it should use the dev version
 release_version = re.match(r"(\d+\.\d+(\.\d+)?)", __version__).group(0)
 
-# Assert that version is not 0.1 (which is the default version in the setup.py)
-assert release_version != "0.1", "setuptools_scm could not find the version number"
-
+assert not __version__.startswith(
+    "0.0"
+), "Could not determine version of powerplantmatching."
```
FabianHofmann, line 40: I guess that should be "pypsa"? :)

lkstrp: :(

PR #1004, opened 2024-08-19
doc update and revisions (without examples)

fneum:
Documentation update and revisions (without examples)


doc/contributing/troubleshooting.rst
```diff
@@ -27,13 +28,13 @@ See :ref:`upgrading-pypsa`.
 Consistency check on network
 ============================
 
-A consistency check can be performed using the function :py:meth:`pypsa.Network.consistency_check`.
+A consistency check can be performed using the function :py:meth:`pypsa.n.consistency_check`.
```
lkstrp, line 31: If you rename just the reference the link will break. We could still rename it and keep the reference via 
```
:py:meth:`n.consistency_check <pypsa.Network.consistency_check>`
```

I have a mini helper script somewhere for those things, will check tomorrow

doc/contributing/troubleshooting.rst
```diff
@@ -27,13 +28,13 @@ See :ref:`upgrading-pypsa`.
...
 
 
 Problems with power flow convergence
 ====================================
 
-If your ``network.pf()`` is not converging there are two possible reasons:
+If your ``n.pf()`` is not converging there are two possible reasons:
```
lkstrp, line 37: Same for those, we could use references here so you can straight jump to the docstring

lkstrp (approved):
Good to go, just two comments FYI.
I will fix all references with some regex in another PR. Would be nice to have them everywhere.
And the test should show if those refs can not be resolved.

PR #1007, opened 2024-08-21
Make extendability and committability compatible, when modularity is on

### LINKED ISSUE #1000: Make extendable components committable when modularity is used
@davide-f and I would like to modify the mathematical formulation of extendability and committability to make these two features compatible. Currently, the system works with a set of eqs. like:
 
$p_{nom} * p_{min, t} * status_t \leq p_t \leq p_{nom} * p_{max, t} * status_t$
 
$p_{nom}$ and $status_t$ cannot be optimized simultaneously and to address this, we propose the following changes:
1. Change the definition of $`status_t`$ from binary to integer, where $`0 \leq status_t \leq n_{mod}`$
2. Modify the eqs to: $`p^{mod}_{nom} * p_{min, t} * status_t \leq p_t \leq p^{mod}_{nom} * p_{max, t} * status_t`$
3. Update related features such as start-up and shut-down costs according to point 1.
 
To maintain compatibility with previous versions, we will limit these changes to cases where $`p^{mod}_{nom}`$  is provided as input. Despite this, the change mentioned in point 1 could make modularity compatible with committability, which it currently is not.
 
Any suggestions for improvements or points to consider? Anyone interested in helping with this?

enricogiglio:
Closes #1000 

@davide-f and I would like to propose a MILP formulation for extendability and committability, only when modularity is on, to make these two features compatible. We have also added a test for the feature we have implemented, and believe that the proposed changes are consistent with the rest of the codebase. In particular, the goal is to keep everything compatible with the previous version without introducing breaking changes.

At the moment we'd like to continue working on improving the proposed code changes, but perhaps we'll also propose a new test that's more accurate with respect to the changes introduced.

## Checklist

- [X] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.

### DIFF
--- a/pypsa/consistency.py
+++ b/pypsa/consistency.py
@@ -17,6 +17,14 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+nominal_attrs = {
+    "Generator": "p_nom",
+    "Line": "s_nom",
+    "Transformer": "s_nom",
+    "Link": "p_nom",
+    "Store": "e_nom",
+    "StorageUnit": "p_nom",
+}
 
 def _bus_columns(df: pd.DataFrame) -> pd.Index:
     return df.columns[df.columns.str.startswith("bus")]
@@ -314,10 +322,11 @@ def check_assets(network: Network, component: Component) -> None:
     if component.name in {"Generator", "Link"}:
         committables = network.get_committable_i(component.name)
         extendables = network.get_extendable_i(component.name)
-        intersection = committables.intersection(extendables)
+        not_mod_i = network.df(component.name).query(f"({nominal_attrs[component.name]}_mod==0)").index
+        intersection = committables.intersection(extendables).intersection(not_mod_i)
         if not intersection.empty:
             logger.warning(
-                "Assets can only be committable or extendable. Found "
+                "Assets can only be committable or extendable, if not modular. Found "
                 f"assets in component {component.name} which are both:"
                 f"\n\n\t{', '.join(intersection)}"
             )
--- a/pypsa/optimization/constraints.py
+++ b/pypsa/optimization/constraints.py
@@ -8,7 +8,7 @@ import logging
 import linopy
 import pandas as pd
 from linopy import LinearExpression, merge
-from numpy import inf, isfinite
+from numpy import inf, isfinite, round
 from scipy import sparse
 from xarray import DataArray, Dataset, concat
 
@@ -25,12 +25,12 @@ from pypsa.optimization.common import reindex
 logger = logging.getLogger(__name__)
 
 
-def define_operational_constraints_for_non_extendables(
+def define_operational_constraints_for_non_extendables_and_non_committables(
     n, sns, c, attr, transmission_losses
 ):
     """
     Sets power dispatch constraints for non-extendable and non-commitable
-    assets for a given component and a given attribute.
+    assets for a given component and a given attribute,  whether they are modular or not.
 
     Parameters
     ----------
@@ -62,19 +62,19 @@ def define_operational_constraints_for_non_extendables(
         dispatch_lower = (1, dispatch_lower), (-1, loss)
         dispatch_upper = (1, dispatch_upper), (1, loss)
     n.model.add_constraints(
-        dispatch_lower, ">=", lower, name=f"{c}-fix-{attr}-lower", mask=active
+        dispatch_lower, ">=", lower, name=f"{c}-fix-non-comm-{attr}-lower", mask=active
     )
     n.model.add_constraints(
-        dispatch_upper, "<=", upper, name=f"{c}-fix-{attr}-upper", mask=active
+        dispatch_upper, "<=", upper, name=f"{c}-fix-non-comm-{attr}-upper", mask=active
     )
 
 
-def define_operational_constraints_for_extendables(
+def define_operational_constraints_for_extendables_but_non_committables(
     n, sns, c, attr, transmission_losses
 ):
     """
-    Sets power dispatch constraints for extendable devices for a given
-    component and a given attribute.
+    Sets power dispatch constraints for extendable but non commitable devices for a given
+    component and a given attribute,  whether they are modular or not.
 
     Parameters
     ----------
@@ -87,7 +87,8 @@ def define_operational_constraints_for_extendables(
         name of the attribute, e.g. 'p'
     """
     ext_i = n.get_extendable_i(c)
-
+    ext_i = ext_i.difference(n.get_committable_i(c)).rename(ext_i.name)
+    
     if ext_i.empty:
         return
 
@@ -105,16 +106,409 @@ def define_operational_constraints_for_extendables(
         lhs_upper += ((1, loss),)
 
     n.model.add_constraints(
-        lhs_lower, ">=", 0, name=f"{c}-ext-{attr}-lower", mask=active
+        lhs_lower, ">=", 0, name=f"{c}-ext-non-comm-{attr}-lower", mask=active
     )
     n.model.add_constraints(
-        lhs_upper, "<=", 0, name=f"{c}-ext-{attr}-upper", mask=active
+        lhs_upper, "<=", 0, name=f"{c}-ext-non-comm-{attr}-upper", mask=active
     )
 
-
-def define_operational_constraints_for_committables(n, sns, c):
+def  define_operational_constraints_for_extendables_and_committables_but_non_modular(
+    n, sns, c, attr, transmission_losses
+):
     """
-    Sets power dispatch constraints for committable devices for a given
+    Sets power dispatch constraints for extendable, committable, but non modular devices for a given
+    component and a given attribute. Extendability is preferred over committability (committability
+    is simply ignored).
+
+    Parameters
+    ----------
+    n : pypsa.Network
+    sns : pd.Index
+        Snapshots of the constraint.
+    c : str
+        name of the network component
+    attr : str
+        name of the attribute, e.g. 'p'
+    """
+    ext_i = n.get_extendable_i(c)
+    com_i = n.get_committable_i(c)
+    not_mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod==0)").index
+
+    ext_i = ext_i.intersection(com_i).intersection(not_mod_i).rename(ext_i.name)
+    
+    if ext_i.empty:
+        return
+
+    min_pu, max_pu = map(DataArray, get_bounds_pu(n, c, sns, ext_i, attr))
+    dispatch = reindex(n.model[f"{c}-{attr}"], c, ext_i)
+    capacity = n.model[f"{c}-{nominal_attrs[c]}"]
+
+    active = get_activity_mask(n, c, sns, ext_i) if n._multi_invest else None
+
+    lhs_lower = (1, dispatch), (-min_pu, capacity)
+    lhs_upper = (1, dispatch), (-max_pu, capacity)
+    if c in n.passive_branch_components and transmission_losses:
+        loss = reindex(n.model[f"{c}-loss"], c, ext_i)
+        lhs_lower += ((-1, loss),)
+        lhs_upper += ((1, loss),)
+
+    n.model.add_constraints(
+        lhs_lower, ">=", 0, name=f"{c}-ext-comm-non-mod-{attr}-lower", mask=active
+    )
+    n.model.add_constraints(
+        lhs_upper, "<=", 0, name=f"{c}-ext-comm-non-mod-{attr}-upper", mask=active
+    )
+
+def define_committability_variables_constraints_for_non_modular(n, sns, c, attr):
+    """
+    This function sets the upper limit of committable variables (status, start-up, shut-down) to 1 for all
+    committable components for which modularity is not used, regardless of whether they are extendable or not.
+    _____
+
+    Parameters
+    ----------
+    n : pypsa.Network
+    sns : pd.Index
+        Snapshots of the constraint.
+    c : str
+        name of the network component
+    attr : str
+        name of the attribute, e.g. 'p_nom'
+    """
+    com_i = n.get_committable_i(c)
+    not_mod_i = n.df(c).query(f"({attr}_mod==0)").index
+    com_i = com_i.intersection(not_mod_i).rename(com_i.name)
+
+    if com_i.empty:
+        return
+
+    m = n.model
+    ext_i = n.get_extendable_i(c)
+    active = get_activity_mask(n, c, sns, ext_i) if n._multi_invest else None
+
+    status = m.variables[f"{c}-status"].loc[sns,com_i]
+    n.model.add_constraints(
+        status, "<=", 1, name=f"{c}-status-{attr}-upper-nonmod", mask=active
+    )
+    start_up = m.variables[f"{c}-start_up"].loc[sns, com_i]
+    n.model.add_constraints(
+        start_up, "<=", 1, name=f"{c}-start_up-{attr}-upper-nonmod", mask=active
+    )
+
+    shut_down = m.variables[f"{c}-shut_down"].loc[sns,com_i]
+    n.model.add_constraints(
+        shut_down, "<=", 1, name=f"{c}-shut_down-{attr}-upper-nonmod", mask=active
+    )
+
+def define_committability_variables_constraints_for_modular_and_non_extendables(n, sns, c, attr):
+    """
+    This function sets the upper limit of committable variables (status, start-up, shut-down) to the
+    installed number of modules for all committable and non-extendable components. The number of modules
+    is calculated as the nominal power divided by the nominal dimension of the individual module (e.g. p_nom/p_nom_mod).
+
+    If the number of modules is not an integer number, the function returns an error message.
+    _____
+
+    Parameters
+    ----------
+    n : pypsa.Network
+    sns : pd.Index
+        Snapshots of the constraint.
+    c : str
+        name of the network component
+    attr : str
+        name of the attribute, e.g. 'p_nom'
+    """
+    com_i = n.get_committable_i(c)
+    fix_i = n.df(c).query(f"not {attr}_extendable").index
+    mod_i = n.df(c).query(f"({attr}_mod>0)").index
+
+    inter_i = com_i.intersection(mod_i).intersection(fix_i).rename(com_i.name)
+    
+    if inter_i.empty:
+        return
+
+    m = n.model
+    
+    ext_i = n.get_extendable_i(c)
+    active = get_activity_mask(n, c, sns, ext_i) if n._multi_invest else None
+
+    n_mod = n.df(c)[attr].loc[inter_i]/n.df(c)[attr + "_mod"].loc[inter_i]
+    diff_n_mod = abs(n_mod - round(n_mod))
+    non_integers_n_mod_i = diff_n_mod[diff_n_mod > 10**-6].index
+    
+    if  not non_integers_n_mod_i.empty:
+        msg = (
+        "For non-extendable but committable assets, if both p_nom and p_nom_mod are declared, p_nom must be a"
+        "must be a multiple of p_nom_mod. Found "
+                f"assets in component {c} do not respect this criterion:"
+                f"\n\n\t{', '.join(non_integers_n_mod_i)}"
+        )
+        raise ValueError(msg)
+    
+    
+    rhs = pd.DataFrame(0, sns, inter_i)
+    rhs.loc[sns, inter_i] = n_mod.loc[inter_i].values
+
+    status = m.variables[f"{c}-status"].loc[sns,inter_i]
+    n.model.add_constraints(
+        status, "<=", rhs, name=f"{c}-status-{attr}-upper-mod_nonext", mask=active
+    )
+
+    start_up = m.variables[f"{c}-start_up"].loc[sns, inter_i]
+    n.model.add_constraints(
+        start_up, "<=", rhs, name=f"{c}-start_up-{attr}-upper-mod_nonext", mask=active
+    )
+
+    shut_down = m.variables[f"{c}-shut_down"].loc[sns, inter_i]
+    n.model.add_constraints(
+        shut_down, "<=", rhs, name=f"{c}-shut_down-{attr}-upper-mod_nonext", mask=active
+    )
+
+def define_committability_variables_constraints_for_modular_and_extendables(n, sns, c, attr):
+    """
+    This function sets the upper limit of committable variables (status, start-up, shut-down) to the
+    variable n_mod for all committable, extendable and modular components. 
+    _____
+
+    Parameters
+    ----------
+    n : pypsa.Network
+    sns : pd.Index
+        Snapshots of the constraint.
+    c : str
+        name of the network component
+    attr : str
+        name of the attribute, e.g. 'p_nom'
+    """
+    com_i = n.get_committable_i(c)
+    ext_i = n.df(c).query(f"{attr}_extendable").index
+    mod_i = n.df(c).query(f"({attr}_mod>0)").index
+
+    inter_i = com_i.intersection(mod_i).intersection(ext_i).rename(com_i.name)
+
+    if inter_i .empty:
+        return
+
+    m = n.model
+    
+    
+    active = get_activity_mask(n, c, sns, ext_i) if n._multi_invest else None
+
+    n_mod = n.model[f"{c}-n_mod"].loc[inter_i]
+    
+    #ALLERT DIMENSIONI DISEQUAZIONE
+    status = m.variables[f"{c}-status"].loc[sns,inter_i]
+    lhs = ((1, status), (-1, n_mod))
+    n.model.add_constraints(
+        lhs, "<=", 0, name=f"{c}-status-{attr}-upper-mod_ext", mask=active
+    )
+
+    #ALLERT DIMENSIONI DISEQUAZIONE
+    start_up = m.variables[f"{c}-start_up"].loc[sns,inter_i]
+    lhs = ((1, start_up), (-1, n_mod))
+    n.model.add_constraints(
+        lhs, "<=", 0, name=f"{c}-start_up-{attr}-upper-mod_ext", mask=active
+    )
+    
+    #ALLERT DIMENSIONI DISEQUAZIONE
+    shut_down = m.variables[f"{c}-shut_down"].loc[sns,inter_i]
+    lhs = ((1, shut_down), (-1, n_mod))
+    n.model.add_constraints(
+        lhs, "<=", 0, name=f"{c}-shut_down-{attr}-upper-mod_ext", mask=active
+    )
+    
+def define_operational_constraints_for_committables_and_modular(n, sns, c):
+    """
+    Sets power dispatch constraints for committable and modular devices for a given
+    component and a given attribute, whether they are extendable or not. The linearized
+    approximation of the unit commitment problem is inspired by Hua et al. (2017) DOI:
+    10.1109/TPWRS.2017.2735026.
+
+    Parameters
+    ----------
+    n : pypsa.Network
+    sns : pd.Index
+        Snapshots of the constraint.
+    c : str
+        name of the network component
+    """
+    name_func = "-committable-modular"
+
+    com_i = n.get_committable_i(c)
+    mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod>0)").index
+    com_i = com_i.intersection(mod_i).rename(com_i.name)
+
+    if com_i.empty:
+        return
+    
+    # variables
+    status = n.model[f"{c}-status"].loc[sns, com_i]
+    start_up = n.model[f"{c}-start_up"].loc[sns, com_i]
+    shut_down = n.model[f"{c}-shut_down"].loc[sns, com_i]
+    status_diff = status - status.shift(snapshot=1)
+    p = reindex(n.model[f"{c}-p"], c, com_i)
+    
+    active = get_activity_mask(n, c, sns, com_i) if n._multi_invest else None
+
+    # parameters
+    nominal = DataArray(n.df(c)[nominal_attrs[c]+'_mod'].reindex(com_i))
+    min_pu, max_pu = map(DataArray, get_bounds_pu(n, c, sns, com_i, "p"))
+    lower_p = min_pu * nominal
+    upper_p = max_pu * nominal
+    min_up_time_set = n.df(c).min_up_time[com_i]
+    min_down_time_set = n.df(c).min_down_time[com_i]
+    ramp_up_limit = nominal * n.df(c).ramp_limit_up[com_i].fillna(1)
+    ramp_down_limit = nominal * n.df(c).ramp_limit_down[com_i].fillna(1)
+    ramp_start_up = nominal * n.df(c).ramp_limit_start_up[com_i]
+    ramp_shut_down = nominal * n.df(c).ramp_limit_shut_down[com_i]
+    up_time_before_set = n.df(c)["up_time_before"].reindex(com_i)
+    down_time_before_set = n.df(c)["down_time_before"].reindex(com_i)
+    initially_up = up_time_before_set.astype(bool)
+    initially_down = down_time_before_set.astype(bool)
+
+    # check if there are status calculated/fixed before given sns interval
+    if sns[0] != n.snapshots[0]:
+        start_i = n.snapshots.get_loc(sns[0])
+        # get generators which are online until the first regarded snapshot
+        until_start_up = n.pnl(c).status.iloc[:start_i][::-1].reindex(columns=com_i)
+        ref = range(1, len(until_start_up) + 1)
+        up_time_before = until_start_up[until_start_up.cumsum().eq(ref, axis=0)].sum()
+        up_time_before_set = up_time_before.clip(upper=min_up_time_set)
+        initially_up = up_time_before_set.astype(bool)
+        # get number of snapshots for generators which are offline before the first regarded snapshot
+        until_start_down = ~until_start_up.astype(bool)
+        ref = range(1, len(until_start_down) + 1)
+        down_time_before = until_start_down[
+            until_start_down.cumsum().eq(ref, axis=0)
+        ].sum()
+        down_time_before_set = down_time_before.clip(upper=min_down_time_set)
+        initially_down = down_time_before_set.astype(bool)
+
+    # lower dispatch level limit
+    lhs = (1, p), (-lower_p, status)
+    n.model.add_constraints(lhs, ">=", 0, name=f"{c}-com-mod-p-lower"+name_func, mask=active)
+
+    # upper dispatch level limit
+    lhs = (1, p), (-upper_p, status)
+    n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-mod-p-upper"+name_func, mask=active)
+
+    # state-transition constraint
+    rhs = pd.DataFrame(0, sns, com_i)
+    rhs.loc[sns[0], initially_up] = -1
+    lhs = start_up - status_diff
+    n.model.add_constraints(
+        lhs, ">=", rhs, name=f"{c}-com-mod-transition-start-up"+name_func, mask=active
+    )
+
+    rhs = pd.DataFrame(0, sns, com_i)
+    rhs.loc[sns[0], initially_up] = 1
+    lhs = shut_down + status_diff
+    n.model.add_constraints(
+        lhs, ">=", rhs, name=f"{c}-com-mod-transition-shut-down"+name_func, mask=active
+    )
+
+    # min up time
+    mask = get_activity_mask(n, c, sns[1:], com_i)
+    expr = []
+    min_up_time_i = com_i[min_up_time_set.astype(bool)]
+    if not min_up_time_i.empty:
+        for g in min_up_time_i:
+            su = start_up.loc[:, g]
+            expr.append(su.rolling(snapshot=min_up_time_set[g]).sum())
+        lhs = -status.loc[:, min_up_time_i] + merge(expr, dim=com_i.name)
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(
+            lhs, "<=", 0, name=f"{c}-com-mod-up-time"+name_func, mask=mask[min_up_time_i]
+        )
+
+    # min down time
+    expr = []
+    min_down_time_i = com_i[min_down_time_set.astype(bool)]
+    if not min_down_time_i.empty:
+        for g in min_down_time_i:
+            su = shut_down.loc[:, g]
+            expr.append(su.rolling(snapshot=min_down_time_set[g]).sum())
+        lhs = status.loc[:, min_down_time_i] + merge(expr, dim=com_i.name)
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(
+            lhs, "<=", 1, name=f"{c}-com-mod-down-time"+name_func, mask=mask[min_down_time_i]
+        )
+
+    # up time before
+    timesteps = pd.DataFrame([range(1, len(sns) + 1)] * len(com_i), com_i, sns).T
+    if initially_up.any():
+        must_stay_up = (min_up_time_set - up_time_before_set).clip(lower=0)
+        mask = (must_stay_up >= timesteps) & initially_up
+        name = f"{c}-com-mod-status-min_up_time_must_stay_up"+name_func
+        mask = mask & active if active is not None else mask
+        n.model.add_constraints(status, "=", 1, name=name, mask=mask)
+
+    # down time before
+    if initially_down.any():
+        must_stay_down = (min_down_time_set - down_time_before_set).clip(lower=0)
+        mask = (must_stay_down >= timesteps) & initially_down
+        name = f"{c}-com-mod-status-min_down_time_must_stay_up"+name_func
+        mask = mask & active if active is not None else mask
+        n.model.add_constraints(status, "=", 0, name=name, mask=mask)
+
+    # linearized approximation because committable can partly start up and shut down
+    cost_equal = all(
+        n.df(c).loc[com_i, "start_up_cost"] == n.df(c).loc[com_i, "shut_down_cost"]
+    )
+    # only valid additional constraints if start up costs equal to shut down costs
+    if n._linearized_uc and not cost_equal:
+        logger.warning(
+            "The linear relaxation of the unit commitment cannot be "
+            "tightened since the start up costs are not equal to the "
+            "shut down costs. Proceed with the linear relaxation "
+            "without the tightening by additional constraints. "
+            "This might result in a longer solving time."
+        )
+    if n._linearized_uc and cost_equal:
+        # dispatch limit for partly start up/shut down for t-1
+        lhs = (
+            p.shift(snapshot=1)
+            - ramp_shut_down * status.shift(snapshot=1)
+            - (upper_p - ramp_shut_down) * (status - start_up)
+        )
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-mod-p-before"+name_func, mask=active)
+
+        # dispatch limit for partly start up/shut down for t
+        lhs = p - upper_p * status + (upper_p - ramp_start_up) * start_up
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-mod-p-current"+name_func, mask=active)
+
+        # ramp up if committable is only partly active and some capacity is starting up
+        lhs = (
+            p
+            - p.shift(snapshot=1)
+            - (lower_p + ramp_up_limit) * status
+            + lower_p * status.shift(snapshot=1)
+            + (lower_p + ramp_up_limit - ramp_start_up) * start_up
+        )
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(
+            lhs, "<=", 0, name=f"{c}-com-mod-partly-start-up"+name_func, mask=active
+        )
+
+        # ramp down if committable is only partly active and some capacity is shutting up
+        lhs = (
+            p.shift(snapshot=1)
+            - p
+            - ramp_shut_down * status.shift(snapshot=1)
+            + (ramp_shut_down - ramp_down_limit) * status
+            - (lower_p + ramp_down_limit - ramp_shut_down) * start_up
+        )
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(
+            lhs, "<=", 0, name=f"{c}-com-mod-partly-shut-down"+name_func, mask=active
+        )
+
+def define_operational_constraints_for_committables_non_modular_non_extendables(n, sns, c):
+    """
+    Sets power dispatch constraints for committable but non modular and non extendable devices for a given
     component and a given attribute. The linearized approximation of the unit
     commitment problem is inspired by Hua et al. (2017) DOI:
     10.1109/TPWRS.2017.2735026.
@@ -128,14 +522,18 @@ def define_operational_constraints_for_committables(n, sns, c):
         name of the network component
     """
     com_i = n.get_committable_i(c)
+    not_mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod==0)").index
+    fix_i = n.get_non_extendable_i(c)
+
+    com_i = com_i.intersection(not_mod_i).intersection(fix_i).rename(com_i.name)
 
     if com_i.empty:
         return
 
     # variables
-    status = n.model[f"{c}-status"]
-    start_up = n.model[f"{c}-start_up"]
-    shut_down = n.model[f"{c}-shut_down"]
+    status = n.model[f"{c}-status"].loc[sns, com_i]
+    start_up = n.model[f"{c}-start_up"].loc[sns, com_i]
+    shut_down = n.model[f"{c}-shut_down"].loc[sns, com_i]
     status_diff = status - status.shift(snapshot=1)
     p = reindex(n.model[f"{c}-p"], c, com_i)
     active = get_activity_mask(n, c, sns, com_i) if n._multi_invest else None
@@ -176,25 +574,25 @@ def define_operational_constraints_for_committables(n, sns, c):
 
     # lower dispatch level limit
     lhs = (1, p), (-lower_p, status)
-    n.model.add_constraints(lhs, ">=", 0, name=f"{c}-com-p-lower", mask=active)
+    n.model.add_constraints(lhs, ">=", 0, name=f"{c}-com-non-mod-fix-p-lower", mask=active)
 
     # upper dispatch level limit
     lhs = (1, p), (-upper_p, status)
-    n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-p-upper", mask=active)
+    n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-non-mod-fix-p-upper", mask=active)
 
     # state-transition constraint
     rhs = pd.DataFrame(0, sns, com_i)
     rhs.loc[sns[0], initially_up] = -1
     lhs = start_up - status_diff
     n.model.add_constraints(
-        lhs, ">=", rhs, name=f"{c}-com-transition-start-up", mask=active
+        lhs, ">=", rhs, name=f"{c}-com-non-mod-fix-transition-start-up", mask=active
     )
 
     rhs = pd.DataFrame(0, sns, com_i)
     rhs.loc[sns[0], initially_up] = 1
     lhs = shut_down + status_diff
     n.model.add_constraints(
-        lhs, ">=", rhs, name=f"{c}-com-transition-shut-down", mask=active
+        lhs, ">=", rhs, name=f"{c}-com-non-mod-fix-transition-shut-down", mask=active
     )
 
     # min up time
@@ -208,7 +606,7 @@ def define_operational_constraints_for_committables(n, sns, c):
         lhs = -status.loc[:, min_up_time_i] + merge(expr, dim=com_i.name)
         lhs = lhs.sel(snapshot=sns[1:])
         n.model.add_constraints(
-            lhs, "<=", 0, name=f"{c}-com-up-time", mask=mask[min_up_time_i]
+            lhs, "<=", 0, name=f"{c}-com-non-mod-fix-up-time", mask=mask[min_up_time_i]
         )
 
     # min down time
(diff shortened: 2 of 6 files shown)


pypsa/consistency.py
```diff
@@ -314,10 +323,15 @@ def check_assets(network: Network, component: Component) -> None:
     if component.name in {"Generator", "Link"}:
         committables = network.get_committable_i(component.name)
         extendables = network.get_extendable_i(component.name)
-        intersection = committables.intersection(extendables)
+        not_mod_i = (
+            network.df(component.name)
+            .query(f"({nominal_attrs[component.name]}_mod==0)")
```
davide-f, line 328: Since this part of the code applies only to links and generators, it is safe to specify p_nom instead of using the global dictionary.
The dictionary is not needed.
Moreover, instead of using == 0, I'd recommend to compare using "p_nom>0" and adapt the code.
The reason is that >0 is fail-prone to nan values and negative values, while ==0 is not.

pypsa/optimization/constraints.py
```diff
@@ -62,19 +62,64 @@ def define_operational_constraints_for_non_extendables(
...
+        dispatch_lower, ">=", lower, name=f"{c}-fix-non-comm-{attr}-lower", mask=active
     )
     n.model.add_constraints(
-        dispatch_upper, "<=", upper, name=f"{c}-fix-{attr}-upper", mask=active
+        dispatch_upper, "<=", upper, name=f"{c}-fix-non-comm-{attr}-upper", mask=active
     )
 
 
-def define_operational_constraints_for_extendables(
+def define_operational_constraints_for_extendables_but_non_committables(
```
davide-f, line 72: Why do we need all these cases? Can't we have some as special cases of more general formulations?

pypsa/optimization/constraints.py
```diff
@@ -62,19 +62,64 @@ def define_operational_constraints_for_non_extendables(
         dispatch_lower = (1, dispatch_lower), (-1, loss)
         dispatch_upper = (1, dispatch_upper), (1, loss)
     n.model.add_constraints(
-        dispatch_lower, ">=", lower, name=f"{c}-fix-{attr}-lower", mask=active
+        dispatch_lower, ">=", lower, name=f"{c}-fix-non-comm-{attr}-lower", mask=active
```
davide-f, line 65: Better not to change existing function names and constraints name for consistency with the past; I'd recommend to revert these changes, including the name of the function

pypsa/optimization/constraints.py
```diff
@@ -105,16 +154,391 @@ def define_operational_constraints_for_extendables(
...
     n.model.add_constraints(
-        lhs_lower, ">=", 0, name=f"{c}-ext-{attr}-lower", mask=active
+        lhs_lower, ">=", 0, name=f"{c}-ext-comm-non-mod-{attr}-lower", mask=active
+    )
+    n.model.add_constraints(
+        lhs_upper, "<=", 0, name=f"{c}-ext-comm-non-mod-{attr}-upper", mask=active
+    )
+
+
+def define_committability_variables_constraints_for_non_modular(n, sns, c, attr):
```
davide-f, line 164: I'd recommend to have a single function that takes care of setting the status variable for all components; there is quite a strong duplication among them.
The more general function is basically equivalent to the modular one, where the non-modular items get a 1.

pypsa/optimization/constraints.py
```diff
@@ -105,16 +154,391 @@ def define_operational_constraints_for_extendables(
...
+    if inter_i.empty:
+        return
+
+    m = n.model
+
+    active = get_activity_mask(n, c, sns, ext_i) if n._multi_invest else None
+
+    n_mod = n.model[f"{c}-n_mod"].loc[inter_i]
+
+    # ALLERT DIMENSIONI DISEQUAZIONE
```
davide-f, line 306: Can you explain these comments?

pypsa/optimization/constraints.py
```diff
@@ -105,16 +154,391 @@ def define_operational_constraints_for_extendables(
...
+        lhs, "<=", 0, name=f"{c}-shut_down-{attr}-upper-mod_ext", mask=active
+    )
+
+
+def define_operational_constraints_for_committables_and_modular(n, sns, c):
+    """
+    Sets power dispatch constraints for committable and modular devices for a given
+    component and a given attribute, whether they are extendable or not. The linearized
+    approximation of the unit commitment problem is inspired by Hua et al. (2017) DOI:
+    10.1109/TPWRS.2017.2735026.
```
davide-f, line 333: Revise

pypsa/optimization/constraints.py
```diff
@@ -105,16 +154,391 @@ def define_operational_constraints_for_extendables(
...
+            - (lower_p + ramp_down_limit - ramp_shut_down) * start_up
+        )
+        lhs = lhs.sel(snapshot=sns[1:])
+        n.model.add_constraints(
+            lhs, "<=", 0, name=f"{c}-com-mod-partly-shut-down" + name_func, mask=active
+        )
+
 
-def define_operational_constraints_for_committables(n, sns, c):
+def define_operational_constraints_for_committables_non_modular_non_extendables(
```
davide-f, line 537: As mentioned, avoid renaming functions and constraints unless strictly necessary.
I'd avoid that for the moment for backward compatibility

pypsa/optimization/constraints.py
```diff
@@ -415,40 +859,88 @@ def p_previous(idx):
             active.index, columns=fix_i
         )
         m.add_constraints(
-            lhs, ">=", rhs, name=f"{c}-fix-{attr}-ramp_limit_down", mask=mask
+            lhs, ">=", rhs, name=f"{c}-fix-non-comm-{attr}-ramp_limit_down", mask=mask
+        )
+
+    # ----------------------- Extendable but non committable Generators ----------------------------- #
```
davide-f, line 865: Why is this section needed?
For non-commitable generators I didn't expect much changes no? the previous constrains should be enough

pypsa/optimization/constraints.py
```diff
@@ -415,40 +859,88 @@ def p_previous(idx):
...
+            lhs, ">=", rhs, name=f"{c}-ext-non-comm-{attr}-ramp_limit_down", mask=mask
         )
 
-    # ----------------------------- Extendable Generators ----------------------------- #
+    # ----------------------- Extendable, committable but non modular Generators ----------------------------- #
+
+    ext_i = n.get_extendable_i(c)
+    com_i = n.get_committable_i(c)
+    not_mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod==0)").index
+    ext_i = ext_i.intersection(com_i).intersection(not_mod_i).rename(ext_i.name)
```
davide-f, line 904: Given the frequent use of these block, we may create a n.get_extendable_and_commitable function maybe?

pypsa/optimization/constraints.py
```diff
@@ -415,40 +859,88 @@ def p_previous(idx):
...
+            mask=mask,
         )
 
-    # ----------------------------- Committable Generators ----------------------------- #
+    # ---------------- Committable but non extendable and non modular Generators ----------------------------- #
+
+    com_i = n.get_committable_i(c)
+    fix_i = n.get_non_extendable_i(c)
+    not_mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod==0)").index
+    com_i = com_i.intersection(fix_i).intersection(not_mod_i).rename(com_i.name)
```
davide-f, line 943: No need to reeat this block so frequently; many lines are duplicated before; better to avoid duplication

davide-f: the same applies before and below

pypsa/optimization/optimize.py
```diff
@@ -22,16 +22,21 @@
 )
 from pypsa.optimization.common import get_strongly_meshed_buses, set_from_frame
 from pypsa.optimization.constraints import (
```
davide-f, line 24: I see a lot of new functions; we should reduce them and avoid changing the name to the existing ones

pypsa/optimization/optimize.py
```diff
@@ -46,14 +51,18 @@
     define_transmission_volume_expansion_limit,
 )
 from pypsa.optimization.variables import (
+    # Removed function no longer used
+    # define_start_up_variables,
+    # Removed function no longer used
+    # define_status_variables,
+    define_integer_committability_variables,
```
davide-f, line 58: I'd recommend to avoid changing too much the structure; better to keep the previous functions and adapt them.
What do you think?

test/test_compatibility_ext_and_com.py
```diff
@@ -0,0 +1,89 @@
+from numpy.testing import assert_array_almost_equal as equal
+
+import pypsa
+from pypsa.descriptors import nominal_attrs
+
+
+def test_compatibility_ext_and_comt():
```
davide-f, line 7: It would be good to add a case copy using extendable on and commitable with p_nom = p_nom_mod.
This means that the solution should be equivalent to the one of a standard unit commitment, without extendability being on.
As template, you could use one test of unit commitment.
What do you think?

davide-f (commented):
Thanks @enricogiglio :D
I've gone through the PR and added some preliminary comments.

Please, see that the pre-commit has added edits.
Do:
git fetch
git pull
to have them locally.

Afterwards, ensure to install the pre-commit locally using:
pre-commit install

This should prevent future online commits related to the precommit

pypsa/consistency.py
enricogiglio: I agree. Done :)

pypsa/optimization/optimize.py
enricogiglio: In my opinion we cannot. As you can see, we now have 5 functions, and:
1) Before PyPSA used 3 functions, which I've simply renamed, since the previous names were based on the idea that committability and extendability are not compatible and now it could lead to misunderstandings.
2) One of the 3 functions is used twice with different names to avoid misunderstandings when choosing the right idx (see attached file).
3) I have just added a single new feature that could potentially replace 2 out of 4 of the previous features, but we prefer not to do so in order to maintain backwards compatibility.

To explain the need for these 5 functions I have prepared the attached pdf file. I have also included the Google Drive link so that you can comment on the file if you prefer: https://docs.google.com/document/d/1CUDymdWgfhw56-dcihNnF0sKwZSMe1W9/edit?usp=sharing&ouid=112809165806691547710&rtpof=true&sd=true

[PR_comment.docx.pdf](https://github.com/user-attachments/files/16735809/PR_comment.docx.pdf)

pypsa/optimization/optimize.py
enricogiglio: It works for me. The idea behind "one single function" follows.

PyPSA now has 3 functions to create the variables status, startup and shutdown. Since these variables are always binary, we only have 3 functions that do the same thing by assigning different names to the variables (each function creates the variable and also sets the constraint <=1).

Going back to our case, for each of the 3 variables mentioned, we need
a) a function to create the integer variable
b) a function to limit the variable to <= n_mod variable (when referring to expandable components)
c) a function to limit the variable to <= P_nom/P_nom_mod (when referring to a non-extendable component).
d) a function to limit the variable to <= 1 (when modularity is not used - see actual situation).

It follows that in our case the 3 functions would become 12. To avoid this, and to limit the "increase of functions" (that you mention in the previous comment), I prefer to use only one function for each of the 4 actions a), b), c) ,d). 

However, having clarified the idea behind it, if you suggest keeping 9 different functions, I'll move in that direction.

test/test_compatibility_ext_and_com.py
enricogiglio: @cfvescovo and @LoZack19 told me that they would like to give a help with this PR, designing together a more appropriate test, since I would like to improve the current one, which is not so challenging.

I got the point, but obiusly keeping extendability the results won't be comparable with actual "unit commitment test". maybe we could add a case like the one you suggested directly in the network proposed as test. What do you think?

pypsa/optimization/constraints.py
enricogiglio: I think they comments for my self that I forgot to remove xD

pypsa/optimization/constraints.py
enricogiglio: What should I revise?

pypsa/optimization/constraints.py
enricogiglio: I know and I agree. As explained before, it seemed necessary to me, since these names are based on the idea that ext and comm are not compatible. However, tell me if you do not agree that I will change.

pypsa/optimization/constraints.py
enricogiglio: In previous comments I explained the need for 3 functions in the definition of constraints. But to summarise, from my point of view we have 3 different situations:
a) status variable <= n_mod, where n_mod is a variable and therefore belongs to lhs
b) status variable <= P_nom/P_nom_mod, where element is not extensible and P_nom/P_nom_mod belongs to rhs
c) status variable <= 1, for elements for which p_nom_mod is not declared, regardless of whether the element is extendable or not (since event for extendable elements, where p_nom_mod is not declared, but we have to use 1 as upper limit). 

Cases b) and c) can be combined into one, but looking at other functions it looks to me preferable to separate the cases. Of course, another way to reduce the functions would be to select the IDX to which the function is to be applied outside the function itself, but in actually used functions the selection of the cases is done in the function, so I kept this logic. What do you think?

pypsa/optimization/constraints.py
enricogiglio: [PR_comment.docx.pdf](https://github.com/user-attachments/files/16735958/PR_comment.docx.pdf)

pypsa/optimization/constraints.py
enricogiglio: [Uploading PR_comment.docx.pdf…]()

test/test_compatibility_ext_and_com.py
cfvescovo: > @cfvescovo and @LoZack19 told me that they would like to give a help with this PR, designing together a more appropriate test, since I would like to improve the current one, which is not so challenging.

Chiming in to confirm my availability.
W.r.t. the functions naming/duplication issue, I think a minimalist approach should be followed. Creating 12 functions to handle separately every single case should be avoided.

cfvescovo: Another point to consider: is backwards compatibility crucial for a library which hasn't reached API stability (1.0) yet?
IMHO cleaner code and a simpler API should be preferred over backwards compatibility.

Disclaimer: I am not a maintainer so take this with a grain of salt.

LoZack19: I agree with @cfvescovo about the need to create clean code, especially at this stage of pypsa development.

pypsa/optimization/constraints.py
enricogiglio: Absolutely. If you think it could be useful, I can do it for sure :)

pypsa/consistency.py
```diff
@@ -17,6 +17,15 @@
...
 
+nominal_attrs = {
+    "Generator": "p_nom",
+    "Line": "s_nom",
+    "Transformer": "s_nom",
+    "Link": "p_nom",
+    "Store": "e_nom",
+    "StorageUnit": "p_nom",
+}
+
```
davide-f, line 28: This is not needed

test/test_compatibility_ext_and_com.py
```diff
@@ -0,0 +1,89 @@
...
+        start_up_cost=2,
+    )
+
+    n.add("Load", "load", bus="bus", p_set=[4000, 6000, 5000, 800])
+
+    n.optimize()
+
+    f_obj = 0
+    for c in nominal_attrs.keys():
+        f_obj += (n.df(c)[f"{nominal_attrs[c]}_opt"] * n.df(c).capital_cost).sum()
```
davide-f, line 81: Please revise without the use of nominal_attrs

pypsa/optimization/constraints.py
```diff
@@ -98,24 +100,32 @@ def define_operational_constraints_for_extendables(
     attr : str
         name of the attribute, e.g. 'p'
     """
-    lhs_lower: DataArray | tuple
-    lhs_upper: DataArray | tuple
-
     ext_i = n.get_extendable_i(c)
+    com_i = n.get_committable_i(c)
+    mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod>0)").index
```
davide-f, line 105: Committables apply for link and generators: p_nom_mod

pypsa/optimization/constraints.py
```diff
@@ -98,24 +100,32 @@ def define_operational_constraints_for_extendables(
...
 
-    if ext_i.empty:
+    inter_i = ext_i.difference(n.get_committable_i(c)).rename(ext_i.name)
+
+    if inter_i.empty:
+        inter_i = ext_i.intersection(com_i).difference(mod_i).rename(ext_i.name)
+    else:
+        inter_i = inter_i.union(
+            ext_i.intersection(com_i).difference(mod_i).rename(ext_i.name)
+        )
```
davide-f, line 114: Probably a good idea to explain this.
We may find a compact way to do this, maybe with a smart query search; but let's keep this for later

pypsa/optimization/constraints.py
```diff
@@ -127,13 +137,166 @@ def define_operational_constraints_for_extendables(
...
+        rhs.loc[sns, inter_i] = n_mod.loc[inter_i].values
+
+    ##############################################################
+    # rhs is completed with element "1" for combinable components, but not modular. rhs = 1
+
+    com_i = n.get_committable_i(c)
+    mod_i = n.df(c).query(f"({attr}_mod>0)").index
+    com_i = com_i.difference(mod_i).rename(com_i.name)
+
+    if not com_i.empty:
```
davide-f, line 203: This set of if cases is not clear and not tested

pypsa/optimization/constraints.py
```diff
@@ -192,57 +374,69 @@ def define_operational_constraints_for_committables(
...
-    lhs_tuple = (1, p), (-lower_p, status)
-    n.model.add_constraints(lhs_tuple, ">=", 0, name=f"{c}-com-p-lower", mask=active)
+    lhs = (1, p), (-lower_p, status)
+    n.model.add_constraints(lhs, ">=", 0, name=f"{c}-com-p-lower", mask=active)
 
     # upper dispatch level limit
-    lhs_tuple = (1, p), (-upper_p, status)
-    n.model.add_constraints(lhs_tuple, "<=", 0, name=f"{c}-com-p-upper", mask=active)
+    lhs = (1, p), (-upper_p, status)
+    n.model.add_constraints(lhs, "<=", 0, name=f"{c}-com-p-upper", mask=active)
```
davide-f, line 382: please revert

pypsa/optimization/constraints.py
```diff
@@ -437,11 +634,20 @@ def p_previous(idx: pd.Index) -> DataArray:
...
     # ----------------------------- Extendable Generators ----------------------------- #
 
+    ext_i = n.get_extendable_i(c)
+    com_i = n.get_committable_i(c)
+    mod_i = n.df(c).query(f"({nominal_attrs[c]}_mod>0)").index
+    inter_i = ext_i.difference(com_i).rename(ext_i.name)
+    inter_i = inter_i.union(
+        ext_i.intersection(com_i).difference(mod_i).rename(ext_i.name)
+    )
+    ext_i = inter_i
```
davide-f, line 644: these redefinitions are confusing, we should define everything at the beginning as it was done previously.
It is likely that we need a get_{something} to improve the clarity

pypsa/optimization/constraints.py
enricogiglio: Yes, I agree, and initially I repeated the same function twice to avoid things like this. I agree that repeating the same function twice is not a good option, but I will find a way to handle this.

test/test_compatibility_ext_and_com.py
enricogiglio: done

pypsa/consistency.py
enricogiglio: fixed

pypsa/optimization/constraints.py
enricogiglio: done :)

enricogiglio:
@davide-f I integrated the changes you suggested as discussed, but 1) we also added more tests and 2) now we have passed all the checks.  Any other suggestions?

davide-f:
> @davide-f I integrated the changes you suggested as discussed, but 1) we also added more tests and 2) now we have passed all the checks.  Any other suggestions?

To me the PR looks ready to be revised by the maintainers.
You can convert it to ready

cfvescovo (approved):

davide-f:
Hello :)
Just posting here to align on the expected steps on this to facilitate the review, if deemed useful.

We believed this feature is interesting for upstream, as it is currently supported by PLEXOS and numerical results allow quite large time reduction with respect to enumeration of generators.

lkstrp:
You are right @davide-f , thanks for the ping ! We'll get to this one soon.

davide-f:
> Hi @davide-f and @enricogiglio,
> 
> okay, we've finally had a closer look, thanks for the contribution! Extendability and committability at the same time would be a good feature and the new formulation looks great.
> 
> The main problem we see at the moment is that it's quite a lot of code, which makes the optimisation module (even) more complicated for a rather small feature.
> 
> We are currently playing around with a general overhaul of the optimisation module, with a different data structure in the background (xarray instead of pandas), which should hopefully simplify things massively. That's why we want to postpone this PR for even a bit longer, so we can work on both at the same time. But we are getting there !

I've incorporated the latest changes, which probably were the main concern last December.
Take into consideration that about 500 lines is the length of the additional test case to test this feature and combinations of modularity yes/no, commitment yes/no, extendability yes/no.
Additional changes are needed to incorporate the above combinations and multiple architecture are possible.
We can also align on doubts; probably good to have some preliminary comments first though

davide-f:
Hello @lkstrp :)
As v1.0 is out, I can come back here when time is appropriate as discussed.

To clarify the difference of features:
1. #1367 aims to allow the UC in the linearized form p_nom * status * p_min_pu <= p <= p_nom * status * p_max_pu [this is expressed in a non-linear form, but essentially the two formulations are equivalent.
2. this PR aims to allow the UC combined with modularity in the form p_nom_mod * status * p_min_pu <= p <= p_nom_mod * status * p_max_pu, where status becomes an integer variable identifying the number of units being dispatched

The difference is: let's suppose we have a generator of total capacity 4GW (4 units of 1GW each, e.g. nuclear plant) p_min_pu=0.6, composed. With 1, the effective constraint is p_nom * status * 0.6 <= ...., where p_nom is the total capacity. Therefore, the solver shall decide between either shut the unit down or dispatching 4*0.6=2.4GW.
With 2, it is actually possible to dispatch only a single unit of 1GW with 60% minimum capacity.
Obviously the above extends to costs of dispatching ramps etc.

Both formulations are potentially interesting as they answer different problems.

Anyway, this PR shall be adapted to the new format and changes are expected.
To work effectively, I see this PR had fewer changes than #1367 ; to reduce further, the only view I see is to merge all committable options in one function only, which reduces the lines changed due to documentation.
In the preliminary version that was not done to keep functions separate, but they can be merged.
Is this what you were looking for?

Do you envision to merge #1367 first and this can build on top? shall we start with an approach?

Many thanks!

FabianHofmann:
@davide-f many thanks for you comment and clarification! I really like the approach of allowing a generator representing a fleet of modular units, a powerful feature. 

After discussing with @lkstrp we are going to 

* pull #1367 into here and consolidate the functionality make the difference clear in the doc. this will already make the big bulk of work which is aligning with the new optimization conventions 
* after everything is resolved merge both into the master

If no obligations, I would proceed and prioritize this so we can merge it in asap

davide-f:
> @davide-f many thanks for you comment and clarification! I really like the approach of allowing a generator representing a fleet of modular units, a powerful feature. 
> 
> After discussing with @lkstrp we are going to 
> 
> * pull #1367 into here and consolidate the functionality make the difference clear in the doc. this will already make the big bulk of work which is aligning with the new optimization conventions 
> * after everything is resolved merge both into the master
> 
> If no obligations, I would proceed and prioritize this so we can merge it in asap 


Sounds great! Let me know how me and Enrico can be of help, e.g. predigest the integration if deemed useful (FYI I have write rights on the branch)

FabianHofmann:
@davide-f could you have a look at the pr to update this one? https://github.com/enricogiglio/PyPSA/pull/2

davide-f:
@lkstrp I've been exploring the diff of this PR with respect to https://github.com/PyPSA/PyPSA/pull/1367:
https://github.com/PyPSA/PyPSA/compare/committable-extendable...enricogiglio:PyPSA:AdvancedModularity

there are some unexpected differences on documentation and pre-commit. Just sharing to be sure the differences were intended

As mentioned, thanks for the patience of last week: I came back to office yesterday

lkstrp:
Hey @davide-f,

the diff is clean now. The branch just needed to be updated

davide-f:
Many thanks @lkstrp !
Way better for reviewing :) On the architecture point, you wished to have fewer code. To further reach that, we could think of merging the functions define_committability_variables_constraints_with_fixed_upper_limit and define_committability_variables_constraints_with_variable_upper_limit .

That should simplify the code and make it cleaner. That's the only major alignment needed.
On the reviewing side, it looks like our test passes, however the linearized UC is failing and cannot see why yet.

test/test_compatibility_ext_and_com.py
```diff
@@ -209,11 +217,11 @@ def test_unit_commitment_mod():
 
     expected_status = np.array([[1, 1, 1, 0], [0, 0, 0, 1]], dtype=float).T
 
-    equal(n.generators_t.status.values, expected_status)
+    equal(n.generators.dynamic.status.values, expected_status)
 
     expected_dispatch = np.array([[4000, 6000, 5000, 0], [0, 0, 0, 800]], dtype=float).T
 
-    equal(n.generators_t.p.values, expected_dispatch)
+    equal(n.generators.dynamic.p.values, expected_dispatch)
```
lkstrp: I will properly review, but just a quick note: This needs to be `n.c.generators.dynamic` to support both APIs

FabianHofmann: yeah, forgot about that again

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,223 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
 These constraints are set in the function `define_modular_constraints()`.
 
 
+## Combined Formulations
+
```
davide-f: I'd suggest to revise the introductory section to clarify the difference between the two formulations.
To my understanding, the difference is related to the option `p_nom_mod`: when null, the Big-M applies, otherwise the modular one.

```suggestion
## Combined Formulations for Unit Commitment and Capacity Expasion

When components are both **committable** (`committable=True`) and **extendable** (`p_nom_extendable=True`), the optimizer co-optimizes both capacity expansion and operational unit commitment decisions. The following two formulations apply depending on whether components are modular (`p_nom_mod > 0`) or not (default option).

- non-modular components `p_nom_mod=0` (default): Big-M formulation applies
- modular components `p_nom_mod>0`: Modular unit commitment applies with integer unit commitment

The detailed mathematical procedure is detailed in the following.
```

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,223 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
+### Committable and Extendable Components
+
+When components are both **committable** (`committable=True`) and **extendable** (`p_nom_extendable=True`), the optimizer co-optimizes both capacity expansion and operational unit commitment decisions. This enables modeling scenarios where both the optimal capacity to build AND the on/off operational schedule must be determined simultaneously.
```
davide-f: This introduction shall be slightly revised to adapt to the introduction. Below you find a draft, feel free to revise.


```suggestion
### Big-M formulation for Non-Modular, Committable and Extendable Components

When components are **non-modular** (`p_nom_mod=0`), **committable** (`committable=True`) and **extendable** (`p_nom_extendable=True`), the optimizer co-optimizes both capacity expansion and operational binary unit commitment decisions. This enables modeling scenarios where both the optimal capacity to build AND the on/off operational schedule must be determined simultaneously.
```

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,223 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
+        | $u_{l,t}$         | `n.links_t.status` | Decision variable |
+        | $M$               | auto-inferred or `pypsa.options.params.optimize.committable_big_m` | Parameter |
+        | $\underline{f}_{l,t}$ | `n.links_t.p_min_pu` | Parameter |
+        | $ru_{l,t}$        | `n.links.ramp_limit_up` | Parameter |
+        | $rd_{l,t}$        | `n.links.ramp_limit_down` | Parameter |
+
+
+### Modular and Committable Components
+
+When components have both **modularity** (`p_nom_mod > 0`) and **unit commitment** (`committable=True`), the formulation differs from the standard binary unit commitment. Instead of a binary on/off status variable, the status variable becomes an **integer** representing the **number of committed modules**.
```
davide-f:
```suggestion
### Modular and Committable Components

When components have both **modularity** (`p_nom_mod > 0`) and **unit commitment** (`committable = True`), the formulation differs from the standard binary unit commitment. Instead of a binary on/off status variable, the status variable becomes an **integer** representing the **number of committed modules**. This formulation is compatible for **expandable** (`p_nom_extendable = True`) and **non-expandable** (`p_nom_extendable = Fakse`) components.
```

pypsa/components/components.py
```diff
@@ -836,6 +836,33 @@ def committables(self) -> pd.Index:
...
         return idx
 
+    @property
+    def modulars(self) -> pd.Index:
+        """Get the index of modular elements of this component.
+
+        Modular components have a positive module size (e.g., p_nom_mod > 0)
+        which introduces integer variables for capacity expansion.
+
+        <!-- md:badge-version v1.0.1 -->
```
lkstrp:
```suggestion
        <!-- md:badge-version v1.1.0 -->
```

lkstrp: Also add to other new introduced methods

pypsa/components/descriptors.py
```diff
@@ -164,7 +165,12 @@ def active_assets(self) -> pd.Series:
 
         """
         active_assets = self.get_active_assets()
-        return active_assets[active_assets].index.get_level_values("name").unique()
+        active = active_assets[active_assets]
```
lkstrp: active state currently doesn't differ between scenarios

pypsa/data/component_attrs/generators.csv
```diff
@@ -24,7 +24,7 @@ build_year,int,year,0,Build year of the generator.,Input (optional)
 lifetime,float,years,inf,Lifetime of the generator.,Input (optional)
 capital_cost,float,currency/MW,0,"Fixed period costs of extending `p_nom` by 1 MW, including periodized investment costs and periodic fixed O&M costs (e.g. annuitized investment costs).",Input (optional)
 efficiency,static or series,per unit,1,"Ratio output and primary energy carrier input (e.g. 0.4 MWh~elec~/MWh~fuel~). This is required for global constraints on primary energy in optimization.",Input (optional)
-committable,boolean,n/a,False,Apply unit commitment constraints. This is only possible with `p_nom_extendable=False`.,Input (optional)
+committable,boolean,n/a,False,Apply unit commitment constraints. Can be combined with `p_nom_extendable=True`.,Input (optional)
```
lkstrp: Keep in line with docs for links.csv, so add Big M

pypsa/optimization/constraints.py
```diff
@@ -246,18 +280,21 @@ def define_operational_constraints_for_committables(
 
     """
     c = as_components(n, component)
-    com_i = c.committables.difference(c.inactive_assets)
+    com_i: pd.Index = c.committables.difference(c.inactive_assets)
```
lkstrp: All those extra types seem unnecessary

pypsa/optimization/constraints.py
lkstrp: Why can't we use the the `c.da` accessor here? Would also allow to remove the `DataArray(..)` cast below. We should get rid of all of them

pypsa/optimization/constraints.py
```diff
@@ -286,29 +324,162 @@ def define_operational_constraints_for_committables(
...
+
+        p_nom_max_vals = c.da.p_nom_max.sel(name=com_ext_i)
+        max_pu_vals = max_pu.sel(name=com_ext_i).max("snapshot")
+
+        big_m_default = options.params.optimize.committable_big_m
+        if big_m_default is None:
+            big_m_default = _infer_big_m_scale(n, component)
+        else:
+            if not np.isfinite(big_m_default):
+                msg = f"options.params.optimize.committable_big_m must be finite, got {big_m_default}."
```
lkstrp:
```suggestion
                msg = f"pypsa.options.params.optimize.committable_big_m must be finite, got {big_m_default}."
```

pypsa/optimization/constraints.py
```diff
@@ -286,29 +324,162 @@ def define_operational_constraints_for_committables(
...
+                raise ValueError(msg)
+            if big_m_default <= 0:
+                msg = f"options.params.optimize.committable_big_m must be positive, got {big_m_default}."
```
lkstrp:
```suggestion
                msg = f"pypsa.options.params.optimize.committable_big_m must be positive, got {big_m_default}."
```

pypsa/optimization/optimize.py
```diff
@@ -922,6 +942,88 @@ def assign_duals(self, assign_all_duals: bool = False) -> None:
                 ", ".join(unassigned_constraints),
             )
 
+    def _warn_big_m_exceeded(self) -> None:
```
lkstrp: Can we move the two big_m specific functions in another module? They are quite specific and optional

pypsa/optimization/optimize.py
```diff
@@ -922,6 +942,88 @@ def assign_duals(self, assign_all_duals: bool = False) -> None:
...
+
+        def _format_index(label: Any) -> str:
+            if isinstance(label, tuple):
+                return "/".join(str(part) for part in label)
+            return str(label)
+
+        n = self._n
+
+        for c_name in ["Generator", "Link"]:
+            c = as_components(n, c_name)
```
lkstrp: Just iterate across network components directly

pypsa/optimization/optimize.py
```diff
@@ -1010,6 +1114,39 @@ def v_ang_for_(sub: SubNetwork) -> pd.DataFrame:
                     [v_ang_for_(sub) for sub in n.c.sub_networks.static.obj], axis=1
                 ).reindex(columns=n.c.buses.static.index, fill_value=0.0)
 
+    def _validate_no_modular_committables(self) -> None:
```
lkstrp: I wanna move validation/ warnings/ user feedback away from optimize. Why not integrate within consistency checks? It's a bit different, but there is tons of strict user feedback coming in and we don't wanna have it all spread out everywhere

pypsa/optimization/optimize.py
```diff
@@ -1010,6 +1114,39 @@ def v_ang_for_(sub: SubNetwork) -> pd.DataFrame:
...
+            if len(modular_committables) > 5:
+                components_str += f", ... ({len(modular_committables)} total)"
+            msg = (
+                f"linearized_unit_commitment=True cannot be used with modular "
+                f"committable components: {components_str}. "
+                f"Modular components use integer status variables representing the "
+                f"number of committed modules, which cannot be meaningfully relaxed "
+                f"to continuous values. Use standard unit commitment "
+                f"(linearized_unit_commitment=False) or remove modular sizing "
+                f"(set p_nom_mod=0)."
```
lkstrp: Can we also link to the docs here?

pypsa/_options.py
```diff
@@ -471,6 +471,13 @@ def option_context(*args: Any) -> Generator[None, None, None]:
     "conditioning. None defaults to True with a FutureWarning (changes to False in v2.0).",
 )
 
+options._add_option(
```
lkstrp: This is not a params option. All params options are just changing defaults of parameters you can pass to methods. big_m can't me passed, so this should be moved. I am not even sure if this is how we should pass it at all. Why a fallback in the options? This will change model results

pypsa/consistency.py
```diff
@@ -447,12 +447,20 @@ def check_assets(n: NetworkType, component: Components, strict: bool = False) ->
...
+            logger.info(
+                "Components in %s are both committable and extendable: %s. "
+                "This is supported using a big-M formulation for mixed-integer optimization.",
+                component.name,
+                ", ".join(intersection),
+            )
+        intersection = committables.intersection(extendables).intersection(modulars)
+        if not intersection.empty:
+            logger.info(
+                "Components in %s are both committable, extendable, and modular: %s. "
```
lkstrp: What is this telling me

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,223 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
 These constraints are set in the function `define_modular_constraints()`.
 
 
+## Combined Formulations
+
+### Committable and Extendable Components
```
davide-f: The classification "committable and expandable" versus "modular and committable" can be interpreted as the latter not being compatible with expandability.
To minimize changes, "Modular Expansion with Unit Commitment" (as in the notebook) is better for the second section unless you see otherwise.

FabianHofmann:
wonderful, thanks for the comments @lkstrp  and @davide-f. will tackle them tomorrow!

pypsa/consistency.py
FabianHofmann: that what is says :) yes, it is a info. but I see, removing it

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Committable + Extendable Components\n",
+    "\n",
+    "This tutorial demonstrates how to create components (Generators and Links) that are both **committable** and **extendable** simultaneously. This enables co-optimizing capacity expansion with unit commitment decisions—determining both the optimal capacity to build AND when to turn units on or off.\n",
+    "\n",
+    "PyPSA uses a **big-M formulation** to linearize the nonlinear product of binary status variables and continuous capacity variables, maintaining the Mixed-Integer Linear Programming (MILP) structure.\n",
+    "\n",
+    "/// note | Mathematical Formulation\n",
```
lkstrp: `///` is not a thing. I guess it meant `!!!`

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+    "\n",
+    "For the complete mathematical formulation, constraint names, and detailed explanation of the big-M linearization, see [Capacity Limits: Committable and Extendable Components](https://docs.pypsa.org/latest/user-guide/optimization/capacity-limits/#committable-and-extendable-components).\n",
```
lkstrp, line 35: The link heading doesn't resolve

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+    "\n",
+    "///\n",
+    "\n",
+    "For discrete capacity expansion with unit commitment (where capacity is built in fixed blocks), see the [Modular Committable Components](modular-committable.ipynb) tutorial."
```
lkstrp: Link doesn't resolve, since this is in a jupyter notebook, you can't use the quick ref plugin. Our own shortcode implementations (the `<!-- pd ` stuff) should still work, but os probably the easiest to just use full links as above.

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Big-M Configuration\n",
+    "\n",
+    "The big-M formulation uses a large constant $M$ to linearize constraints. PyPSA automatically infers an appropriate value based on the network's peak load ($M = 10 \\times$ peak load), but you can override it manually if needed.\n",
+    "\n",
+    "/// note | Big-M Theory\n",
```
lkstrp: See above

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+    "## Summary\n",
+    "\n",
+    "This tutorial demonstrated committable + extendable components in PyPSA:\n",
+    "\n",
+    "1. **Basic Usage**: Set both `committable=True` and `p_nom_extendable=True` on a Generator or Link\n",
+    "2. **Mixed Portfolios**: Combine different generator types with various attribute combinations\n",
+    "3. **Ramp Limits**: Fully compatible with ramp rate constraints\n",
+    "4. **Configuration**: Automatic big-M inference with manual override option\n",
+    "\n",
+    "/// note | Learn More\n",
```
lkstrp: Another ///

docs/examples/committable-extendable.ipynb
```diff
@@ -0,0 +1,381 @@
...
+    "\n",
+    "- [Capacity Limits: Committable and Extendable Components](https://docs.pypsa.org/latest/user-guide/optimization/capacity-limits/#committable-and-extendable-components) - Complete mathematical formulation\n",
```
lkstrp: First links heading does not resolve, second link is 404, same as above

docs/examples/modular-committable.ipynb
```diff
@@ -0,0 +1,480 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Modular Expansion with Unit Commitment\n",
+    "\n",
+    "This tutorial demonstrates **modular expansion** combined with **unit commitment** in PyPSA. When both `p_nom_mod > 0` and `committable=True` are set, the status variable represents the **number of committed modules** (integer) rather than a simple binary on/off.\n",
+    "\n",
+    "This formulation is ideal for modeling technologies that come in standardized sizes: modular gas turbines, nuclear reactors, or HVDC interconnectors. The optimizer co-optimizes both how many modules to build AND how many to operate at each time step.\n",
+    "\n",
+    "/// note | Mathematical Formulation\n",
```
lkstrp: Another ///

docs/examples/modular-committable.ipynb
```diff
@@ -0,0 +1,480 @@
...
+    "\n",
+    "For the complete mathematical formulation, constraint names, and detailed explanation of the module-level commitment formulation, see [Capacity Limits: Modular and Committable Components](https://docs.pypsa.org/latest/user-guide/optimization/capacity-limits/#modular-and-committable-components).\n",
```
lkstrp, line 35: Link heading does not resolve

docs/examples/modular-committable.ipynb
```diff
@@ -0,0 +1,480 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Verify the capacity is indeed a multiple of the module size\n",
+    "assert abs(n_modules - round(n_modules)) < 1e-6, (\n",
+    "    \"Capacity should be a multiple of module size!\"\n",
```
lkstrp, line 130: Is this another potential consistency check? Doesn't need to be added, just to have it on the list

docs/examples/modular-committable.ipynb
```diff
@@ -0,0 +1,480 @@
...
+    "\n",
+    "This tutorial demonstrated modular expansion with unit commitment in PyPSA:\n",
+    "\n",
+    "1. **Modular Capacity**: Set `p_nom_mod` to force capacity to be built in discrete blocks\n",
+    "2. **Module-Level Commitment**: When combined with `committable=True`, status represents the number of committed modules (integer variable)\n",
+    "3. **Generators and Links**: The feature works for both component types\n",
+    "4. **Start-up/Shut-down Costs**: Apply when the number of committed modules changes\n",
+    "5. **Minimum Load**: `p_min_pu` is enforced per committed module\n",
+    "\n",
+    "/// note | Learn More\n",
```
lkstrp: See above

docs/examples/modular-committable.ipynb
```diff
@@ -0,0 +1,480 @@
...
+    "\n",
+    "- [Capacity Limits: Modular and Committable Components](https://docs.pypsa.org/latest/user-guide/optimization/capacity-limits/#modular-and-committable-components) - Complete mathematical formulation\n",
```
lkstrp, line 452: Both links dont resolve, as above

docs/examples/committable-extendable.ipynb
lkstrp: Nevermind, that one makes sense because the heading doesn't exist yet. Would be nice to get all the mkdocs linking features in notebooks as well. 

But the other link still doesn't resolve

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,225 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
 
 
+## Committable and Extendable Components
+
+When components are both **committable** (`committable=True`) and **extendable** (e.g. `p_nom_extendable=True`), the optimizer co-optimizes both capacity expansion and operational unit commitment. The following two formulations apply depending on whether components are modular (`p_nom_mod > 0`) or not (default option).
+
+- non-modular components `p_nom_mod=0` (default): Big-M formulation applies
+- modular components `p_nom_mod>0`: Modular unit commitment applies with integer unit commitment
+
+The detailed mathematical procedure is described below.
```
lkstrp:
```suggestion
```

A bit verbose compared to the rest of our lovely user guide. Could just be integrated in the following headings. Also a little intro sentence to extendability would be nice, isnt it?

lkstrp: Otherwise wonderful docs!

docs/user-guide/optimization/unit-commitment.md
```diff
@@ -35,9 +35,14 @@ i.e. $u_{*,t} = 1$:
     | $f_{l,t} \geq u_{l,t} \cdot \underline{f}_{l,t} \cdot \hat{f}_{l}$ | `Link-com-p-lower` |
     | $f_{l,t} \leq u_{l,t} \cdot \bar{f}_{l,t} \cdot \hat{f}_{l}$ | `Link-com-p-upper` |
 
-!!! warning
+/// note | Combined with Capacity Expansion
```
lkstrp: Another ///

pypsa/optimization/constraints.py
```diff
@@ -367,27 +533,25 @@ def define_operational_constraints_for_committables(
             "<=",
             1,
             name=f"{c.name}-com-down-time",
-            mask=DataArray(active[min_down_time_i]).sel(snapshot=sns[1:]),
+            mask=active.loc[sns[1:], min_down_time_i],
         )
     # up time before
-    timesteps = pd.DataFrame([range(1, len(sns) + 1)] * len(com_i), com_i, sns).T
+    timesteps = xr.DataArray(
```
lkstrp, line 482: Can we not get the same from `c.da`? Better than the `.T` pandas version, but still looks like it could annoy us in future

pypsa/optimization/optimize.py
```diff
@@ -575,6 +589,11 @@ def create_model(
         sns = as_index(n, snapshots, "snapshots")
         n._linearized_uc = int(linearized_unit_commitment)
         n._multi_invest = int(multi_investment_periods)
+        n._committable_big_m = committable_big_m
```
lkstrp, line 619: A discussed: All those internal attributes should be better handled and fully stored for reproducibility.

Also below the 

```
        if linearized_unit_commitment:
            check_no_modular_committables(n)
```

is very inconsistent with how we treat consistency checks. `linearized_unit_commitment` should be passed through the general consistency checks and treated there. Preferably with a better system on how we store the attributes above. And then `consistency_checks` should be split up in: (1) strict checks, which can not be turned off and always run and (2) warnings, which you could turn off in `n.optimize`. But this is all internal, so I would rather move that to another PR. It get's also more wild since `check_big_m_exceeded` is a check which just runs in post processing and can't be merged, and we also have all the plotting checks with the `n.sanitize` function

pypsa/consistency.py
```diff
@@ -886,8 +887,6 @@ def consistency_check(
             check_for_zero_impedances(self, c, "zero_impedances" in strict)
             # Checks transformers
             check_for_zero_s_nom(c, "zero_s_nom" in strict)
-            # Checks generators and links
-            check_assets(self, c, "assets" in strict)
```
lkstrp, line 902: Why was this removed?

pypsa/networks.py
```diff
@@ -163,6 +164,9 @@ def __init__(
         self._model: linopy.Model | None = None
         self._objective: float | None = None
         self._objective_constant: float | None = None
+        self._linearized_uc: int = 0
```
lkstrp: This is just claude who likes static attributes?

`._model`, `_objective`, and `_objective_constant` are internals for their public attribute and therefore read only (`.model`, `objective`, `objective_constant`). 

We should get rid of the rest, which is dynamically assigned so far without a public equivalent, instead of adding them here

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,225 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
+
+The challenge when combining capacity expansion with unit commitment is that the upper bound on dispatch becomes nonlinear: $g_{n,s,t} \leq u_{n,s,t} \cdot G_{n,s}$ (the product of binary status $u$ and continuous capacity $G$). To maintain a Mixed-Integer Linear Programme (MILP), PyPSA uses a **big-M formulation** that replaces this nonlinear constraint with two linear constraints:
+
+=== "Generator"
+
+    | Constraint | Name |
+    |-------------------|------------------|
+    | $g_{n,s,t} \leq u_{n,s,t} \cdot M$ | `Generator-com-ext-p-upper-bigM` |
+    | $g_{n,s,t} \leq G_{n,s}$ | `Generator-com-ext-p-upper-cap` |
+    | $g_{n,s,t} \geq u_{n,s,t} \cdot \underline{g}_{n,s,t} \cdot M$ | `Generator-com-ext-p-lower` |
```
Irieo: It says here $g_{n,s,t} \geq u_{n,s,t} \cdot \underline{g}_{n,s,t} \cdot M$ 

Suppose  u=1 and min_pu=0.3 and M=10000. This yields p >= 1 * 0.3 * 10000
This means the constraint forces dispatch to be >= 3000 MW regardless of capacity. It doesn't feel right.

If I lookup the implementation in constraints.py, it goes as
```
        lhs_lower_ext = (1, p_ext), (-min_pu_ext, p_nom_ext), (-M_values, status_ext)
        n.model.add_constraints(
            lhs_lower_ext,
            ">=",
            -M_values,
            name=f"{c.name}-com-ext-p-lower",
            mask=active_ext,
        )
```
Which means in a readable format: p - min_pu * p_nom - M * u >= -M
and can be rearranged to: p >= min_pu * p_nom + M * (u - 1)
(or to p >= p_min_pu * p_nom - M * (1 - u) that expands to identical constraint)

I suppose: implementation is correct, docs are wrong

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,225 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
+### Non-Modular Components (default)
+
+The challenge when combining capacity expansion with unit commitment is that the upper bound on dispatch becomes nonlinear: $g_{n,s,t} \leq u_{n,s,t} \cdot G_{n,s}$ (the product of binary status $u$ and continuous capacity $G$). To maintain a Mixed-Integer Linear Programme (MILP), PyPSA uses a **big-M formulation** that replaces this nonlinear constraint with two linear constraints:
+
+=== "Generator"
+
+    | Constraint | Name |
+    |-------------------|------------------|
+    | $g_{n,s,t} \leq u_{n,s,t} \cdot M$ | `Generator-com-ext-p-upper-bigM` |
+    | $g_{n,s,t} \leq G_{n,s}$ | `Generator-com-ext-p-upper-cap` |
```
Irieo: It is minor comment but implementation actually says 
`lhs_upper_cap = (1, p_ext), (-max_pu_ext, p_nom_ext)`

(where `max_pu_ext = max_pu.sel(name=com_ext_i)`)
although max_pu default is 1, I'd still spell it out explicitly in the math docs

pypsa/optimization/constraints.py
```diff
@@ -281,34 +318,163 @@ def define_operational_constraints_for_committables(
...
+            fallback_values,
+        )
+        p_ext = p.sel(name=com_ext_i)
+        status_ext = status.sel(name=com_ext_i)
+        p_nom_ext = p_nom_var.sel(name=com_ext_i)
+        min_pu_ext = min_pu.sel(name=com_ext_i)
+        max_pu_ext = max_pu.sel(name=com_ext_i)
+
+        active_ext = active.sel(name=com_ext_i)
+        lhs_lower_ext = (1, p_ext), (-min_pu_ext, p_nom_ext), (-M_values, status_ext)
```
Irieo, line 309: It is so much easier to follow linopy code in plain vector form, and it can be directly compared to math in docs. I'd suggest using that instead of linear expression in tuples unless there's a reason for a latter.

pypsa/optimization/constraints.py
```diff
@@ -732,6 +896,12 @@ def p_previous(idx: pd.Index) -> DataArray:
                 mask=mask,
             )
     # ----------------------------- Committable Components ----------------------------- #
+    # For committable+extendable (non-modular) components, skip committable ramp constraints.
```
Irieo: Are we sure this is in docs?

docs/user-guide/optimization/capacity-limits.md
```diff
@@ -41,6 +41,225 @@ If `{p,s,e}_nom_mod>0`, the nominal capacity is given by:
...
+|---------|----------------------------------|----------------------|
+| Start-up costs | ✓ | ✓ |
+| Shut-down costs | ✓ | ✓ |
+| Minimum part-load (`p_min_pu`) | ✓ | ✓ (per module)¹ |
+| Ramp limits (`ramp_limit_up/down`) | ✓ | ✓² |
+| Stand-by costs | ✓ | ✓ |
+| Minimum up time (`min_up_time`) | ✓ | ✗ |
+| Minimum down time (`min_down_time`) | ✓ | ✗ |
+| Up time before (`up_time_before`) | ✓ | ✗ |
+| Down time before (`down_time_before`) | ✓ | ✗ |
```
Irieo: This is fantastic way to present the summary, really neat! I think it should be expanded with few things though.

1. It mentions `ramp_limit_up/down` but not `ramp_limit_start_up/down` (NB ramp_limit_start_up/down don't work here at all, I expand more about it in summary review comment)

2. can also mention what happens for linearized UC formulation. I tested that for Committable + Extendable (Big-M) works fine (very prompt test by re-running example notebook case with linearized flag), and Modular + Committable is captured via consistency check.

3. What about multi-period optimization? I also don't see any tests for multi-period for either left or right case. One should always be cautious with the how the status variables (status, u, start_up, shut_down, ...) transition between periods

docs/user-guide/optimization/capacity-limits.md
FabianHofmann: I have restructured the document now to improve the readability. perhaps you can have a short look later

pypsa/consistency.py
FabianHofmann: good question, reinserted

pypsa/networks.py
FabianHofmann: let's keep it consistent now and remove them all in a different pr?

pypsa/optimization/constraints.py
FabianHofmann: you are right, however the tuple list expressions are still everywhere and somewhat more efficient, fine if we keep that for another pr?

Irieo: of course

davide-f:
@Irieo Adding a comment on

> The reported status and dispatch variables are 5 & 500 MW, but there are 4 startups (since first one is already started up), which also means we pay in objective for 4 UC events and the first one PyPSA gives for free 🎁. User can control it by adjusting default status obv, and I don't think this is something worth fixing, but it is worth documenting.

That phenomenon is intended, though agree worth documenting. With modularity options, a single PyPSA object simulates the dispatch of N independent units, which saves lots of variables (continuous and binary ones). As such, if any of the unit is dispatched, then the startup cost is paid.
Savings can be significant [factor 10x], see https://ieeexplore.ieee.org/abstract/document/11050131

lkstrp:
🚀

PR #1010, opened 2024-08-23
feat: add static types

lkstrp:
- Adds type annotation everywhere
- Adds `mypy` as type checker to `pre-commit`
  - See  rules in `pyproject.toml`
- Also updates versioning scheme (just a deprecation from `setuptools_scm`)

### DIFF
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -158,7 +158,7 @@ class Network(Basic):
         pypsa.components.components. See git repository examples/new_components/.
     override_component_attrs : pypsa.descriptors.Dict of pandas.DataFrame
         If you want to override pypsa.component_attrs, follow its format.
-        See :doc:`/user-guide/components#custom-components` for more information.
+        See :doc:`/user-guide/components` for more information.
     kwargs
         Any remaining attributes to set
 
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -81,6 +81,7 @@ from pypsa.utils import as_index
 
 if TYPE_CHECKING:
     import linopy
+    from scipy.sparese import spmatrix
 
 logger = logging.getLogger(__name__)
 warnings.simplefilter("always", DeprecationWarning)
@@ -1519,11 +1520,11 @@ class SubNetwork(Common):
     pqs: pd.Index
     pvs: pd.Index
     slack_bus: str
-    B: Any
-    K: Any
-    C: Any
-    PTDF: Any
-    BODF: Any
+    B: spmatrix
+    K: spmatrix
+    C: spmatrix
+    PTDF: spmatrix
+    BODF: spmatrix
 
     list_name = "sub_networks"
 
--- a/pypsa/components.py
+++ b/pypsa/components.py
@@ -1,13 +1,15 @@
 """Power system components."""
 
+from __future__ import annotations
+
 import copy
 import logging
 import os
 import warnings
 from collections import namedtuple
-from collections.abc import Collection
+from collections.abc import Collection, Iterator, Sequence
 from pathlib import Path
-from typing import Union
+from typing import TYPE_CHECKING, Any
 from weakref import ref
 
 import geopandas as gpd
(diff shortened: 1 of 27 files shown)


fneum:
Some issues with Python 3.8?

lkstrp:
@fneum All fixed now, but no idea what is happening with codecov

fneum:
I'll wait until the Codecov is fixed with the review. The diff is a bit hard to read with all the Codecov warnings :waffle:

lkstrp:
Codecov issue fixed in #1014

pypsa/components.py
```diff
@@ -480,7 +546,7 @@ def df(self, component_name):
         """
         return getattr(self, self.components[component_name]["list_name"])
 
-    def pnl(self, component_name):
+    def pnl(self, component_name: str) -> Dict:
```
lkstrp, line 549:
```suggestion
    def pnl(self, component_name: str) -> dict:
```

pypsa/components.py
```diff
@@ -1441,9 +1511,26 @@ class SubNetwork(Common):
...
+    buses_o: pd.Index
+    pvpqs: pd.Index
+    pqs: pd.Index
+    pvs: pd.Index
+    slack_bus: str
+    B: Any
+    K: Any
+    C: Any
+    PTDF: Any
+    BODF: Any
```
lkstrp, line 1526: define type sparse matrix

pypsa/graph.py
```diff
@@ -132,7 +149,10 @@ def adjacency_matrix(
             sel = c.ind
         else:
             active = get_active_assets(
-                network, c.name, investment_period, network.snapshots
+                # TODO: get_active_assets takes 3 arguments, no snapshots
+                network,
+                c.name,
+                investment_period,  # network.snapshots
```
lkstrp, line 155:
```suggestion
                network,
                c.name,
                investment_period,
```

pypsa/io.py
```diff
@@ -18,6 +24,14 @@
...
+if TYPE_CHECKING:
+    from typing import TracebackType  # type: ignore
+
+    import xarray as xr
+    from pandapower.auxiliary import pandapowerNet
+
+    from pypsa import Network
+
 try:
     import xarray as xr
```
lkstrp, line 36:
```suggestion
import xarray as xr
```

pypsa/pf.py
```diff
@@ -956,15 +953,30 @@ def apply_transformer_types(network):
     # TODO: status, rate_A
 
 
-def wye_to_delta(z1, z2, z3):
+def wye_to_delta(
+    z1: int | float,
+    z2: int | float,
+    z3: int | float,
```
lkstrp, line 959:
```suggestion
    z1: float,
    z2: float,
    z3: float,
```

fneum (approved):
LGTM

FabianHofmann:
wow, such a good job @lkstrp. since mypy in pre-commit is not always checking the whole package (only touched section) we could add mypy to the ci.yaml. Also if we want to be strict, we disallow non-typed functions by adding these

```
[tool.mypy]
disallow_any_generics = true
disallow_untyped_defs = true
```

lkstrp:
> since mypy in pre-commit is not always checking the whole package

I wish it would. But `pre-commit.ci` always checks all files. Just the local installed hook checks touched sections only. I am looking for a way to check only touched sections in the CI anyway, if you have something in mind there I would be glad. 
We could think about moving `mypy `to GitHub actions:
- To keep the local hook super fast, but I think it's fine. It is still pretty fast when cached.
- To allow for more dependency type checks. Right now we check for `numpy`, `pandas` and `xarray`. The pre-commit storage is not large enough for more. But these checks are also less important and we don't have many more typed dependencies (mainly `linopy`).

The mypy config got lost, thanks for the catch.

https://github.com/PyPSA/PyPSA/blob/57290e4ee834ce75777f3d6920fcb6a44ed91c04/pyproject.toml#L104C1-L118C28

`disallow_any_generics = true` is not enabled yet. Might happen when I do some structural changes as well. I tried to stick to types here.

pypsa/components.py
lkstrp: @fneum 
Just FYI, there is `dict` and `from typing import Dict` for type checks. Both do the same but we wanna use the built-in `dict`.
But this refers to our PyPSA special `Dict` object (should be renamed):

https://github.com/PyPSA/PyPSA/blob/8093c685c87d9666e7e7dd3bdb35f7a4ca29f12f/pypsa/descriptors.py#L22-L69

FabianHofmann:
yes, I'd vote up for adding mypy to the CI.yaml 

To mirror the CI test routine locally, what we have in enertrag's repo is a Makefile (we used to have that in pypsa as well but nobody used it). One rule is `make checks` which runs the same things as the CI on GH including mypy. All together we then have pre-commit for quick checks (formatting/linting) and the `make checks` for full unit test and mypy. (I think langchain has it too)

PR #1009, opened 2024-08-23
Feature: Adding interactive n.explore() function

bobbyxng:
## Changes proposed in this Pull Request
- Added basic n.explore() function based on folium and geopandas to interactively explore the network.
image
- Function will detect automatically what components are available (e.g. buses, links, transformers, generators, storage_units, loads)
- Layer control can be expanded to select rendered components
- Full functionality of folium map, e.g. map can be exported to self-contained html.

Example:
```
n = pypsa.Network("network.nc")
map = n.explore()  # folium map stored to map
map  # Call map to show interactive map e.g. in jupyter notebook
map.save("map_export.html")  # to create self-contained html.
```

## Checklist

- [X] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [X] Newly introduced dependencies are added to `environment.yaml`,  (folium is pretty heavy, should be installed manually).
- [X] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [X] I consent to the release of this PR's code under the MIT license.

PR #1011, opened 2024-08-23
Review suggestions for #1009: `n.explore()`

fneum:


bobbyxng (approved):
Tested it, definitely much more efficient! :)

PR #1023, opened 2024-09-06
cloud object storage io with cloudpathlib

ollie-bell:
Closes # (if applicable).

Enable import and export directly from files in cloud object storage. Optional dependency on `cloudpathlib`, installable via `pypsa[cloudpath]`.

## Checklist

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -2,17 +2,17 @@
 Release Notes
 #######################
 
-Upcoming Release
-================
+.. Upcoming Release
+.. ================
 
-.. warning:: 
+.. .. warning:: 
   
-  The features listed below are not released yet, but will be part of the next release! 
-  To use the features already you have to install the ``master`` branch, e.g. 
-  ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
+..   The features listed below are not released yet, but will be part of the next release! 
+..   To use the features already you have to install the ``master`` branch, e.g. 
+..   ``pip install git+https://github.com/pypsa/pypsa#egg=pypsa``.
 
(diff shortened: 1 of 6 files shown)


lkstrp (changes requested):
This is a neat integration, thanks!
- I assume the `cloudpathlib.AnyPath` will handle all normal Paths or strings as well, just like `pathlib` would do?
- Could you add a very small note to the `path` argument in the the docstrings as well?
- And maybe also explicitly list AWS S3, Google Cloud Storage, and Azure Blob etc. in the docs section, so it can be found via search.

ollie-bell:
> This is a neat integration, thanks!
> 
> * I assume the `cloudpathlib.AnyPath` will handle all normal Paths or strings as well, just like `pathlib` would do?
> * Could you add a very small note to the `path` argument in the the docstrings as well?
> * And maybe also explicitly list AWS S3, Google Cloud Storage, and Azure Blob etc. in the docs section, so it can be found via search.
> * And a small test would be great

Yep! When you give a string representing a local file to `AnyPath` it constructs a `pathlib.Path` object as normal.

I've added the documentation changes you requested.

Proper tests are unfortunately a bit tricky here, because it requires having a cloud storage bucket which you can authenticate to with read/write access. As far as I am aware there are no public buckets with write access to use for testing. It would be possible to set up a free tier / very low cost bucket to use for testing in CI only though, but this a bit cumbersome to maintain 😰

However, I have added a small test which checks that when local paths are used then pathlib.Path is instantiated as normal.

lkstrp (approved):
Great, thank you!

Just the pathlib check is fine.

FabianHofmann:
very cool!

PR #1028, opened 2024-09-09
refactor: introduce naming conventions

lkstrp:
Start to introduce common naming convention. This one changes all `network` (and related, like `_parent`) to `n`. 

Since most of it is network object internal anyway, the API changes are pretty minor. But if the decorator is used, it's also backwards compatible with a `DeprecationWarning`.

If you agree and this is what you have in mind for a common naming conventions, next up is:
- [x] `network` -> `n`
- [x] `sub_n`, `sn`, `sub` -> `sub_network`
- [x] `subnetwork` (docs) -> `sub-network`
- [ ] `snapshots` -> `sns` & `snapshots` in api
- [ ] `Component` (class) -> `c`
- [ ] component name (str), `cls_name`, `class_name` -> `c_name`
- [ ] component list name (str), `list_name` -> `c_list_name`
- [ ] variables representing attributes like `p`, `q`, `p_nom` -> `attr`
- [x] `dataframe` -> `df`
- [x] `df` -> `static`
- [x] `pnl`, `varying`, `series` etc -> `dynamic` 

Feel free to add more

### DIFF
--- a/pypsa/clustering/spatial.py
+++ b/pypsa/clustering/spatial.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any
 import networkx as nx
 import numpy as np
 import pandas as pd
+from deprecation import deprecated
 from packaging.version import Version, parse
 from pandas import Series
 
@@ -455,6 +456,15 @@ class Clustering:
     n: Any
     busmap: pd.Series
     linemap: pd.Series
+
+    @property
+    @deprecated(
+        deprecated_in="0.32",
+        removed_in="1.0",
+        details="Use `clustering.n` instead.",
+    )
+    def network(self) -> Network:
+        return self.n
 
 
 def get_clustering_from_busmap(
--- a/pypsa/clustering/spatial.py
+++ b/pypsa/clustering/spatial.py
@@ -181,7 +181,7 @@ def aggregateoneport(
 
     Parameters
     ----------
-    network : Network
+    n : Network
         The network containing the generators.
     busmap : dict
         A dictionary mapping old bus IDs to new bus IDs.
@@ -447,7 +447,7 @@ def aggregatelines(
 
 @dataclass
 class Clustering:
-    network: Any
+    n: Any
     busmap: pd.Series
     linemap: pd.Series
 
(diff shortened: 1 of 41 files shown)


FabianHofmann:
love it :heart: 

I always have the feeling that `series` is a bit confusing in context of pandas. I think I would prefer `dynamic`/`varying` instead

fneum:
Yes, please! Agree with @FabianHofmann about “varying”

lkstrp:
@fneum @FabianHofmann 
Again, quite a few changes, but nothing that touches too much API/functionality.
- I have not yet changed the components. Would fix that in one go when we also point the component class to n.components (see #1038) in another pr.
- `n.df` and `n.pnl` are not deprecated (yet), they are just aliases for the new names.
- I will double check with pypsa-eur, but otherwise this is good to go.

pypsa/graph.py
```diff
@@ -51,16 +53,16 @@ def graph(
...
         else:
             branch_components = set(branch_components)
-        buses_i = network.buses.index
-    elif isinstance(network, components.SubNetwork):
+        buses_i = n.buses.index
+    elif isinstance(n, components.SubNetwork):
         if branch_components is None:
-            branch_components = network.network.passive_branch_components
-        buses_i = network.buses_i()
+            branch_components = n.n.passive_branch_components
```
FabianHofmann, line 64: I am not sure whether subnetworks should rather have the attribute `network` instead of `n`.

pypsa/clustering/spatial.py
```diff
@@ -517,39 +532,39 @@ def get_clustering_from_busmap(
...
-            for attr, df in generators_pnl.items():
+            for attr, df in generators_dynamic.items():
                 if not df.empty:
-                    io._import_series_from_dataframe(clustered, df, "Generator", attr)
+                    io._import_series_from_df(clustered, df, "Generator", attr)
 
     for one_port in aggregate_one_ports:
         one_port_components.remove(one_port)
-        new_df, new_pnl = aggregateoneport(
+        new_df, new_dynamic = aggregateoneport(
```
FabianHofmann, line 543: this could also be named new_static instead of new_df

pypsa/graph.py
lkstrp: Hmm, I don't think so. All other related classes or accessors now use `self.n` instead of `network`, `_parent`, etc. Not sure why this should be different for `SubNetwork`?

FabianHofmann (approved):

PR #1034, opened 2024-09-13
add prices in statistics

p-glaum:
Closes # (if applicable).

Adds the metric `prices` to statistics to calculate the prices of the buses. They can be calculated "time" or "load" weighted.

Not happy with the implementation yet, but I already wanted to add it as a draft PR that the draft is not lost somewhere. 

## Checklist

### DIFF
--- a/pypsa/statistics.py
+++ b/pypsa/statistics.py
@@ -1245,3 +1245,52 @@ class StatisticsAccessor:
         df.attrs["name"] = "Market Value"
         df.attrs["unit"] = "currency / MWh"
         return df
+
+    def prices(
+        self,
+        price_weighting: str = "load",
+        bus_carrier: Sequence[str] | str | None = None,
+    ) -> pd.DataFrame:
+        """
+        Calculates the prices at each in the network in currency/unit_{bus_carrier}.
+
+        Parameters
+        ----------
+        price_weighting : str, optional
+            Type of price weighting to use. If 'load' the prices are weighted by the load of the buses and if time they are weighted by snapshot weightings. Defaults to 'load'.
+        bus_carrier :  Sequence[str] | str, optional
+            Carrier of the buses for which the prices should be calculated. If None, all buses are considered. Defaults to None.
+        """
+
+        def get_price_weightings(price_weighting: str) -> pd.DataFrame:
+            if price_weighting == "load":
+                weights = (
+                    self.withdrawal(
+                        groupby=get_bus_and_carrier_and_bus_carrier,
+                        nice_names=False,
+                        aggregate_time=False,
+                    )
+                    .groupby("bus")
+                    .sum()
+                    .T
+                )
+            else:
+                weights = self._parent.snapshot_weightings.objective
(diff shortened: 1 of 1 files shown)


fneum:
@lkstrp I dabbled a bit with @p-glaum's implementation, refined it a bit, and it does now what I would like it to do. So from that side it's ready.

But the structure is a bit different from the other statistics functions (not component-based). So I have a few questions:

- There is an auxiliary `n = self._n` that I don't know how to remove.

- Is it a problem that the function `n.statistics.prices()` does not take all the arguments that the other statistics functions take? (It should not show up in `n.statistics()`)

- What do I have to do to configure `n.statistics.prices.plot()`. Will it work even though structure is different? (edit: this is where CI currently fails)

- Would this `aggregate_time=False` implementation be bad style?:

```py
            prices = prices.loc[:, mask]
        
        if not aggregate_time:
            return prices.T 
            
        if weighting == 'load':
            ...
```

lkstrp:
>  There is an auxiliary `n = self._n` that I don't know how to remove.

The comment refers to that `self.n` should be made internal, which it is now, so this can be removed. Using either `n = self._n` or simply `self._n` is fine

>  Is it a problem that the function `n.statistics.prices()` does not take all the arguments that the other statistics functions take? (It should not show up in `n.statistics()`)

It is a bit. Currently, the logic is that the "Parameters" section defines parameters that all statistic functions use, and only the "Other Parameters" section defines additional ones. Ideally, we would stick to that and just define different defaults or not allow changes to them. I don't know enough about the logic, but I assume that we never have prices for any components other than per bus per snapshot? We should then at least allow all the "base" parameters and raise a specific warning with a short explanation that they cannot be used with `.prices` when passing them.

> What do I have to do to configure `n.statistics.prices.plot()`. Will it work even though structure is different? (edit: this is where CI currently fails)

In general, you only need to add it to schema.py and, if appropriate, define different defaults for the arguments of the plotting method. Since so many parameters are dropped, further changes are needed, since some of these are expected. I will take a look once we are sure about the logic.

> Would this `aggregate_time=False` implementation be bad style?:

Again the API is the issue. `aggregate_time=False` would work fine, but it also takes by default `sum` in most of the other methods, which here is the same than passing `weightings="time"`. Which is very arbitrary and what is happening if you mix things? So in here I would include "load" as default in `aggregate_time` and drop `weightings` with `weightings="time"`. Which has the downfall again that `aggregate_time` combines turning aggregation off and on, adds a basic aggregation option (via pandas defaults) and the "special" `load` aggregation, which might be very confusing. But let's tackle that in another issue/ general statistics parameter redefining

lkstrp:
In summary, I mirrored the API, but the arguments are more strict:
- `aggregate_time=False` is now possible (with new naming `groupby_time`) and no method can be passed
-  `aggregate_buses` is aligned to `groupby` and just takes less options then. Instead of True, you need to pass `='bus_carrier`. 

Plotting works again, but they still need some general improvements. Not in here, though.

PR #1038, opened 2024-09-17
Add active attribute to components

FabianHofmann:
This PR introduces a new component `active` which, as discussed, allow to enable/disable assets from pypsa core functionalities. These are 

* optimization
* power flow
* statistics

When set to inactive, the optimization masks the corresponding asset out. The asset then still appears in the index but does not contain any value. To save memory (in particular for the myopic optimization), we might change this to a harder filter such that inactive assets do not even appear in the variable blocks in the first place. My preference would be to tackle that after #848. 

The PR further introduces as `Component` class. Note that this is purely non-breaking and non-intrusive. We had a `Component` object before, which was a `namedtuple`. With the new class we gain more control when iterating over components. It might as well pave the way to a more formal `Component` behaviour as discussed on #936.   
 
(Do not merge yet, I would like to see some more tests on the optimization and power flow.)

## Checklist


problem.lp
fneum, line 1: Ups :)

pypsa/components.py
```diff
@@ -115,34 +115,92 @@
...
+    """
 
-    def __init__(self, name: str = ""):
-        self.name = name
+    name: str
+    network: Network  # in future we probably avoid this kind of circular dependency
+    list_name: str
+    attrs: pd.DataFrame
+    df: pd.DataFrame
+    pnl: Dict | dict
```
lkstrp, line 144: Can we drop `dict`?

FabianHofmann:
thanks for the comments @lkstrp! As for the active assets function, what is needed quite regularly is the index of active assets. With the current `get_active_assets` that would be

`c.df.index[c.get_active_assets()]`

which I shortcut to

`c.active_assets`

but I agree this is confusing. I will revert that and only define `get_active_assets()` for the `Component` class. At some point, we should think about a better naming. As I see, the name `get_active_assets` suggest that you get a listlike object with all active assets, but instead you get a boolean series indicating whether it is active or not. With that proposal, I wanted to introduce a better naming, but let's go step by step.

FabianHofmann:
For more background: The general issue here is that the activity of an asset has two sources of information 

a) the global (new) attribute `active` which defines activity for all pypsa functionality. This should be independent of any build_year or investment year and enable to fully control whether an assets is there or not. 

b) the investment year dependent activity which adds on top and is only considered when doing and multi investment optimization. This is dependent on build_year (input), life_time (input) and investment_year (function argument). 

When thinking about different approaches, my feeling is that the rule that static `active` columns should remain an input only thus not be touched/altered by the code.

pypsa/components.py
FabianHofmann: sure

pypsa/components.py
```diff
@@ -115,34 +115,92 @@
...
+        -------
+            pd.Index
+                An index of active assets.
 
-    def __init__(self, network: Network, name: str = ""):
-        Basic.__init__(self, name)
-        self._network = ref(network)
+        """
+        if "active" in self.df.columns:
+            return self.df[self.df["active"]].index
```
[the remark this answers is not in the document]
FabianHofmann: removed that because is it actually not needed. the active column is inserted with value True is loading a network from a previous version

pypsa/statistics.py
```diff
@@ -247,12 +247,15 @@ def aggregate_timeseries(
...
     n: Network, c: str, df: pd.Series | pd.DataFrame
-) -> pd.DataFrame:
+) -> pd.Series | pd.DataFrame:
     """
     For static values iterate over periods and concat values.
     """
-    if not isinstance(n.snapshots, pd.MultiIndex) or isinstance(df, pd.DataFrame):
+
+    if isinstance(df, pd.DataFrame):
         return df
```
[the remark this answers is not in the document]
FabianHofmann: yes

FabianHofmann:
and about the comment about the dataclass, I am happy to remove the dataclass decorator if this is what you meant

lkstrp:
> and about the comment about the dataclass, I am happy to remove the dataclass decorator if this is what you meant

Na, keep it is a it is. It is also so much cleaner than before

pypsa/components.py
```diff
@@ -115,34 +115,72 @@
...
 
+        Returns
+        -------
+            pd.Series
+                A series containing the identifiers of active assets.
 
-Component = namedtuple("Component", ["name", "list_name", "attrs", "df", "pnl", "ind"])
+        """
+        if investment_period is None:
+            return self.df.active
```
lkstrp, line 179: This check can be ignored now

lkstrp (approved):

lkstrp:
As discussed:
- Completely moves `active_asset` logic to `Component` class
- Removes `Component` reference to `Network`
- Moves it to own module (`definitions`). That's how we named it in pypsa-eur. Not sure if it's the best name. But a composition is needed, otherwise we get circular import errors.

FabianHofmann:
looks great, thanks lukas!

PR #1040, opened 2024-09-22
Stochastic Network: refactor and add tests

FabianHofmann:


pypsa/components.py
```diff
@@ -1911,29 +1895,19 @@ def get_bounds_pu(
...
-                _sns = self.snapshots.get_level_values("snapshot").unique()
-                new_index = pd.MultiIndex.from_product(
-                    [max_hours.index, _sns], names=["scenario", "snapshot"]
-                )
-                max_pu = max_hours.loc[max_hours.index.repeat(len(_sns))].set_index(
-                    new_index
-                )
-                min_pu = pd.DataFrame(0, *max_pu.axes)
+            if attr == "state_of_charge":
+                max_pu = self.get_switchable_as_dense(c, "max_hours", sns, inds=index)
```
Irieo, line 1899: @FabianHofmann for `state_of_charge` case this line does not work since
`sn.pnl(component).keys()` does not contain `max_hours` attr.
we can (i) roll back to my workaround, or (ii) more generally, make `get_stichable_as_dense` handle such cases. 
go for the latter?

FabianHofmann: yes, let's go for the latter. I'll push a commit

PR #1052, opened 2024-10-07
Bugfix in post_discretization

toniseibold:
The motivation of this PR was, that there are edge cases in which `nom_max >= nom_opt`.
With the following variables
```
nom_opt = 10
unit_size = 10
threshold = 0.3
min_units = 2
nom_max = 10
fractional_last_unit_size = False
```
the function `discretized_capacity` would return `nom=20`.
Since the variable `min_units` is not available as a config it is abolished. Instead if there is a `nom_max` that is smaller than one unit size, the `nom_max` is built regardless of the `fractional_last_unit_size`.
The test was altered as well.

## Checklist


lkstrp (approved):
Looks good.
I just added a deprecation instead of removing the argument, which was introduced before `0.31.0` and would be breaking.

PR #1075, opened 2024-11-07
feat: introduce component class

lkstrp:
Ref #936

### Summary
- Creates an additional layer to move all component-specific data from the network class to a new component class
- Cleans up all cross references of `n.component()`, `n.generators`, `n.generators_t` and the equivalents in sub-networks
- Allows for tons of API changes and improvements because of this extra layer
- Does not change any existing API and also does not add too much new (yet)

#### Components Class
- Structure: `ComponentsData` -> `Components` -> `Generators`, `GenericComponents` etc.
  - `ComponentsData`: Simple data class containing all stored data
    - As discussed, I played around a lot with changing the storage structure here (e.g. `static` and `dynamic`). Especially regarding the extra dimension that comes with stochastic networks.
      - But there is no nice way to map any new structure (no matter if `xarray` or `pandas`) back to the existing API and keep full backwards compatibility. So `static` and `dynamic` are untouched.
  - `Components`: Abstract base class that contains all logic relevant to all components.
    - Abstract, so should not be initialised by itself (could easily be, but this is an API discussion)
  - `Generators`, `GenericComponents` etc: child classes to be used
    - Contains only component-specific logic
    - Could be then used via
       ``` python
       >>> pypsa.Generators(names=[1,2,3], p_nom=5, p_max_pu=p_max_pu)
       PyPSA 'Generator' Components
       ----------------------------
       Components: 3
       ```
      - Adding components during initialisation is not yet implemented
- Attached status
  - To allow for logic that needs other information from the network, but lets a Components object be self-contained, it can be either attached (`self.n = None`) or not attached (`self.n = n`).
    - `self.n` points to the attached network, `self.attached` is just a boolean indicating the same, and `self.n_save` can be used for functionality that requires an attached in any case
  - Right now this only affects `self.get_active_assets()`, which needs access to investment periods (and even that could be integrated into the components class).
  - If indices, crs, etc. match, would be checked during the attachment process.
    - e.g:
       ``` python
       n = pypsa.Network()
       generators =  pypsa.Generators(names=[1,2,3], p_nom=5, p_max_pu=p_max_pu)
       n.add(generators)
       ```
      - Also not implemented yet, since this PR doesn't move existing logic
- `SubNetwork`
  - Components in Sub-Networks also just return a filtered view (of subnetwork specific components) of the parent Network
  - `sn.generators()`, `sn.generators_i()` , `sn.components()` is not deprecated yet, but points to the same generalized view

#### Components Store
- All components are stored behind `n.components`.
  - e.g.
    ``` python
    n.components.generators.static # same as n.generators
    n.components.generators.dynamic # same as n.generators_t
    ```
    - This is just where the data is stored and does not have to be the preferred interface. It could be pointed to from anywhere (as done already). 
    - Full backwards compatibility with the existing `n.components` is maintained.
- The store is similar to the existing custom `Dict`, but with the possibility of some extra functionality:
  - e.g.
    ``` python
    >>> components = ComponentsStore()
    >>> components["generator"] = Generators()
    >>> components["storage"] = Storage()
    >>> components["generator"] = 'generator class instance
    generator instance
    >>> components[["generator", "storage"]]
    >>> components[["generator", "storage"]]
    [generator instance, storage instance].
    ```
    - which could replace `n.iterate_components()`.
  - Other things are possible. I am not sure how the iterator should work:
    - Same behaviour as dict (`.keys`, `.values`, `.items` could be used)
      ``` python
      >>> for c_name in n.components:
      >>> print(c_name)
      generators
      ```
      - So only the key (`str`) is returned. To do anything you would always have to state `c = n.components[c_name]` and `c_name` could also be accessed via `c.name`.
    - Another option (and my vote) is list behaviour, which would also fit the above multiple indexing
    ``` python
     >>> for c in n.components:
     >>> print(c)
    generator instance
     ```

#### Components Type
- Component types (data stored in `components.csv`, `component_attrs/`) are now stored at pypsa module level instead of dynamically during network initialisation.
- They are stored in an immutable `ComponentsTypeInfo` data class with all fixed attributes: `name`, `list_name`, `description`, `category`, `defaults` (previously `attrs`) and `standard_types` (which I moved there).
- Each `Components` instance is associated with a component type (`ct`) and therefore does not need to store any of the above information.
  - e.g. only contains `ct`, `n`, `static` and `dynamic`.
- Custom components can be added at module level and then added during network initialisation via a list of components.

### ToDos
In this PR
- [x] Add tests
- [x] Check with pypsa-eur
- [x] Performance tests
- [x] Implement custom component class workflow
- [x] Better alias access

API discussion and Docs, moving logic, new features, `xarray` view and so on, I would move to other PRs.


doc/user-guide/components.rst
```diff
@@ -478,17 +478,17 @@ Custom Components
...
-``pypsa/component_attrs/*.csv``.
-
-``components`` is a pandas.DataFrame with the component ``name``,
-``list_name`` and ``description``. ``component_attrs`` is a
-pypsa.descriptors.Dict of pandas.DataFrame with the attribute
-properties for each component.  Just follow the formatting for the
-standard components.
+components by passing the arguments ``override_components`` and 
+``override_component_attrs`` when initialising a network via 
+:meth:`pypsa.Network() <pypsa.Network>`.
```
fneum, line 483: What happens now if I read a network from file which has custom components?

pypsa/components/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+            f"Cannot import '{name}' from 'pypsa.components'. "
+            "Import it from 'pypsa' instead."
+        )
+        raise ImportError(msg)
+    return globals()[name]
+
+
+__all__ = [
+    "Component",
+    "Components",
```
fneum, line 23: What is the difference between `Component` and `Components`?

pypsa/components/abstract.py
```diff
@@ -0,0 +1,252 @@
...
+        )
+
+    @staticmethod
+    def _get_data_containers(ct: ComponentTypeInfo) -> tuple[pd.DataFrame, Dict]:
+        static_dtypes = ct.defaults.loc[ct.defaults.static, "dtype"].drop(["name"])
+        if ct.name == "Shape":
+            # TODO: Needs general default crs
+            crs = CRS.from_epsg(
+                "4326"
+            )  # if n is None else n.crs #TODO attach mechanism
```
fneum, line 111: 4326 is good default

pypsa/components/abstract.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+        else:
+            static = pd.DataFrame(
+                {k: pd.Series(dtype=d) for k, d in static_dtypes.items()},
+                columns=static_dtypes.index,
+            )
+        static.index.name = ct.name
+
+        # # it's currently hard to imagine non-float series,
+        # but this could be generalised
```
fneum, line 125: I agree! Don't think non-float time series are very relevant.

pypsa/components/abstract.py
```diff
@@ -0,0 +1,252 @@
...
+        dynamic = Dict()
+        # TODO: Needs general default
+        snapshots = pd.Index(
+            ["now"]
+        )  # if n is None else n.snapshots #TODO attach mechanism
```
fneum, line 130: "now" is a good default, not?

pypsa/components/abstract.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def get(self, attribute_name: str, default: Any = None) -> Any:
+        return getattr(self, attribute_name, default)
+
+    @property
+    def attached(self) -> bool:
+        return self.n is not None
+
+    @property
+    def n_save(self) -> Any:
```
fneum, line 179: Not sure I got why `n_save` is required.

pypsa/components/components.py
```diff
@@ -0,0 +1,126 @@
...
+from pypsa.definitions.structures import Dict
+
+logger = logging.getLogger(__name__)
+
+
+class GenericComponents(Components):
+    """
+    Generic Components class
+
+    This class is used for components that do not have a specific class implementation.
```
fneum, line 27: For which components would this be the case?

pypsa/components/components.py
```diff
@@ -0,0 +1,126 @@
...
+        It is not recommended to use this class outside of PyPSA.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+
+CLASS_MAPPING = {
+    "Generator": Generators,
+}
```
fneum, line 59: Am I correct that this would have to be expanded by the other components before merging?

pypsa/definitions/components.py
```diff
@@ -1,120 +1,126 @@
...
+        """
+        Index single and multiple items from the dictionary.
+
+        Similar behavior to pandas.DataFrame.__getitem__.
+
+        Examples
+        --------
+        >>> components = ComponentsStore()
+        >>> components["generator"] = Generators()
+        >>> components["storage"] = Storage()
```
fneum, line 89: Shouldn't this be `StorageUnits()` and `Stores()`?

We also need some consistency for capitalization and plural/singular. So far, we had smallcaps+plural and capitalized+singular.

pypsa/definitions/components.py
```diff
@@ -1,120 +1,126 @@
...
-            only the active attribute is considered and build year and lifetime are ignored.
-            If multiple periods are given the mask is True if the component is active in any
-            of the given periods.
-
-        Returns
-        -------
-        pd.Series
-            Boolean mask for active components
+        Make class iterable over the values.
+        #TODO: Do we want this? So list like behavior or dict like behavior?
```
fneum, line 122: Dict-like might be better? But list-like is also ok, if component type can be inferred in another way.

pypsa/definitions/components.py
```diff
@@ -1,120 +1,126 @@
...
-        active = {}
-        for period in np.atleast_1d(investment_period):
-            if period not in self.investment_periods:
-                raise ValueError("Investment period not in `n.investment_periods`")
-            active[period] = self.static.eval(
-                "build_year <= @period < build_year + lifetime"
-            )
-        return pd.DataFrame(active).any(axis=1) & self.static.active
+        for key, value in super().items():
+            if key != "Network":  # TODO Drop Network completly?
```
fneum, line 125: If I understand correctly, "Network" is not a component anymore. So it can be dropped?

coroa (commented):
Puh that thing is huge, i only covered a small bit of the full amount of code. Most of it is looking good, but well alot.

doc/user-guide/components.rst
lkstrp: Should work as before as `io.py` handles component data independently from network custom components in network initialization.

pypsa/components/abstract.py
lkstrp: Yes, the TODO just means it would be better to define the default somewhere and import it, instead of defining it multiple times (in components and network). Done now via `constants.py`.

pypsa/definitions/components.py
lkstrp: Removed. If there are people who rely on `n.components.Network`, this will break now.

pypsa/clustering/spatial.py
```diff
@@ -485,7 +485,7 @@ def get_clustering_from_busmap(
 ) -> Clustering:
     if aggregate_one_ports is None:
         aggregate_one_ports = {}
-    from pypsa.components import Network
+    from pypsa import Network
```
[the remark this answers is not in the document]
lkstrp: I like, thanks!

pypsa/definitions/components.py
```diff
@@ -1,120 +1,126 @@
...
+        >>> components["generator"] = Generators()
+        >>> components["storage"] = Storage()
+        >>> components["generator"]
+        'Generator class instance'
+        >>> components[["generator", "storage"]]
+        >>> components[["generator", "storage"]]
+        ['Generator class instance', 'Storage class instance']
+        """
+        if isinstance(item, (list, set)):
+            return [self[key] for key in item]
```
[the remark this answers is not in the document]
lkstrp: Hmm. I see, but not sure if I agree. For DataFrame consistency it makes sense and for them it also is useful since you wanna use DataFrame specific features on your sliced `df`. 
But here there is no use case where you would need any ComponentsStore functionality on a sliced version. And then it just complicates things for no reason. If this returns a list, the ComponentsStore is always a complete representation of all components for a specific Network.

FabianHofmann:
@lkstrp I still can't find the time to look into this (even though I intrinsically would like to). So merging after the tests pass would be fine for me as this will unleash so much new potential for the package 🤘

FabianHofmann:
@lkstrp I still can't find the time to look into this (even though I intrinsically would like to). So merging after the tests pass would be fine for me as this will unleash so much new potential for the package 🤘

fneum:
Ok, to merge from my side when these two TODOs are checked:
- Check with pypsa-eur
- Performance tests

lkstrp:
I will also add a couple of improvements and more checks for custom components class since I got some example use cases now. But will also not merge before `v0.32` is released.

PR #1078, opened 2024-11-14
feat: add flexible selector for groupby in statistics module

FabianHofmann:
* The statistics and expressions modules now support setting the groupby argument by keys to allow for more flexibility in grouping. For example, `n.statistics.energy_balance(groupby=["bus_carrier", "carrier"])` groups the energy balance by bus carrier and carrier. 
* Added custom grouping function registration via `Groupers` class:
    - Register custom groupers using `n.statistics.groupers.register_grouper(key, function)`
    - The key will be used as identifier in the `groupby` argument
    - Custom grouper functions must accept:
      * n (Network): The PyPSA network instance
      * c (str): Component name
      * port (str): Component port as integer string
      * nice_names (bool, optional): Whether to use nice carrier names
    - The function must return a pandas Series with the same length as the component index

  * Built-in groupers include: "carrier", "bus_carrier", "name", "bus", "country", "unit"

  * Example:
    ```python
    def my_custom_grouper(n, c, port=""):
        return n.static(c)["my_column"].rename("custom")
        
    n.statistics.groupers.register_grouper("custom", my_custom_grouper)
    n.statistics.energy_balance(groupby=["custom", "carrier"])
    ```

## Checklist

- [x] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [x] Unit tests for new features were added (if applicable).
- [x] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [x] I consent to the release of this PR's code under the MIT license.

PR #1079, opened 2024-11-18
fix: fix missing indices (#1044 follow up)

lkstrp:

- #1044 lead to inconsistent indices for empty results when using `n.statistics`
- This PR introduces very simple grouper classes to attach index information to a grouper function to fix the problem.

- This is just a hotfix. In another PR with some changes for #1078, I would improve those by
  - merge multi-groupers to just be a concatenation of single groupers
  - better handle any passed callable

Let me know if you have any objections in general to a grouper class structure @FabianHofmann 

## Checklist

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -1,0 +1,110 @@
+# SPDX-FileCopyrightText: : 2021-2024 The PyPSA-Eur Authors
+#
+# SPDX-License-Identifier: CC0-1.0
+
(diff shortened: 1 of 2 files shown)


FabianHofmann:
let me have a short look whether there can be an easier solution by allowing to group empty linear expression in linopy. if that is the case we can remove the non-grouped return and have the exact behavior as before

lkstrp:
Closed in favor of

Issue #979, opened 2024-07-12
Example: Update statistics example with newest functionality


fneum:
Part of the statistics example (https://pypsa.readthedocs.io/en/latest/examples/statistics.html) might be outdated.

It would be good if it covered the new features (e.g. https://pypsa.readthedocs.io/en/latest/api/statistics.html#statistic-groupers).

xref https://groups.google.com/g/pypsa/c/cz1OSnKzkhc

PR #1093, opened 2024-11-27
feat: groupers and statistics package refactor

lkstrp:
Closes #961

- Addition to #1078 
  - Probably we wanna keep that as two PRs though, so merge 1078 first and change base
- Restructure `statistics.py` to `statistics` package
- Put groupers on module level `pypsa.statistics.groupers`
  - Merge single and multiple groupers instead of many get_x methods, e.g.
    - `groupers.bus` or `groupers['bus']` and
    -  groupers["bus", "carrier"]
  - Deprecates all `get_x` methods and also any access via `n`

Todos:
- [x] Tests
- [x] Docs
- [x] pass single str as groupby
- [ ] maybe even auto register for custom callables, this would clean up a lot again

## Checklist

### DIFF
--- a/pypsa/optimization/expressions.py
+++ b/pypsa/optimization/expressions.py
@@ -16,11 +16,11 @@ from xarray import DataArray
 
 from pypsa.descriptors import nominal_attrs
 from pypsa.statistics import (
-    AbstractStatisticsAccessor,
     get_transmission_branches,
     get_weightings,
     port_efficiency,
 )
+from pypsa.statistics.abstract import AbstractStatisticsAccessor
 from pypsa.utils import pass_none_if_keyerror
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


pypsa/optimization/expressions.py
```diff
@@ -136,7 +139,7 @@ def capex(
         comps: Sequence[str] | str | None = None,
         aggregate_groups: str = "sum",
         aggregate_across_components: bool = False,
-        groupby: list[str] | Callable | None = None,
```
FabianHofmann, line 139: I would still allow passing list of strings as this is much more convenient. Otherwise one always has to import the grouper for the function arguments

pypsa/optimization/expressions.py
```diff
@@ -340,7 +343,7 @@ def energy_balance(
         aggregate_time: str | bool = "sum",
         aggregate_groups: str = "sum",
         aggregate_across_components: bool = False,
-        groupby: list[str] | Callable | None = ["carrier", "bus_carrier"],
```
FabianHofmann, line 343: this too should be reverted, the default is now the deprecated function.

FabianHofmann (commented):
@lkstrp great structure, let's do it. Though I would keep the support for passing list of strings to the `groupby` argument. This is more convenient as it does not require the initialization of a grouper object

FabianHofmann:
as for the point on the todo list, passing a single string was supported as far as I remember (though it was missing in the type annotation)

fneum (commented):
Conceptually, I agree with the proposed changes!

There have been quite a few changes in the statistics module lately. Could we perhaps also update the statistics usage example and amend it with a wider variety of grouping operations?

FabianHofmann:
> Conceptually, I agree with the proposed changes!
> 
> There have been quite a few changes in the statistics module lately. Could we perhaps also update the statistics usage example and amend it with a wider variety of grouping operations?

already added a small section in https://github.com/PyPSA/PyPSA/pull/1078 but might be a bit incomplete

lkstrp:
I will update the grouper sections for everything what is deprecated. 
But the docs overhaul also includes docstrings everywhere with an improved API reference section (which I think is even more helpful). And that will be easier with input from Philipp, once he is back.

Ref #979

FabianHofmann:
awesome!

Issue #1105, opened 2024-12-03
Full docstring coverage


lkstrp:
We want to achieve full docstring coverage for 
- easier maintenance
- Better access for new contributors 
- Improved API reference for great docs without double maintenance

While we already have pretty good coverage for most parts of the code, it can be improved. The numpydoc format is used for this. Check out the full style guide [here](https://numpydoc.readthedocs.io/en/latest/format.html). 
All methods and functions should have at least the following [sections](https://numpydoc.readthedocs.io/en/latest/format.html#sections):
1. [short-summary](https://numpydoc.readthedocs.io/en/latest/format.html#short-summary)
2. [extended-summary](https://numpydoc.readthedocs.io/en/latest/format.html#extended-summary)
3. [parameters](https://numpydoc.readthedocs.io/en/latest/format.html#parameters)
4. [returns](https://numpydoc.readthedocs.io/en/latest/format.html#returns)
5. [examples](https://numpydoc.readthedocs.io/en/latest/format.html#examples)
    - Examples will be useful for better documentation, but will also be used as additional tests.

 Other sections such as [see-also](https://numpydoc.readthedocs.io/en/latest/format.html#see-also), [notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) and [references](https://numpydoc.readthedocs.io/en/latest/format.html#references) may also be relevant.

To get to this docstring quality, most things can already be checked using ruff linter rules (which are also included in the pre-commit). To docstring specific ones are not yet enabled for all parts of the code. But to enable them for a specific module to get a todo list of what changes the docstrings still need, you can adapt the following line and add the relevant module (e.g. `"!**/{grouping.py,components.py}" = [ `

https://github.com/PyPSA/PyPSA/blob/04548b3dcdd2f2cd87a77d6b961e1375e375d68c/pyproject.toml#L157

Issue #231, opened 2021-04-30
Improve test coverage


fneum:
There are some modules where test coverage can improve. Let's target step by step:

`io`:
- [ ] import and export functionality, starts with #229 

`components`: 
- [ ] `madd()`
- [ ] `mremove()`
- [ ] `copy()`
- [ ] `__getitem__()` aka `network[:,:]`
- [ ] `consistency_check()`

`linopt`:
- [ ] `get_sol()`
- [ ] `get_dual()`

`linopf`:
- [ ] ramp limits
- [ ] store constraints
- [ ] global constraints
- [ ] `ilopf()`

`opf`:
- [ ] ramp limits, shut down, start up

`pf`:
- [ ] `sub_network_pf_singlebus()`
- [ ] `aggregate_multi_graph()` 

as well as
- [ ] `networkclustering`
- [ ] `plots`
- [ ] `stats`

fneum:
See also https://codecov.io/gh/PyPSA/PyPSA/tree/master/pypsa

rockstaedt:
Hey! I would like to contribute to this issue :) Can you recommend any unit/function I could start with?

FabianHofmann:
Great @rockstaedt! It depends a bit on how deep you are in the pypsa code. I'd say for a first step try to tackle the functions `madd` and `mremove`. 
For `madd`:
1. Test that components are actually added to the static as well as the time dependent dataframe (e.g. n.generators_t.p_max_pu). This can be done via checking the index. 
2. Check that defaults are set correctly. These come from `n.component_attrs` and should fill values which are not specified in `madd` (only test one or two single values) 
3. `madd` should fail with a misspelled component (something like 'Gnerators')
4. Re-adding the same generator name should fail
...

For `mremove`:
1. Check that components are actually removed from static and time-dependent
2. Removal of non-existent components should fail
3. Removal with misspelled component names should fail

The test functions could go into a new test script like `test_components.py`.

rockstaedt:
Great! Thanks 👍🏼 I'm on it!

rockstaedt:
In the above commit, I created the unit tests for `mremove()`. 

How do you want to proceed? Should I already open a pull request?

FabianHofmann:
Wonderful @rockstaedt, looks clean are usable! If you want, you can start a PR in draft mode and continue adding things (at least the madd would be good). As soon as you think you are ready we review and pull it in.

rockstaedt:
[...]

PR #1114, opened 2024-12-19
test: add doctests

lkstrp:
Ref #1105 
Ref #231 

- Add doctests to testing to test any example code in docstrings.
  - This way we can increase test coverage through good documentation, and don't have to add a unit test for every simple case.
- Changes naming convention from Component Type (Generator, Store) etc. to Component Variant

## Checklist


doc/user-guide/design.rst
```diff
@@ -50,7 +50,7 @@ To see which data is stored for each component, see :doc:`/user-guide/components
 Static data
 -----------
 
-For each component type (line, transformer, generator, etc.), which must be
+For each component variant (line, transformer, generator, etc.), which must be
```
fneum, line 53: I see where this comes from (as `class` and `type` are Python jargon), but do you think it's feasible/sensible to name it "component type" or "component class"?

pypsa/networks.py
```diff
@@ -735,6 +730,12 @@ def snapshots(self) -> pd.Index | pd.MultiIndex:
         pypsa.networks.Network.timesteps : Get the timestep level only.
         pypsa.networks.Network.periods : Get the period level only.
 
+        Notes
+        -----
+        Note that Snapshots are a dimension, while timesteps and and periods are
+        only levels of the snapshots dimension, similar to coords in xarray.
+        That is because timesteps and periods are not necessarily unique or entire
```
fneum, line 737:
```suggestion
        That is because timesteps and periods are not necessarily unique or complete
```
This?

pypsa/plot.py
```diff
@@ -905,8 +905,30 @@ def autogenerate_coordinates(n, assign=False, layouter=None):
 
     Examples
     --------
-    >>> autogenerate_coordinates(n)
-    >>> autogenerate_coordinates(n, assign=True, layouter=nx.circle_layout)
+    >>> import pypsa
+    >>> n = pypsa.examples.ac_dc_meshed()
+    >>> autogenerate_coordinates(n) # doctest: +SKIP
```
fneum, line 910: not sure if this function is deterministic

fneum (changes requested):
Looks great, but I am unsure about the name "Component Variant". I find "component type" or "component class" or simply "component(s)" more intuitive. What do you think?

lkstrp:
The thing is, we have a couple of "components":
- `ComponentVariant`: Default values and attributes for each component, `e.g.` generator variant or carrier variant. Basically just the information from `components.csv` and represents data for a single component.
- `Components` class: Stores concrete information for a number of component instances: e.g. `line_1`, `line_2`, similar to the df in n.generators. It references the corresponding `Variant` but does not inherit from it.
  - This also includes subclasses like `Generators`, `GenericComponents` for variant-specific logic.

So "component" and "component class" should better refer to the second one, which is also the one mostly used. I wouldn't start by only distinguishing between a single "component" and multiple "components". That already lead to confusion with the current named tuple implementation. And the "component" namespace is also already taken for backwards compatibility.

I do not like "type" because the `ComponentVariant` also had a `type` attribute (from `components.csv`: `controllable_one_port`, `passive_one_port` etc.). This has been renamed to `category`, but is still accessible via `type`. Then you would have a `Components` class of type `Generators` with type `controllable_one_port`?
Also, not only do you have the Python jargon problem, but Python type and the type/variant attribute would refer to completely different things:
``` python
type(c) # -> "Generators" class
c.type # -> ComponentType instance
c.type.list_name # -> "generators"
c.type.category # -> "controllable_one_port"
c.type.type # -> "controllable_one_port" # Deprecated
```
vs
``` python
type(c) # -> "Generators" class
c.variant # -> ComponentType instance
c.variant.list_name # -> "generators"
c.variant.category # -> "controllable_one_port"
c.variant.type # -> "controllable_one_port" # Deprecated
```

But I agree that "variant" might not be a very intuitive name.

So three options:
1. Rename to "variant"
2. Stick with "type" and rename the access to something like `c.component_type` or `c.ctype`.
3. Come up with something completely different?

Maybe I like the second option most, but there might be another better name as well. Let me know what you think. Also pinging the PyPSA trinity here @FabianHofmann @coroa

pypsa/plot.py
lkstrp: It is not, you need to add `# doctest: +SKIP` for these cases from now on

fneum:
Yes, let's hear other opinions. I would tend to go for option 2.

FabianHofmann:
mmh, I actually don't know, no strong opinion here. If I understand correctly that is about the common structure of component classes I find ComponentVariant nice

lkstrp:
Let's go with option 2!
No need to change the common notation so far, and we just fix the ambiguity with `c.component_type` and `c.ctype`.

PR #1116, opened 2025-01-06
2025 copyright update

lkstrp:


.git-blame-ignore-revs
```diff
@@ -1,6 +1,2 @@
-# SPDX-FileCopyrightText: 2022 The PyPSA Authors
-#
-# SPDX-License-Identifier: CC0-1.0
```
fneum, line 3: Sure these don't need SPDX anymore?

fneum (approved):

.git-blame-ignore-revs
lkstrp: This is just an artifact, probably because the file was copied from eur. Reuse and SPDX was never set up for PyPSA

Issue #1106, opened 2024-12-04
Refactor Plotting Architecture with Accessors


FabianHofmann:
**Description**:

*Proposal:*
To enhance the modularity and extensibility of the Pypsa package's plotting functionality, we may want to restructure the current `network.plot` function into a more versatile accessor-based system. This new structure would allow for a range of plotting types, offering both static and interactive visualizations.

*Benefits:*
- **Modularity**: Easy to extend with new plot types in the future.
- **Clarity**: A clear, organized structure for different supported funcitonalities
- **Extensibility**:  Switch between static and interactive plots for different presentation and analysis puposes and quick checks

*Proposed Accessor Structure:*
1. `n.plot`: First Version of plotting module
   - **`n.plot`**: This will serve as the main accessor for static plots, with subcategories including:
     - `maps`: For network topology on static and interactive maps.
     - `bar`: For bar plots of network statistics.
     - `line`: For time series data.
     - `area`: For stacked area plots, useful for overlapping data like production curves.
   - API is `n.plot.optimal_capacity()`
   - [x] #1156 

2. **`n.iplot`**: Plotly integration
   - A parallel accessor for interactive plots, mirroring the structure of `n.plot`. This will enable users to create interactive visualizations with ease, offering a seamless transition between static and dynamic plots.\
   - [x] Interactive Plotting Module

3. **`n.dashboard()`**: Dashboard integration
   - Plotly plots (also static plots) could be directly integrated into an upcoming dashboard functionality
   - [ ] Dashboard Integration

lkstrp:
Closed with #1156 #1189 and #1212

PR #1118, opened 2025-01-13
Plotting library

FabianHofmann:
Closes #912 #1102 
addresses #1106

Introduce a new plotting accessor scheme, to quickly generate plots based on the statistics module (as described in #1106). The implementation now only works for static plots. Run it with `n.plot.map.energy_balance` or `n.plot.bar.optimal_capacity` etc. 

### Future work
- think about moving code to components class where possible to avoid redundancy  
- extend for multiinvestment networks (have stochastic networks in mind)
- extend for interactive plots, via `iplot`. we can recycle a lot of code here
-  extend for plotting multiple networks at once
 

## Checklist

- [x] make the old namespace `pypsa.plot deprecated in favour` of `pypsa.plot.maps`
- [ ] autogenerate_coordinates was renamed to `apply_layouter` (keep backwards compat)
- [ ] @lkstrp I kept the namespace order now this way, as it helped me structuring things, I think we can still change if to `n.plot.energy_balance.map`, once we agreed on the fundamentals.


lkstrp:
Closed via #1156

FabianHofmann:
@lkstrp 

I have applied the modularization here. This is not a push to merge this instead of #1156 but this should ease the comparison of default conventions

FabianHofmann:
@fneum thanks for testing it out. as far as I see the following points were raised:

* unit scaling in chart plots: yes, that I would like to support that. I am thinking this can be a good follow-up, as we can also introduce some auto-scaling as we have it for the maps. in maps we would have to provide  a manual way to set the scale
* the axis is not a supported arguments. the reason is that is becomes incompatible with the plots as soon as you set facet_col, facet_row which returns multiple axes.  the idea is that you can alter the axis or axes object which you get as a return value. but what is confusing is, that there is no value error raised...
* aggregation to location in maps, this can be made possible after merging in #1182 and adding an additional argument. 
* supporting undefined carrier colors: carriers in ["", " - ", None] should now be working, however I would skip a color palette for other undefined carriers. we could think about a helper function `n.components.carrier.set_colors()` which you could call one time in advance. 
* We did not include any argument support in the `n.statistics.x.plot` function as arguments between `kind="bar"` and `kind="map"` are not shared. bringing that together, would be quite a hassle giving that both `n.statistics.x.plot.bar/map` support kwargs themselves.  
* pandas time stamp on x-axis. could look into it... however would not set a prio on that. 
* sorting of bar value.. the sorting comes from seaborn in most cases, it think is alphabetical (with some upper/lower case distinction it seems?)

FabianHofmann:
sorry clicked on delete with comment :(

PR #1120, opened 2025-01-15
Strict consistency checks by default

lkstrp:
- Adds option for granular strict consistency checks within `n.consistency_check()`
  - Pass it as a list, e.g. `n.consistency_check(strict=["unknown_buses"])`
- Runs the following checks as strict by default in `n.optimise`
  - `unknown_buses`

We could also add more strict checks in `n.optimize`. Let me know what you think

**Context**
> I think we should change the behavior in pypsa that if there is a link with undefined  buses that this throws either an error. I know that it throws a warning that buses of certain links are not defined, but this is often the output in prepare_sector_network because the bus is added later on.

## Checklist

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -47,9 +47,13 @@ Features
   :meth:`n.has_periods <pypsa.networks.Network.has_periods>` to simplified level access
   of the snapshots dimension. (https://github.com/PyPSA/PyPSA/pull/1113)
 
-* Consistency checks can now be run with the parameter ``strict``. If set to 
-  ``True``, the consistency check will raise an error if any of the checks fail.
-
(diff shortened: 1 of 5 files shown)


FabianHofmann:
:rocket:

p-glaum (commented):
Looks great, thanks for tackling this so fast! So if I understand correctly, how you implemented it means that only  "unknown_buses" is checked when calling n.optimize, correct? Because I think the other inconsistencies like "unknown_carriers" or "disconnected_buses" do not lead to wrong results in the optimisation.

lkstrp:
Yes, only `unknown_buses` by default right now. But we could add more, not sure if there is another test that might be worth checking strictly

FabianHofmann (approved):

PR #1131, opened 2025-01-28
Deprecate custom components

lkstrp:
As discussed @fneum 

- Since #1128 got way to big, I would publish a release to get those components classes out
- After 1128 custom components can be implemented way nicer then done before. Another question is if this is even needed. The API will change significantly in any case


fneum (approved):
Good to go once tests pass :)

PR #1136, opened 2025-02-05
Add update available note

lkstrp:
Adds a small note during package init if an update is available
``` python
>>> import pypsa
INFO:PyPSA:New version 0.32.1 available! (Current: 0.32.0)
```
## Checklist


fneum (approved):
No objection, but I have never seen a Python package do that on import. Will there be an option to silence this warning?

lkstrp:
True. I have a couple of CLI tools which do that and I think for the quite specific and scientific user base this could be helpful and solve a lot of problems up front (aka are you on the latest version).

But breaks a bit of consistency to other packages. Maybe network init instead of import is better? Not sure.

Option to silence the warning could be added to the new option config.

PR #1151, opened 2025-02-17
Fix pandas dtype warning in transformer calculations

tinegachris:
This PR addresses a FutureWarning related to setting an item of incompatible dtype in the pf.py file. The warning occurs when assigning values to the n.transformers DataFrame with incompatible data types.

`FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '[0 0 0 0 0]' has dtype incompatible with int32, please explicitly cast to a compatible dtype first.
  n.transformers.loc[trafos_with_types_b, attr] = t[attr]`

This fix ensures that the values being assigned have the same data type as the target DataFrame column by explicitly casting the values to the appropriate data type.


pypsa/pf.py
```diff
@@ -936,7 +936,9 @@ def apply_transformer_types(n: Network) -> None:
 
     # now set calculated values on live transformers
     for attr in ["r", "x", "g", "b", "phase_shift", "s_nom", "tap_side", "tap_ratio"]:
-        n.transformers.loc[trafos_with_types_b, attr] = t[attr]
+        n.transformers.loc[trafos_with_types_b, attr] = t[attr].astype(
+            n.transformers[attr].dtype
+        )
```
lkstrp, line 941:
```suggestion
    attrs = ["r", "x", "g", "b", "phase_shift", "s_nom", "tap_side", "tap_ratio"]
    n.transformers.loc[trafos_with_types_b, attrs] = t[attrs].astype(n.transformers[attrs].dtypes)
```
Could you remove the loop in one go?

lkstrp (approved):
Thanks for your contribution @tinegachris !

I am wondering why this is not catched by our global warning rule
https://github.com/PyPSA/PyPSA/blob/c97c88d976ee170a0d1a1e024d1eab060e83ec79/pyproject.toml#L95
Ref #1128

pypsa/pf.py
tinegachris: I did! Also wondering why it was not caught by the global warning rule.

fneum (approved):

PR #1037, opened 2024-09-17
Add stochastic programming formulation

Irieo:
# Implement stochastic programming formulation in PyPSA

This is **a draft**. 

This PR will:
-- add an option to cast a two-stage stochastic programming formulation with pypsa out-of-the-box.
-- add related documentation
-- add related notebook examples

This PR will close https://github.com/PyPSA/PyPSA/issues/575
Eventually, it should be possible to replicate this stochastic problem out-of-the-box with a few lines of code: https://pypsa.readthedocs.io/en/latest/examples/stochastic-problem.html

## Update

there is an ongoing restructuring process in PyPSA with the introduction of the component class (https://github.com/PyPSA/PyPSA/pull/1075) and a component xarray view (https://github.com/PyPSA/PyPSA/pull/1076) and the coming update of the optimization module (all optimization parameters will be handled as xarray objects instead of dataframes. PR to be created soon). 
This pipeline, once complete, will simplify the handling of stochastic programming in PyPSA by eliminating the need for case-specific equation handling, the StochasticNetwork class, and update of all relevant descriptors, as this PR suggests. 
We complete the three PRs mentioned above and revise this PR accordingly.

## work log

- [x] introduce a StochasticNetwork class
- [x] reindex snapshots with scenario names (SP: branches)
- [x] update snapshot_weightings with scenario probabilities
- [x] reindex time series parameters and variables (!) with scenario names (SP: branches)
- [x] introduce scenario-dependent static components.
- [x] add get_switchable_by_scenario in descriptors.py
- [x] add get_switchable_as_dense_by_scenario in descriptors.py
- [x]  add scenario dim to operational variables (variables.py) 
- [x] broadcast static data w/o _s components
- [x] make sure `__repr__` and `iterate_components` are compatible
- [x] fancy printout
- [x] --------------------------------------------------------------------------------
- [x] move new descriptors into StochNetwork class
- [x] make sure `get_switchable_as_dense` is compatible for varying and static cases in SP
- [x] make sure  `get_bounds_pu` is compatible with SP
- [x] make sure get_bounds_pu returns right xarray object for equations to work out-of-the-box
- [x] make n.df(c) to be multi-index dataframe instead of a dict. Make everything above compatible.
- [x] fix specific case of StorageUnit - state_of_charge - in get_bounds_pu for SP
- [x] rework get_switchable_by_scenario for multi-index n.df(c)
- [x] bring back explicit "scenario" dim to operational variables
- [x] add explicit scenario dim to operational constraints
[...]

PR #1076, opened 2024-11-07
feat: add component data `xarray` view

lkstrp:
Adding up on #1075 and in preparation for #1037

## Changes proposed in this Pull Request
- Add `xarray` view of component data via `c.ds`.
  - Should be treated as immutable and probably only internally. Better setter checks and caching will be added. 
- Merge static and dynamic data
- Handles `timestep`, `period` and `scenario` dimensions
- Removes arbitrary naming of snapshot index (e.g. `snapshot` if not multi horizon, `['timestep', 'period']`) 
  - Rename `snapshot` -> `timestep
  - If people are somehow working with index names, this will break

## Todos
- [ ] Caching
- [ ] Immutable checks

### Example 
Create network with investment periods
``` python
n = pypsa.examples.ac_dc_meshed()
n.snapshots = pd.MultiIndex.from_product([[2013], n.snapshots])
n.investment_periods = [2013]
gens_i = n.generators.index
rng = np.random.default_rng()
n.generators_t.p[gens_i] = rng.random(size=(len(n.snapshots), len(gens_i)))
```
then you can run:
``` python
>>> n.components.generators.ds
<xarray.Dataset> Size: 10kB
Dimensions:                  (Generator: 6, timestep: 10, period: 1)
Coordinates:
  * Generator                (Generator) object 48B 'Manchester Wind' ... 'Fr...
  * timestep                 (timestep) datetime64[ns] 80B 2015-01-01 ... 201...
  * period                   (period) int64 8B 2013
Data variables: (12/45)
    bus                      (Generator) object 48B 'Manchester' ... 'Frankfurt'
    control                  (Generator) object 48B 'PQ' 'PQ' 'PQ' ... 'PQ' 'PQ'
    type                     (Generator) object 48B '' '' '' '' '' ''
    p_nom                    (Generator) float64 48B 80.0 5e+04 ... 110.0 8e+04
    p_nom_mod                (Generator) float64 48B 0.0 0.0 0.0 0.0 0.0 0.0
    p_nom_extendable         (Generator) bool 6B True True True True True True
    ...                       ...
    status                   (Generator, timestep, period) float64 480B nan ....
    mu_upper                 (Generator, timestep, period) float64 480B nan ....
    mu_lower                 (Generator, timestep, period) float64 480B nan ....
    mu_p_set                 (Generator, timestep, period) float64 480B nan ....
    mu_ramp_limit_up         (Generator, timestep, period) float64 480B nan ....
[...]

PR #1154, opened 2025-02-25
New Optimization Module & Stochastic Networks 🔨💥

### LINKED ISSUE #575: Stochastic programming formulation
We have been iteratively coming back to discussion that Stochastic Programming (SP) is one of the most wished features. 
There is an [example notebook](https://discord.com/channels/911692131440148490/1047183573028962355) with SP in PyPSA; however, that formulation is badly scalable. A stochastic investment planning feature in PyPSA itself would likely find many users. 

Below is the a quick sketch how to implement SP with Linopy integration module in PyPSA. This would allow for modelling two-stage SP with investment decisions in the first stage, and operational decision in the second stage. There is a ton of energy-related literature with SP implementation, a really good example is [here](https://econpapers.repec.org/article/eeeeneeco/v_3a34_3ay_3a2012_3ai_3a6_3ap_3a2089-2101.htm).

Eventually, a user should be able to define (i) list of scenarios to model, (ii) realization probabilities of defined scenarios, and (iii) a vector (or a matrix) of uncertain parameter(s). The latter could be a cost coefficient or a rhs parameter (like capacity constraint). 

1. First, we add functionality to `pypsa/optimization/optimize.py`:

- `create_model()` receives new argument `stochastic` (boolean). 
- If `False`, the function stays as is. 
- If `True`: 
  - all operational variables get new coordinate `scen`
  - all constraints get new coordinate  `scen`
  - the global constraints likely stay same 
  
2. `define_objective()` inherits argument stochastic. If `True`, the changes in function are so that line `objective.append((operation * cost).sum())` can eventually be changed to `objective.append((operation[scenario] * cost[scenario]* probability[scenario] )).sum())` The  `investment` part stays same.

3. `post_processing()` inherits argument stochastic. If `True`, the  `p0` and `p1` are defined per `scen`
4. `assign_duals()` inherits argument stochastic. if `True`,  the duals are scaled back by 1/probability.

5. User has to pass three bits of information to model (via **kwargs ?)
- a list of scenario names -> this is read by `create_model() ` to populate `scen`
- one vector (like pandas series) of scenario probabilities (or dictionary with scenario: probability pairs) -> this is passed to `define_objective()`
- a vector (e.g., CO2 prices) or matrix (e.g., time-series) of uncertain parameter
- Other parameters that are NOT changed by user-defined vector/matrix are populated with the same values across scenarios.

A more general form covering e.g. unit-commitment decisions under RES feed-in uncertainty is harder to generalise via `pypsa/optimization/optimize.py`; however, structurally the problem would be of the same form.

### LINKED ISSUE #848: Remove "-ext" and "-fix" suffix in index names in index getter functions

Irieo:
## 🔧 New Optimization Module

Closes #575 
Closes #814 
Closes #848 
Closes #1037

Included PRs:
- #1076
- #1164 
- #1166
- #1229

## Summary
- Implement xarray view via `c.da.<attr>` to get merged view of `c.static` and `c.dynamic` across all dimensions (`name`, `snapshot` (with optional coordinate `period`), `scenario`)
  - Refactor of `optimization` module to make use of xarray operations for all mathematical expressions
- New feature by introducing stochastic problems into PyPSA
  - `c.static` and `c.dynamic` are casted to include the `scenario` dimension (pd.MultiIndex on either index or columns)
  - Make all PyPSA functionality either supported (`plotting`, `statistics`, `io` ..) or not implemented (`clustering`)
- Other
  - Adds `guards.py` with a `pypsa.option` for runtime verification during tests
  - Make `NetworkCollection` work with nested attributes to mirror behavior in `n.statistics` for collections/ stochastic networks (both just add dimensions)
  - Index and weightings data is stored in common attribute (e.g. `self._snapshots_data`), but exposed as usual
- Breaking
  - Index names of `pd.DataFrames` have been changed from `<component_name>` (`Link` etc.) to `name` to align with xarray view
  - Remove `-ext`, `-fix` suffix and simplify their helper functions

## 🔀 New feature: two-stage stochastic optimization

PyPSA now supports stochastic networks for two-stage optimization under uncertainty. This enables modeling scenarios with different probabilities for parameters like fuel prices, demand, or renewable availability.

### Mathematical Formulation

This PR introduces a two-stage stochastic programming formulation as an optional optimization problem in PyPSA:

$$
\begin{align}
\min_{x} \quad & c^T x + \sum_{s \in S} p_s Q(x, \xi_s) \\
\text{s.t.} \quad & A x = b \\
& x \geq 0
\end{align}
$$

Where:

- $x$: First-stage (here-and-now) decision variables (e.g., investment decisions in generation/transmission/storage capacities)
- $\xi_s$: Random parameter realizations in scenario $s$
- $p_s$: Probability of scenario $s$ occurring
- $Q(x, \xi_s)$: Second-stage recourse function for scenario $s$
- $S$: Set of all scenarios

The second-stage problem $Q(x, \xi_s)$ represents:

$$
\begin{align}
Q(x, \xi_s) = \min_{y_s} \quad & q_s^T y_s \\
\text{s.t.} \quad & W y_s = h_s - T x \\
& y_s \geq 0
\end{align}
$$

Where:

- $y_s$: Second-stage (wait-and-see) decision variables for scenario $s$ (e.g., energy dispatch decisions)
- $q_s$: Second-stage cost coefficients for scenario $s$
- $W$: Recourse matrix (coefficient matrix for second-stage variables $y_s$)
- $T$: Technology matrix (coefficient matrix for first-stage variables $x$ in second-stage constraints)
- $h_s$: Right-hand-side vector for scenario $s$

The second-stage constraints $W y_s = h_s - T x$ show how first-stage decisions $x$ affect the feasible region for second-stage variables $y_s$ in each scenario.

### Setting Scenarios
Use `n.set_scenarios()` to convert a regular network into a stochastic network:

```python
# Multiple scenarios with probabilities
n.set_scenarios({"low": 0.4, "medium": 0.3, "high": 0.3})

# If you don't specify probabilities, they default to uniform distribution
n.set_scenarios(["low", "medium", "high"])
```

### Scenario Access
```python
# Check if network has scenarios
if n.has_scenarios:
    print(f"Network has {len(n.scenarios)} scenarios")

# Access scenario probabilities
print(n.scenarios)
# Returns: pd.Series with scenario names as index and probabilities as values
```

### String Representation
Stochastic networks are identified in their string representation:
```python
print(n)
# Output: "Stochastic PyPSA Network 'NetworkName'"
# vs regular: "PyPSA Network 'NetworkName'"
# Includes: "Scenarios: 3" at the bottom
```

### Scenario-Specific Parameters
All component data is automatically expanded across scenarios using MultiIndex.
You can set scenario-specific parameters like this:

```python
# Set different marginal costs per scenario
n.generators.loc[("high", "gas"), "marginal_cost"] = 100
n.generators.loc[("medium", "gas"), "marginal_cost"] = 70
n.generators.loc[("low", "gas"), "marginal_cost"] = 40

# Set scenario-specific time series
n.loads_t.p_set.loc[:, ("high", "load1")] = [120, 144, 132]
n.loads_t.p_set.loc[:, ("low", "load1")] = [80, 96, 88]
```

## Optimization

Stochastic networks are solved using the same interface as regular networks:

```python
# Standard optimization
n.optimize()

# Standard optimization with solver options
status, condition = n.optimize(solver_name="gurobi", solver_options={"MIPGap": 0.01})
```

### Solution Structure
The primal and dual variables are written back to the network object in the same way as for 
regular networks, but with an additional scenario dimension. 

Results maintain the scenario structure:

```python
# Generator dispatch: (scenario, name) columns
n.generators_t.p.columns.names  # ['scenario', 'name']

# Access specific scenario results
high_dispatch = n.generators_t.p.loc[:, ("high", slice(None))]
low_dispatch = n.generators_t.p.loc[:, ("low", slice(None))]

# Optimal capacities (investment decisions are scenario-independent)
n.generators.p_nom_opt.loc[("high", "gen1")]  # Same for all scenarios
```

### Statistics
Statistics methods work seamlessly with stochastic networks:

```python
# Statistics include scenario dimension
n.statistics()
```

## Backward Compatibility
- Regular (non-stochastic) networks continue to work unchanged
- No breaking changes to existing code

##  How do problems scale as a function of scenario count in stochastic optimization?

See detail about a setup of this test case here: https://github.com/PyPSA/PyPSA/pull/1154#issuecomment-3145005302

```markdown
# Same data across scenarios
Benchmark Results:
===============================================================================================
Batch size: 5 runs, values reported: mean (min, max), time in seconds, memory in MB
Scen     Total Time        Build Time         Solver Time       Memory             xTime     xMemory
-----------------------------------------------------------------------------------------------
   0    7.1 (6.3, 7.7)     6.6 (5.9, 7.2)     0.4 (0.4, 0.5)     9.1 (8.9, 9.5)     1.0x     1.0x
   1    8.4 (8.0, 9.0)     8.0 (7.7, 8.6)     0.3 (0.3, 0.4)     9.7 (9.4, 9.9)     1.2x     1.1x
   2    9.1 (8.9, 9.4)     8.8 (8.6, 9.1)     0.3 (0.3, 0.3)     17.8 (17.6, 17.9)  1.3x     2.0x
   3   10.4 (10.2, 10.7)    10.1 (9.9, 10.4)   0.3 (0.3, 0.3)    25.6 (25.4, 25.8)  1.5x     2.8x
   5   13.4 (12.7, 15.1)    13.0 (12.4, 14.5)  0.4 (0.4, 0.5)    41.8 (41.5, 42.0)  1.9x     4.6x
   7   18.2 (18.0, 18.7)    17.7 (17.4, 18.1)  0.6 (0.5, 0.6)    63.8 (63.5, 63.9)  2.6x     7.0x
  11   20.8 (20.3, 21.3)    20.2 (19.7, 20.5)  0.7 (0.6, 0.7)    90.0 (89.8, 90.2)  3.0x     9.9x

Objective tolerance check (data in not modified across scenarios):
All scenarios: ✓ within solution tol

# Noise added to demand loads
Benchmark Results:
===============================================================================================
Batch size: 5 runs, values reported: mean (min, max), time in seconds, memory in MB | 20% load noise
Scen     Total Time        Build Time         Solver Time       Memory             xTime     xMemory
-----------------------------------------------------------------------------------------------
   0    6.9 (6.5, 7.3)     6.5 (6.1, 6.8)      0.4 (0.4, 0.5)     9.1 (8.9, 9.5)     1.0x     1.0x
   1    8.8 (8.5, 9.2)     8.4 (8.1, 8.7)      0.4 (0.4, 0.5)     9.7 (9.4, 9.9)     1.3x     1.1x
   2   10.9 (10.4, 11.3)    10.3 (9.9, 10.6)   0.6 (0.5, 0.7)     17.8 (17.6, 17.9)  1.6x     2.0x
   3   12.0 (11.7, 12.2)    11.1 (10.8, 11.3)  0.9 (0.8, 0.9)     25.6 (25.4, 25.8)  1.7x     2.8x
   5   15.3 (15.0, 15.8)    13.7 (13.5, 14.1)  1.5 (1.4, 1.8)     41.8 (41.6, 42.0)  2.2x     4.6x
   7   20.9 (20.2, 21.7)    18.5 (17.9, 19.1)  2.4 (2.3, 2.6)     63.7 (63.5, 63.9)  3.0x     7.0x
  11   26.0 (25.3, 26.8)    21.7 (21.3, 21.9)  4.3 (3.9, 4.9)     89.9 (89.7, 90.1)  3.8x     9.9x
```

## For More Information

For detailed documentation on stochastic programming theory, advanced examples, and complete API reference, see:

- [PyPSA user guide on stochastic networks](https://pypsa--1250.org.readthedocs.build/en/1250/user-guide/stochastic-optimization/)
- [PyPSA examples on stochastic optimization](https://pypsa--1250.org.readthedocs.build/en/1250/examples/stochastic-optimization/)

## ToDos

In this PR:

- [x] add line in docstrings for which components each constraint applies
- [x] Apply extended ruff rules
- [x] `base_attr` refactor
- [x] #848 
- [x] Agree on API. Currently we have the following:
    - [x] Index Names: ~`component`, `snapshot`, `scenario`~ (vs `name`, `snapshot`, `scenario`)
    - [x] Data Array Access: `c.da.p_nom` vs ~(`c.as_array('p_nom')`)~
    - [x] `_operational_attrs` not exposed
- [x] Resolve all `#todo`
- [x] Test all functionality on stochastic Networks
  - Either make it work or raise correct Errors
- [x] Profiling Non-Stochastic workflows
- [x] align n.scenarios
- [x] fix failing unit tests
- [x] fix stoch networks causing troubles with "Store" | specific config
- [x] fix stoch network issue of writing solution back to network with misaligned scenario labels\
- [x] add unit tests for the edgy cases with stoch networks
- [x] all FN feedback
- [x] all global constraints support stochastic opt (or not implemented flag) 

IR 
- [x] ensure compatibility with assign-all-duals=True
- [x] Remove inspect for different p_set defaults
- [x] solve SCLOPF fail issue
- [x] line/trafo types broadcasting (PyPSA-Eur workflow break)
- [x] multi-links and nodal_balance issue for large stoch network from pypsa-eur
- [x] multiperiod+storage bug
- [x] _extract_names_for_multiindex problem -> and growth limit cons
- [x] add tests for operational limit cons

BX/IR
- [x] Test large problems solved with new backend & stochastic networks

LT 
- [x] `NotImplemented` notes everywhere
- [x] ~Back to pd.Series~
- [x] Profile and fix excess memory usage

In future PRs:
Docs
- [ ] polish docs/examples for release
- [ ] Docs on Breaking Changes (in `v1-docs` branch)

Plotting
- [ ] Plots: unique carriers
- [ ] Plots: merge plots for collections and tolerance for stochastic networks

Other
- [ ] Add slicing to get either collection, scenario or previous behaviour
- [ ] #1056
- [ ] Apply v1 API
- [ ] Profile/dashboard large stochastic problem memory and time 
- [ ] risk aversion


FabianHofmann:
@lkstrp pypsa-eur is working now. should we also apply the fix to pypsa-de?

lkstrp:
That will probably be resolved with the next merge of PyPSA-Eur in PyPSA-DE. @lindnemi would you have time to do that?

lindnemi:
sure, i will merge upstream within a couple of days

pypsa/collection.py
```diff
@@ -327,6 +308,21 @@ def _validate_network_compatibility(self) -> None:
...
+    r"sub_networks|buses|carriers|global_constraints|lines|line_types|"
+    r"transformers|transformer_types|links|loads|generators|storage_units|"
+    r"stores|shunt_impedances|shapes"
+)
+
+_all_component_names = (
+    r"SubNetwork|Bus|Carrier|GlobalConstraint|Line|LineType|"
+    "Transformer|TransformerType|Link|Load|Generator|StorageUnit|"
+    "Store|ShuntImpedance|Shape"
+)
```
fneum, line 321: Low prio, but isn't there a way to get this from somewhere rather than retyping?

Irieo:
@FabianHofmann @lkstrp a context for  #https://github.com/PyPSA/PyPSA/pull/1154/commits/2bedd523e651ed8ba64e1fe8d50e24ea92b8efc4

The bus `control` attribute (PQ/PV/Slack) is derived from connected generators during topology determination, not set by users. So I removed `control` from the consistency check for scenario invariant attributes.  Instead I patched `power_flow.py` -> `find_slack_bus()` to select the same slack bus for each scenarios in  stochastic networks; and add here a dedicated consistency check that looks whether the same bus is chosen as slack across all scenarios. And plus unit tests for these.

Irieo:
@fneum thanks for your review of stochastic network functionality. Quick recap:
- The bug with `Store` component is fixed (problem was `elapsed_hours` broadcasting with MultiIndex instance of stoch network) https://github.com/PyPSA/PyPSA/pull/1154/commits/2378a72d1d7fc7cce2ef94e40391afd593bb4c69
- The bug with (occasionally) wrong labeling of results to scenario dimensions is also fixed (problem was actually before `assign_solution`, instead async order of scenarios in linopy model itself) https://github.com/PyPSA/PyPSA/pull/1154/commits/bfc4a5ca33a4aa06ca35b3f37c22474924bafeca
- Multiperiod investments and stochastic optimization are compatible out-of-box after a small touch in storage unit constraints. And plenty of dedicated tests. https://github.com/PyPSA/PyPSA/pull/1154/commits/b2c1d14f140c3d159e2efd67efe1d87486b43af8

Irieo:
Fun observation: when looking through the `power_flow.py` I realised I can simply remove inspect check from array module w/o any changes to functionality because `pf` code already has treatment for a starting point for the algorithm
```
# power flow calculations require a starting point for the algorithm, while p_set default is n/a
c_p_set = c_p_set.fillna(0)
n.dynamic(c.name).p.loc[sns, c.static.query("active").index] = c_p_set
```
I looked to git blame trace and realised it was *me* who fixed it already 4 weeks ago in a commit dedicated to this specific problem https://github.com/PyPSA/PyPSA/commit/79a52a9f70f5fb4a0874211092ea212670892a3d within new-opt-stoch branch that we merged here. I totally forgot.

another sign that it is time to finish this.

bobbyxng:
Agreed with @Irieo to run some tests comparing old large-scale runs with the same `.nc` file using `new-opt`
**Test case** setup
- Using 2030 network allowing for H2 and CO2 grid expansion, NUTS0-2 resolution, 2190seg temporal resolution
- see https://github.com/bobbyxng/pci-pmi-policy-targets
- Includes custom constraints from PyPSA-Eur `solve_network.py`

**Four cases:**
- no0 (reference): Old network solved using PyPSA master branch
- no1: Same network, updated to newest PyPSA new-opt branch version, no scenarios
- no2: Same network, single scenario with probability 1
- no3: Same network, two scenario with arbitrarily assigned probabilities 0.3 and 0.7

Current issues:
- no1 solves fine, I am in the process of comparing results
- no2 and no3: 
  - In the mapping process for mapping lines, PyPSA is looking for line types which are also indexed by scenario. So for example `'Al/St 240/40 4-bundle 380.0'` is not found as it is now listed as `('scenario-1','Al/St 240/40 4-bundle 380.0')`
  - One could now either revert the linetypes to how they were before or adjust the workflow to handle this updated storing of line types. But option 1 makes more sense imo, as these are a static list of line types.

Irieo:
> Agreed with @Irieo to run some tests comparing old large-scale runs with the same `.nc` file using `new-opt` **Test case** setup
> ......
>   * In the mapping process for mapping lines, PyPSA is looking for line types which are also indexed by scenario. So for example `'Al/St 240/40 4-bundle 380.0'` is not found as it is now listed as `('scenario-1','Al/St 240/40 4-bundle 380.0')`

thanks @bobbyxng for taking your time to setup these tests.
Line types should not be broadcasted. Fixed in aaa701884d822d4623e1e84c177af8f686881c4c

bobbyxng:
For the test case, n0 (reference, PyPSA master) and n1 (new-opt branch), results are almost identical. Differences are likely attributed to different noisy_costs (random):
- Objective value no0: 525237380658.93
- Objective value no1: 524712116860.01
- Objective value delta of 0.1 %
- CO2 price no0: 74.85060162578911
- CO2 price no1: 74.84001674352074

<details>
<summary>Click to expand!</summary>
### Cost by carrier (no0 left, no1 right)



### AC balance by carrier (no0 left, no1 right)



### Total energy balance by carrier (no0 left, no1 right)



### Annual CO2 balance (no0 left, no1 right)



## Maps
### Regional CO2 balance (no0 left, no1 right)



### Regional H2 balance (no0 left, no1 right)


- Small discepancies in pipeline investments, see Germany from north to south direction

</details>

lkstrp:
image

Memory profile for https://github.com/PyPSA/PyPSA/pull/1154/commits/9d93c71334b5622af951867e5db92c81a6c5ec1e of solve_network in PyPSA Eur

bobbyxng:
@lkstrp Nice! Here's the memory log for the large-scale PyPSA-Eur example, no0 reference vs. no1 new-opt run
image

fneum:
What's with the longer solving times? Is that consistent? Did the final LP form change?

fneum:
Noisy costs should be turned off for verification.

Irieo:
> What's with the longer solving times? Is that consistent? Did the final LP form change?

This looks inconsistent. A test by @lkstrp shows new-opt faster, a test by @bobbyxng shows otherwise.

Irieo:
So, how PyPSA problems scale as a function of the number of scenarios in stochastic optimization?

Setup:
- `pypsa.examples.storage_hvdc()` network (I extended it to 168 snapshots to increase temporal complexity)
- Gurobi with pypsa-eur's default options (barrier method, no crossover)
- Each case is run 5 times
- Solver time is taken from `n.model.solver_model.Runtime`
- Linopy build time as diff between total run time and Gurobi runtime

Solver time may depend on the complexity of first-stage decisions. To test this:
1) I first solve the problem without modifying data across scenarios;
2) then solve with ±20\% random noise added to the demand loads. Since all components are expandable 
in this problem, this noise should add the trade-off complexity in first-stage decisions.

Summary:
- Build time scales linearly with scenarios
- Solver time grows exponentially with scenario count when input varies  (due to trade-offs in here-and-now investment decisions)
- Memory usage scales linearly with scenario count
- Overall things are predictable / well-behaved
- With **5 scenarios** and varying scenario data: **2.2xTime** and **4.6xMemory** compared to deterministic problem

```markdown
# Same data across scenarios
Benchmark Results:
===============================================================================================
Batch size: 5 runs, values reported: mean (min, max), time in seconds, memory in MB
Scen     Total Time        Build Time         Solver Time       Memory             xTime     xMemory
-----------------------------------------------------------------------------------------------
   0    7.1 (6.3, 7.7)     6.6 (5.9, 7.2)     0.4 (0.4, 0.5)     9.1 (8.9, 9.5)     1.0x     1.0x
   1    8.4 (8.0, 9.0)     8.0 (7.7, 8.6)     0.3 (0.3, 0.4)     9.7 (9.4, 9.9)     1.2x     1.1x
   2    9.1 (8.9, 9.4)     8.8 (8.6, 9.1)     0.3 (0.3, 0.3)     17.8 (17.6, 17.9)  1.3x     2.0x
   3   10.4 (10.2, 10.7)    10.1 (9.9, 10.4)   0.3 (0.3, 0.3)    25.6 (25.4, 25.8)  1.5x     2.8x
   5   13.4 (12.7, 15.1)    13.0 (12.4, 14.5)  0.4 (0.4, 0.5)    41.8 (41.5, 42.0)  1.9x     4.6x
   7   18.2 (18.0, 18.7)    17.7 (17.4, 18.1)  0.6 (0.5, 0.6)    63.8 (63.5, 63.9)  2.6x     7.0x
  11   20.8 (20.3, 21.3)    20.2 (19.7, 20.5)  0.7 (0.6, 0.7)    90.0 (89.8, 90.2)  3.0x     9.9x

Objective tolerance check (data in not modified across scenarios):
All scenarios: ✓ within solution tol

# Noise added to demand loads
Benchmark Results:
===============================================================================================
Batch size: 5 runs, values reported: mean (min, max), time in seconds, memory in MB | 20% load noise
Scen     Total Time        Build Time         Solver Time       Memory             xTime     xMemory
-----------------------------------------------------------------------------------------------
   0    6.9 (6.5, 7.3)     6.5 (6.1, 6.8)      0.4 (0.4, 0.5)     9.1 (8.9, 9.5)     1.0x     1.0x
   1    8.8 (8.5, 9.2)     8.4 (8.1, 8.7)      0.4 (0.4, 0.5)     9.7 (9.4, 9.9)     1.3x     1.1x
   2   10.9 (10.4, 11.3)    10.3 (9.9, 10.6)   0.6 (0.5, 0.7)     17.8 (17.6, 17.9)  1.6x     2.0x
   3   12.0 (11.7, 12.2)    11.1 (10.8, 11.3)  0.9 (0.8, 0.9)     25.6 (25.4, 25.8)  1.7x     2.8x
   5   15.3 (15.0, 15.8)    13.7 (13.5, 14.1)  1.5 (1.4, 1.8)     41.8 (41.6, 42.0)  2.2x     4.6x
   7   20.9 (20.2, 21.7)    18.5 (17.9, 19.1)  2.4 (2.3, 2.6)     63.7 (63.5, 63.9)  3.0x     7.0x
  11   26.0 (25.3, 26.8)    21.7 (21.3, 21.9)  4.3 (3.9, 4.9)     89.9 (89.7, 90.1)  3.8x     9.9x
```

bobbyxng:
Reverted `line_types_to_use` in `power_flow.py` to @Irieo's version
``` 
line_types_to_use = n.line_types.xs(
            n.line_types.index.get_level_values(0)[0], level=0
        )
``` 

Refactored code introduced with 32aa387 came with breaking changes in networks with scenarios, as this yields an index series, which `df.xs()` cannot use.
``` 
line_types_to_use = n.line_types.xs(n.c.line_types.component_names, level=0)
```

bobbyxng:
Here's an updated plot of profiling a large-scale test case, see https://github.com/PyPSA/PyPSA/pull/1154#issuecomment-3077749897 (noisy costs off), 

image

pypsa/components/_types/storage_units.py
```diff
@@ -28,6 +30,43 @@ class StorageUnits(Components):
...
+            raise ValueError(msg)
+
+        max_pu = self.da.p_max_pu
+
+        if attr == "p_store":
+            max_pu = -self.da.p_min_pu
+            min_pu = xr.zeros_like(max_pu)
+        elif attr == "state_of_charge":
+            max_pu = self.da.max_hours
+            min_pu = xr.zeros_like(max_pu)
```
fneum, line 63: Interesting, quite unconventional, but correct as far as I can see.

pypsa/consistency.py
```diff
@@ -63,7 +67,7 @@ def check_for_unknown_buses(
 
     """
     for attr in _bus_columns(component.static):
-        missing = ~component.static[attr].astype(str).isin(n.buses.index)
+        missing = ~component.static[attr].astype(str).isin(n.c.buses.component_names)
```
fneum, line 68:
```suggestion
        missing = ~component.static[attr].astype(str).isin(n.c.buses.index)
```
Why can this not just be `index` instead of a new attribute `component_names`?

pypsa/network/graph.py
```diff
@@ -113,79 +120,100 @@ def adjacency_matrix(
...
-        no_branches = 0
-        bus0_inds = []
-        bus1_inds = []
-        weight_vals = []
-        for c in self.iterate_components(branch_components):
-            active = c.get_active_assets(investment_period)
-            sel = c.static[active].index
+        # Initialize empty DataFrame with buses as both rows and columns
+        if n.has_scenarios:
+            busorder = busorder.unique("name")
```
fneum, line 170: Do you know if `.unique()` preserves order? Possibly it does not, in which case you need `busorder.get_level_values("name").drop_duplicates()`.

pypsa/network/index.py
```diff
@@ -642,7 +637,119 @@ def investment_period_weightings(self, df: pd.DataFrame) -> None:
...
                 "Applying weightings to all columns of `investment_period_weightings`"
             )
-            df = pd.DataFrame(
-                dict.fromkeys(self._investment_period_weightings.columns, df)
+            df = pd.DataFrame(dict.fromkeys(self._investment_periods_data.columns, df))
+        self._investment_periods_data = df
+
+    # -----------
+    # Scenarios
+    # -----------
```
fneum, line 649:
```suggestion
```

pypsa/network/io.py
```diff
@@ -322,6 +326,11 @@ def save_investment_periods(self, investment_periods: pd.Index) -> None:
                 fn, encoding=self.encoding, quotechar=self.quotechar
             )
 
+    def save_scenarios(self, scenarios: pd.DataFrame) -> None:
+        """Save scenarios data."""
+        msg = "Stochastic networks are not supported in the CSV exporter. Use netcdf instead."
+        raise NotImplementedError(msg)
```
fneum, line 332: Any particular problem that resulted in the decision not to support CSV or Excel exports for stochastic networks? Or just to save time and it will come as follow-up?

pypsa/network/io.py
```diff
@@ -848,21 +875,40 @@ def get_static(self, list_name: str, index_name: str | None = None) -> pd.DataFr
...
-                df.columns.name = "name"
+                try:
+                    df = self.ds[attr].to_pandas()
+                    # df.index.name = "name"
+                    df.columns.name = "name"
+                # Handle multi-indexed (scenarios)
+                except ValueError:
+                    df = (
+                        self.ds[attr]
+                        .stack(combined=("scenario", attr + "_i"))
```
fneum, line 907: What does the suffix `_i` stand for?

pypsa/network/io.py
```diff
@@ -1061,20 +1120,21 @@ def _export_to_exporter(
 
         exporter.save_meta(self.meta)
 
-        # now export snapshots
-        if isinstance(self.snapshot_weightings.index, pd.MultiIndex):
-            self.snapshot_weightings.index.rename(["period", "timestep"], inplace=True)
-        else:
-            self.snapshot_weightings.index.rename("snapshot", inplace=True)
```
fneum, line 1068: I guess the renaming is no longer necessary because it happens elsewhere already?

pypsa/network/io.py
```diff
@@ -1087,19 +1147,22 @@ def _export_to_exporter(
...
 
             col_export = []
             for col in static.columns:
-                # do not export derived attributes
-                if col in ["sub_network", "r_pu", "x_pu", "g_pu", "b_pu"]:
+                # do not export derived attributes and object column of subnetwork
+                if col in ["g_pu", "b_pu"]:
                     continue
+                if col == "obj":
+                    static[col] = np.nan
```
fneum: Is it really no problem to delete the `obj` data when exporting?

I.e. when I want to continue with `n` after exporting I would have to rerun `n.determine_network_topology()`?

pypsa/network/power_flow.py
```diff
@@ -1306,7 +1336,7 @@ def pf(
         if not skip_pre:
             n.calculate_dependent_values()
             self.find_bus_controls()
-            _allocate_pf_outputs(n, linear=False)
```
fneum: Why the switch from `linear=False` to `linear=True`?

test/test_z_docs.py
fneum, line 1: Renamed on purpose?

test/test_stochastic.py
```diff
@@ -0,0 +1,1463 @@
...
+
+
+def test_optimization_advanced(storage_hvdc_network):
+    """
+    Advanced test case for the optimization of a stochastic network.
+    """
+    n = storage_hvdc_network
+    n.set_scenarios({"low": 0.5, "high": 0.5})
+    status, _ = n.optimize(solver_name="highs")
+    assert status == "ok"
```
fneum: Should this test also verify that the objective is equivalent to that of a non-stochastic network?

test/test_stochastic.py
```diff
@@ -0,0 +1,1463 @@
...
+    The bug was:
+    - `expand_series` returns a DataFrame with shape (snapshots, scenarios)
+    - `standing_loss` is a DataArray with shape (snapshots, scenarios, stores)
+    - The power operation (1 - standing_loss) ** eh failed with 3D vs 2D mismatch
+
+    The fix:
+    - Convert expand_series result to DataArray and try to unstack it
+    - This aligns the dimensions properly for the power operation
+    """
+    n = pypsa.examples.model_energy()
```
fneum, line 585: But this network has no non-zero standing losses in the stores. So, is this testing the right thing?

test/test_lopf_unit_commitment.py
```diff
@@ -505,3 +505,41 @@ def test_dynamic_ramp_rates():
...
+        ramp_limit_up=static_ramp_up * 1,
+        ramp_limit_down=static_ramp_down * 1,
+        marginal_cost=gen1_marginal,
+    )
+
+    n.add("Generator", "gen2", bus="bus", p_nom=100, marginal_cost=150)
+
+    status, _ = n.optimize(snapshots=n.snapshots)
+
+    assert status == "ok"
```
fneum, line 545: Perhaps it should assert that ramp limits are met rather than just a successful run?

pypsa/optimization/common.py
```diff
@@ -15,15 +16,20 @@
     from pypsa import Network
 
 
+@deprecated_in_next_major(
+    details="Use xarray functionality instead.",
```
fneum: Maybe give a hint how to do this?

pypsa/networks.py
```diff
@@ -858,6 +882,17 @@ def __getitem__(self, key: str) -> Network:
...
+    def _empty_components(self) -> list:
+        """Get a list of all components that are not empty.
+
+        Returns
+        -------
+        list
+            List of non-empty components.
+
+        """
+        return [c.name for c in self.iterate_components() if c.empty]
```
fneum:
```suggestion
    def _empty_components(self) -> list:
        """Get a list of all components that are empty.

        Returns
        -------
        list
            List of empty components.

        """
        return [c.name for c in self.iterate_components() if c.empty]
```

pypsa/networks.py
```diff
@@ -1036,6 +1119,84 @@ def determine_network_topology(
 
         return self
 
+    def cycles(
+        self, investment_period: str | int | None = None, apply_weights: bool = False
+    ) -> pd.DataFrame:
+        """Get the cycles in the network and represent them as a DataFrame.
+
+        This function identifies all cycles in the network topology and
```
fneum:
```suggestion
        This function identifies a cycle basis of the network topology and
```
Careful, actually, this function (if it's the same as before) identifies a cycle basis for each network (which is less than all cycles as some might be linearly dependent). This means fewer KVL constraints (just as many as necessary to determine the flows). To be accurate, this function should be called `cycle_basis()` rather than `cycles`.

pypsa/networks.py
```diff
@@ -1036,6 +1119,84 @@ def determine_network_topology(
...
+        of the cycle.
+
+        Parameters
+        ----------
+        investment_period : str or int, optional
+            Investment period to use when determining network topology.
+            If not given, all branches are considered regardless of
+            build_year and lifetime.
+        apply_weights : bool, default False
+            Whether to apply weights to the cycles.
```
fneum: Perhaps mention here that weights are reactance for AC and resistance for DC lines.

pypsa/networks.py
```diff
@@ -1036,6 +1119,84 @@ def determine_network_topology(
...
+        branches_i = branches.loc[existing_branch_components].index
+
+        if apply_weights:
+            is_ac = branches.sub_network.map(self.sub_networks.carrier) == "AC"
+            weights = branches.x_pu_eff.where(is_ac, branches.r_pu_eff)
+            weights = weights[cycles_df.index]
+            cycles_df = cycles_df.multiply(weights, axis=0)
+
+        # Reindex to include all branches (even those not in cycles)
+        return cycles_df.reindex(branches_i, fill_value=0).rename_axis(columns="cycle")
```
fneum, line 1200: Not exclusively related to code here, but generally, how are 0 or NaN values handled in LP file creation? Are they skipped when written to LP or written with coefficient 0 (which involves more variables in constraints as necessary)?

pypsa/optimization/variables.py
```diff
@@ -16,134 +16,222 @@
...
+    upper = c.da.inflow.sel(snapshot=sns)
     if (upper.max() <= 0).all():
         return
 
-    active = get_activity_mask(n, c, sns).where(upper > 0, False)
-    n.model.add_variables(0, upper, name="StorageUnit-spill", mask=active)
+    active = c.da.active.sel(snapshot=sns)
+
+    # align "active" and "upper" arrays on the same order across scenario/snapshot/component axes
+    # .align() TODO low high
```
fneum: This still a relevant TODO?

pypsa/optimization/global_constraints.py
```diff
@@ -257,43 +282,70 @@ def define_primary_energy_limit(n: Network, sns: pd.Index) -> None:
...
+        if n.has_scenarios:
+            scenario, constraint_name = name
+            name = scenario + "-" + constraint_name
+        else:
+            scenario = slice(None)
+
         lhs = []
         rhs = glc.constant
-        emissions = n.carriers[glc.carrier_attribute][lambda ds: ds != 0]
+        emissions = n.carriers[glc.carrier_attribute][lambda ds: ds != 0].loc[scenario]
```
fneum: This is an inference by the name of the constraint. Better would be if we handled it similarly to the `investment_period` attribute of `GlobalConstraint`. I.e. introduce a new attribute `scenario`. If given, it specifies that it is only applied to a specific scenario. Otherwise, it is applied to all scenarios.

pypsa/optimization/global_constraints.py
```diff
@@ -257,43 +282,70 @@ def define_primary_energy_limit(n: Network, sns: pd.Index) -> None:
...
 
         if emissions.empty:
             continue
 
         # generators
-        gens = n.generators.query("carrier in @emissions.index")
+        emission_carriers = emissions.index
+        gens = n.generators[n.generators.carrier.isin(emission_carriers)]
```
fneum: Any particular reason to remove the `.query()`? Could be nice in combination with https://github.com/PyPSA/PyPSA/pull/1267.

pypsa/optimization/global_constraints.py
```diff
@@ -333,32 +386,51 @@ def define_operational_limit(n: Network, sns: pd.Index) -> None:
             if isnan(glc.investment_period)
             else sns[sns.get_loc(glc.investment_period)]
         )
+
+        if n.has_scenarios:
+            scenario, constraint_name = name
+            name = scenario + "-" + constraint_name
+        else:
+            scenario = slice(None)
```
fneum: See comment above.

pypsa/optimization/global_constraints.py
```diff
@@ -371,23 +443,28 @@ def define_operational_limit(n: Network, sns: pd.Index) -> None:
...
     """
     m = n.model
     glcs = n.global_constraints.query("type == 'transmission_volume_expansion_limit'")
 
     def substr(s: str) -> str:
         return re.sub("[\\[\\]\\(\\)]", "", s)
 
+    if n.has_scenarios and not glcs.empty:
+        msg = "Transmission volume expansion limits for stochastic networks are not implemented."
+        raise NotImplementedError(msg)
```
fneum: Shouldn't this one work out of the box since it will not be scenario-dependent (as it only includes investment variables)? At least it should be consistent with `define_transmission_expansion_cost_limit`.

pypsa/optimization/optimize.py
```diff
@@ -111,83 +170,110 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
-                    for period in sns.unique("period")
-                },
-                axis=1,
-            )
-            cost = active @ period_weighting * cost
+            weighted_cost = 0
+            for i, period in enumerate(periods):
+                # collapse time axis via any() so capex value isn't broadcasted
+                active = c.da.active.sel(period=period, name=ext_i).any(dim="timestep")
+                weighted_cost += capital_cost * active * period_weighting.iloc[i]
```
fneum: This `.iloc[]` is potentially dangerous, isn't it? E.g. when you solve with multiple investment periods but do not pass all investment periods through the `snapshots` argument (see e.g. https://github.com/PyPSA/PyPSA/pull/1308). Why not a `.loc[period]`?

pypsa/optimization/optimize.py
```diff
@@ -551,47 +664,82 @@ def assign_duals(self, assign_all_duals: bool = False) -> None:
...
+                        # Security constraints: preserve more information to avoid conflicts
+                        # e.g., "fix-s-lower-security-for-Line-outage-in-SubNetwork-0"
+                        attr_parts = attribute_name.split("-")
+                        dual_spec = (
+                            "-".join(attr_parts[1:])
+                            if len(attr_parts) >= 3
+                            else attribute_name
+                        )
+                    elif component_name == "GlobalConstraint":
+                        dual_spec = attribute_name
```
fneum: Global constraints are not snapshot-indexed. Why are they handled here?

pypsa/optimization/optimize.py
```diff
@@ -637,45 +784,50 @@ def post_processing(self) -> None:
...
-            return self._n.buses_t.p.reindex(columns=buses_i) @ Z
-
-        # TODO: if multi investment optimization, the network topology is not the necessarily the same,
-        # i.e. one has to iterate over the periods in order to get the correct angles.
-        # Determine_network_topology is not necessarily called (only if KVL was assigned)
-        if "obj" in self._n.sub_networks:
-            self._n.buses_t.v_ang = pd.concat(
-                [v_ang_for_(sub) for sub in self._n.sub_networks.obj], axis=1
-            ).reindex(columns=self._n.buses.index, fill_value=0.0)
+        if not n.has_scenarios:
```
fneum, line 825: This should also be well-defined for stochastic networks. Why skip it?

pypsa/optimization/constraints.py
```diff
@@ -588,12 +880,19 @@ def define_nodal_balance_constraints(
         ["Transformer", "s", "bus0", -1],
         ["Transformer", "s", "bus1", 1],
         ["Link", "p", "bus0", -1],
-        ["Link", "p", "bus1", get_as_dense(n, "Link", "efficiency", sns)],
+        [
+            "Link",
+            "p",
+            "bus1",
+            # dirty as hell, TODO make sure as_xarray handles case when links empty AND scenarios there
```
fneum: You want to leave that comment? ;)

fneum (dismissed):
Look great! Some minor things and questions for my understanding mainly.

pypsa/collection.py
lkstrp: It would be possible, but this is a regex. If you create that string automatically, it would be very difficult to understand

But I might remove all hard coded component names at some point in the future. There are already more in there then just here

pypsa/consistency.py
lkstrp: Because `n.c.buses.index` might also include the scenario dimension, while `.component_names` is always just the name dimension

pypsa/network/graph.py
lkstrp: It does not preserve the order, but reorders to default sort. So changing this to drop_duplicates is a very good point. I will check wherever else this is used

pypsa/network/index.py
lkstrp: Those headings are there for all dimensions

pypsa/network/io.py
lkstrp: the latter

pypsa/network/io.py
lkstrp: That is simply how dimension coordinates are currently stored in xarray. I don't like it at the moment, since there are basically two ways to create an xarray object (via 'io' and 'c.ds'). This will also be revised in a general io pr

test/test_z_docs.py
lkstrp: Yes, because doctests run globals, which can cause interference issues with other tests. We still want the freedom to show a lot of things in the Examples section, so that is why they just run at the end now

pypsa/networks.py
lkstrp: @FabianHofmann

pypsa/optimization/global_constraints.py
lkstrp: Because in general we don't like `.query()`. We can bring it back with 1267

pypsa/optimization/global_constraints.py
lkstrp: As discussed. All `NotImplementedError` are basically todos to add at some point.

pypsa/optimization/optimize.py
lkstrp: As discussed, there is a test for it for some reason and even if they should never be, we just handle it generally now

pypsa/optimization/optimize.py
Irieo: Right. I don't see a reason for `.iloc[]` here. Fixed.

pypsa/optimization/global_constraints.py
Irieo: The `.set_scenarios()` method broadcasts a static dataframe of global constraints, allowing the user to modify the RHS values.  NB We've also updated global constraints casting: `"scenario"` is now a dimension, so global constraints behave like any other constraints.

pypsa/optimization/global_constraints.py
Irieo: resolved, Q&A below.

test/test_stochastic.py
Irieo: There is another test verifying that the objective is equivalent; however, I agree this test doesn't say much on its own. Also, it uses a different network. I've added new assertions.

test/test_stochastic.py
Irieo: The bug occurred in the operation `(1 - standing_loss)**eh` due to a dimension mismatch with `scenarios`.

I've updated the test parametrization, it now captures both the dimension issue and the impact on storage levels caused by standing losses (again, with `scenarios`).

test/test_lopf_unit_commitment.py
Irieo: Done. There is a test for `ramp_limit_up` and `ramp_limit_start_up` now checking both model creation and that values are considered. This is not related to stoch opt, but `new-opt`'s xarray-based optimization backend.

fneum:
One problem with `n.set_investment_periods()`:

```sh
In [2]: import pypsa

In [3]: n = pypsa.examples.model_energy()
WARNING:pypsa.network.io:Importing network from PyPSA version v0.0.0 while current version is v0.35.1. Read the release notes at https://pypsa.readthedocs.io/en/latest/release_notes.html to prepare your network for import.
INFO:pypsa.network.io:Imported network 'Model-Energy' has buses, carriers, generators, links, loads, storage_units, stores

In [4]: INVESTMENT_PERIODS = [2025, 2035, 2045]

In [5]: n.set_investment_periods(INVESTMENT_PERIODS)
INFO:pypsa.network.index:Repeating time-series for each investment period and converting snapshots to a pandas.MultiIndex.

In [6]: n.investment_periods
Out[6]: Index([2025, 2035, 2045], dtype='int64', name='period')

In [7]: n.investment_period_weightings
Out[7]: 
Empty DataFrame
Columns: []
Index: [2025, 2035, 2045]
```

Expected output (and output in previous versions) would be:

```sh
In [7]: n.investment_period_weightings
Out[7]: 
        objective  years
period
2025          1.0    1.0
2035          1.0    1.0
2045          1.0    1.0
```

EDIT: Fixed in https://github.com/PyPSA/PyPSA/pull/1154/commits/a368e6ec846ce54dfed2634dbf5e16429202c306

pypsa/optimization/global_constraints.py
Irieo: Now implemented

pypsa/optimization/optimize.py
Irieo: We chatted with @FabianHofmann and decided to move it to a dedicated PR. Dropping that NotImplemented error requires changes in pf.py, while this PR is already massive.

pypsa/network/io.py
lkstrp: Yes

pypsa/network/power_flow.py
Irieo: Reverted back

pypsa/optimization/variables.py
Irieo: It is redundant already, spillage var is treated in stoch case

pypsa/networks.py
FabianHofmann: yes, that is correct. thanks for looking over @fneum! I would say the function name can stay as used in graph theory (even though it is rather cycle matrix).

FabianHofmann: should we name it cycle_matrix then, similar to incidence_matrix, adjacency_matrix ... ?

PR #1156, opened 2025-02-26
Plotting Library

lkstrp:
Ref #1106

Combination of a couple of PRs linked below
- #912 
- #1102
- #1118

EDIT:

**Syntax**
``` python
# Statistics plots
# Simplified, only essential arguments
n.statistics.energy_balance.plot()
n.statistics.energy_balance.plot(kind='bar')
# Expert mode, all arguments, but yields same result as default
n.statistics.energy_balance.plot.bar()

# Global plots
n.plot.map()  # moved from n.plot()
n.plot.explore()  # moved from n.explore()
n.plot.sankey_diagram()  # not yet implemented
n.plot.topology()  # not yet implemented
```

## ToDos 
Required
- [x] nice_names is still broken
- [x] Should `row`/ `col` maybe be renamed to `facet_row`, `facet_col` or even be combined into `facet`?
- [x] static vs time series default args in `manage_parameters`
- [x] Still improve `.maps()`. Choose which stats kwargs are needed/ don't pass `stat_kwargs` as argument.
- [x] Better integration of `maps.py` with `/statistics`. 
- [x] Neat docstring for `n.plot.x.map()`
- [x] Check that all plots in any combination work with default parameters
  - `.area(stacked=True)` is currently broken
- [x] Change charts from `seaborn.objects.Plots` to classic matplotlib `Figure`, `Ax` integration
- [x] Apply new discussed syntax: statistics plots to `n.statistics`
- [ ] Better errors and warnings
  - [ ] Wrong kwargs which are passed through
  - [ ] Unsupported statistics kwargs
  - [ ] Unknown carrier columns/ carrier
- [x] Move n.plot, explorer, iplot
- [ ] More unit tests, check coverage
- [x] Find better way to test plots
- [x] We might also wanna add `/plot` to linting

Optional (other PRs)
- [ ] Patch docstrings with actual default value from schema
- [ ] Documentation (maybe moved to another PR, since those interfer with statistics package)


FabianHofmann:
Some comments on the first try/sight: 

I would prefer to have the code for plotting the static maps in a dedicated module not under plot/statistics/. In `plot/statistics/maps.py` we would only define the functions that are used to transfer the statistic plots to the MapPlotter. 

There is still some bugs:
* When setting `nice_name` to True, the colors are not properly parsed. Probably needs a `set_index` of the `nice_name` column somewhere
* `n.plot.energy_balance(bus_carrier="H2", nice_names=False).map()` breaks with `multiple values for keyword argument 'bus_carrier'`

I still find the structure of two function calls still a bit unintuitive, as the arguments may contradict each other/have to be manually aligned. Let's say you want to make a bar plot for supply with carrier on the x-axis and for each country in a separate plot, I would imagine with this implementation it works like 
```
n.plot.energy_balance(bus_carrier="H2", groupby=["country", "carrier"], nice_names=False).bar(x="carrier", col="country")
```
which also seems to work. This is a bit cumbersome as you always have to align the plotting kwargs and statistics kwargs manually. I have the feeling that there is a strict mapping between plotting kwargs to statistic kwargs (as applied, but perhaps not perfectly defined in #1118 ). 
Another point is that things like 
```
n.plot.energy_balance(bus_carrier="H2", groupby=["country", "carrier"], nice_names=False).map()
```
just don't work. So, conflicts of kwargs have to be detected depending on the plotting function. 


Let's talk about it later, but in general I think we need to define a compatibility scheme between plotting kwargs and statistics kwargs. And then, for me, the question arises whether we should not directly go for the pattern `n.plot.energy_balance.bar(...)`

pypsa/plot/accessor.py
```diff
@@ -0,0 +1,480 @@
...
+    BarPlotGenerator,
+    LinePlotGenerator,
+    MapPlotGenerator,
+)
+from pypsa.statistics.expressions import StatisticsAccessor
+
+if TYPE_CHECKING:
+    from pypsa import Network
+
+# TODO fix removed nice_names=False
```
lkstrp, line 25:
```suggestion
```

pypsa/plot/accessor.py
```diff
@@ -0,0 +1,479 @@
...
+        self._n = n
+        self._stats_func = statistics_function
+
+    def bar(
+        self,
+        x: str | None = None,
+        y: str | None = None,
+        color: str | None = "carrier",
+        col: str | None = None,
+        row: str | None = None,
```
lkstrp, line 57: Combine to `facet` or rename to `facet_col` and `facet_row`?

pypsa/plot/accessor.py
```diff
@@ -0,0 +1,479 @@
...
+        draw_legend_lines: bool = True,
+        draw_legend_arrows: bool = False,
+        draw_legend_patches: bool = True,
+        legend_circles_kw: dict | None = None,
+        legend_lines_kw: dict | None = None,
+        legend_arrows_kw: dict | None = None,
+        legend_patches_kw: dict | None = None,
+        bus_split_circles: bool = False,
+        kind: str | None = None,
+        # TODO: Additional stat kwargs needed?
```
lkstrp, line 416: Which additional stat kwargs do we need here

FabianHofmann:
@lkstrp I played around with the new implementation and the main code works okayish now (I had to make one fix in order to make area and line plot with snapshots work). I'm being honest that I was a bit frustrated as only a few convenience features from #1156 worked and I spent really as lot of time efficiently integrating them, and now we would have to fit them in again... but there is always hope ;) 

Since we want to be bullet-proof allow me to make some criticism: 

One major point is that by standardizing the plotting argument, we lose some important nuances of the the functions. For example if you want to plot the energy balance as a map, I would expect to see the flows between buses as well. And since for flows you want to see how large they are, the type of the legend should be a flow legend and not a line legend. To get there, you have to write out:

```
n.plot.energy_balance.map(transmission_flow=True, draw_legend_arrows=True, draw_legend_lines=False)
```
which is way too complicated. Ah and I forgot that actually we want to have the bus circle split for this as well...

Another example is in `n.plot.energy_balance.area`. In the standard use case this should plot a time series. But now, as we standardize the arguments, it shows a quite meaningless plot with carriers and you would have to explicitly say 

```
n.plot.energy_balance.area(x="snapshot", y="value", colo="carrier")
```


Secondly, I am not too happy with with new order in the syntax. Even though grammatically `n.plot.energy_balance.map` makes more sense than `n.plot.map.energy_balance`, it might be quite restrictive. The plotting accessor is now limited to the scope of statistics accessor. As soon as we want to introduce plots that are not related to statistics like `topology` or `sankey_diagram` I am wondering where they should go. I originally would have imagined things like `n.plot.sankey_diagram` or `n.plot.topology` in the future, but this make no sense now. One could tackle that through making the statistic plots an explicit subcategory like `n.plot.statistics.energy_balance.map()`. This tends to be a long function call though... 


A minor point: We now loose access to some more plotting function from the MapPlotter. In #1118 we could do things like 

```
n.plot.map.set_bounds(...)
n.plot.draw_map()
n.plot.map.energy_balance() # respect the bounds
```

I still owe you a Kudos that you were able to restructure the things to fast. However, I am still thinking about reviving #1118 and just live with some redundancies, if things turn out to be too complicated.
Anyway I would really like to push this, and perhaps by the end of next week we can have something merge-ready.

lkstrp:
Let’s go, we wanna make this bullet proof 💃

Probably I did not make the kwarg matching/ schema process clear enough. Your first points all break down to different default arguments. This is still possible, also without reimplementing all methods. I can just take over your defaults from #1156 if they are already nailed down. 

The second one is syntax, I have no hard feelings there. I still think `n.plot.energy_balance.map` is more intuitive than `
`n.plot.map.energy_balance`. While `n.plot.sankey_diagram` isn’t logically the best fit anymore, the accessor is not limited to statistics functions and this can be added. I think it’s the better compromise compared to the less intuitive `n.plot.map.energy_balance`. But let’s get some more opinions here. 

So most things can all be handled. For your last point I don’t see a good solution right now. But getting the MapPlotter should be possible, it’s more of an API question again. I am quite against bringing back all those redundancies. What other convenience features are you missing? 

And yes! Let’s get this thing merged.

pypsa/plot/accessor.py
FabianHofmann: none that I can think of at the moment, should not be to hard to add them later?

pypsa/plot/accessor.py
```diff
@@ -399,32 +417,46 @@ def map(
...
         # TODO: Additional stat kwargs needed?
         **kwargs: Any,
     ) -> tuple[Figure, Axes]:  # Store locals to pass same signature to plotter
-        local_vars = locals()
-        del local_vars["self"]
-        del local_vars["kwargs"]
+        if stats_kwargs is None:
+            stats_kwargs = {}
+
+        plot_kwargs = locals()
```
FabianHofmann, line 441: no luck with the wraps decorator?

lkstrp: Not how I would like to as a replacement for the plot_kwargs dict. But maybe we were talking past each other

FabianHofmann: tried it out as well, did not work

pypsa/plot/accessor.py
FabianHofmann: let's rename to facet_x, this is the same as plotly express has it

lkstrp:
``` python
# Statistics plots
# Simplified, only essential arguments
n.statistics.energy_balance.plot()
n.statistics.energy_balance.plot(kind='bar')
# Expert mode, all arguments, but yields same result as default
n.statistics.energy_balance.plot.bar()

# Global plots
n.plot.map()
n.plot.explorer()
n.plot.sankey_diagram()
n.plot.topology
```

FabianHofmann:
Thanks @fneum for the review, very helpful, lukas and I have already touched basis on this. but here some thoughts


> This PR is very (!) large. Would it be possible to split it up into parts?
> * Separate PR for changes to backend structure of `n.plot()` (without mixing in the statistics plotting functionality).

I know it is huuuge :(  but it will be worth it. yes, could do that :) 

> * I have not yet understood what the advantage of `seaborn` plotting objects is. From what I have seen in the statistics plotting functionality, there are no non-standard plots that wouldn't be straightforward directly with `matplotlib`. Using `matplotlib` would have the advantage that it's familiar to PyPSA users (it is what `pandas` uses in static plotting, it is used in `n.plot()`, they can pass existing axes, etc.. We would avoid mixing two different static plotting frameworks.


we should get rid of the seaborn object plots it is too complicated, I am proposing to use seaborn's facet grid instead with internal custom plots for area and bar, so we get the support for stacked plots back in.


> * Relative to the added functionality, the amount of added code is quite large. I haven't examined it in detail, but are we sure we're not over-engineering this? This aspect also relates to how we imagine the user to use these functions. Is it for creating production ready figures or for quickly exploring the network? For the latter, it might also be sufficient to offer just a reduced set of plotting functions (e.g. time-resolved area charts for energy balances by bus carrier). It's straightforward and concise to build most plots yourself with `n.statistics.x()`. We want to help users for the difficult tasks (maps and energy balances).

the functions should be fully flexible and all combinations of statistics and plot types should be supported. not generalizing it and limiting ourselves to just a few combinations, I don't see a benefit in it. imagine use cases like, you have a multi invest network and you quickly want to see all h2 installations for the planning horizons. this then just one function call. there is loads of other stuff you can think about...

> * Separate PR for `resample_timeseries()`. We had plans to add a clustering module in `pypsa.clustering.temporal`. Wouldn't it make more sense to place it there? It is also not essential for the plotting, so one could leave that out.

you're right. we can kick out the functionality for now, but would vote for supporting them generally (temporal clustering is a better place probably!). 

> * The statistics plotting protocols might be a bit too strict (e.g. regarding missing colors in the consistency checks). Couldn't one assign colors from the standard pallette in this case for the plot?

perhaps that makes sense. not sure about the ambiguity here.


> * I tested the statistics plots with a few larger networks with many carriers. Here it would be more suitable to use horizontal bars and dynamic sizing of the figure depending on the number of elements.

agree, you can change between horizontal and vertical by interchanging `x` and `y`. the dynamic adjustment of the 

> * The plotting functions need a scaling attribute. You want to be able to show bn€ or TWh / GW.

which plots do you mean, the charts or the maps? For maps there is the autoscaling internally. Units are derived from the bus carrier. for the charts there is no units and scaling yet. This is something I would add later

> * There are a lot of unrelated auto-format changes that clutter the diff.

don't think we can avoid that...

fneum:
Much better than on my first attempt!

I went through a couple of use cases I would use the new plotting functionality for:

https://tubcloud.tu-berlin.de/s/QnzA5EExoPkwywJ

There are a few minor things, but they should be possible to address.

I'll test n.plot() next, to ensure nothing broke.

fneum (changes requested):
Make sure to bring in 

https://github.com/PyPSA/PyPSA/pull/1179
https://github.com/PyPSA/PyPSA/pull/1179

Before-after comparison for a more complex plot also looks ok!

image

The only thing that broke was `flow` -> `link_flow`. It would be good if `flow` with MultiIndex were still supported for a transition period with deprecation warning.

pypsa/plot/maps.py
```diff
@@ -0,0 +1,2647 @@
...
+            ):
+                msg = "The `flow` argument is deprecated, use `line_flow`, `link_flow` and "
+                "`transformer_flow` instead. You can't use both arguments at the same time."
+                raise ValueError(msg)
+            if isinstance(flow, pd.Series) and isinstance(flow.index, pd.MultiIndex):
+                msg = (
+                    "The `flow` argument is deprecated, use `line_flow`, `link_flow` and "
+                    "`transformer_flow` instead. Multiindex Series are not supported anymore."
+                )
+                raise ValueError(msg)
```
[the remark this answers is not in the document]
FabianHofmann: good one!

FabianHofmann:
I am getting confused as the two prs have the same name now, so here we go again:

@fneum thanks for testing it out. as far as I see the following points were raised:

* unit scaling in chart plots: yes, that I would like to support that. I am thinking this can be a good follow-up, as we can also introduce some auto-scaling as we have it for the maps. in maps we would have to provide  a manual way to set the scale
* the axis is not a supported arguments. the reason is that is becomes incompatible with the plots as soon as you set facet_col, facet_row which returns multiple axes.  the idea is that you can alter the axis or axes object which you get as a return value. but what is confusing is, that there is no value error raised...
* aggregation to location in maps, this can be made possible after merging in #1182 and adding an additional argument. 
* supporting undefined carrier colors: carriers in ["", " - ", None] should now be working, however I would skip a color palette for other undefined carriers. we could think about a helper function `n.components.carrier.set_colors()` which you could call one time in advance. 
* We did not include any argument support in the `n.statistics.x.plot` function as arguments between `kind="bar"` and `kind="map"` are not shared. bringing that together, would be quite a hassle giving that both `n.statistics.x.plot.bar/map` support kwargs themselves.  
* pandas time stamp on x-axis. could look into it... however would not set a prio on that. 
* sorting of bar value.. the sorting comes from seaborn in most cases, it think is alphabetical (with some upper/lower case distinction it seems?)

FabianHofmann:
@lkstrp @fneum all tests run through now (code cov is not fulfilled though). I would vote for bringing this is and make follow-ups

lkstrp:
@FabianHofmann Agreed. It will be your honour to merge this beauty 💃

The question is which other PR's we wanna put in the release. I will not have much time until Thursday. Maybe we move #1183 and #1149 to another release?

FabianHofmann:
> @FabianHofmann Agreed. It will be your honour to merge this beauty 💃

alright, let's go :)

> The question is which other PR's we wanna put in the release. I will not have much time until Thursday. Maybe we move #1183 and #1149 to another release?

how pressing is the release? I you want to do it in the next days and resource is scarce, I would say, let's skip them for now and just make another release soon. but I am also happy to wait a little with the release.

lkstrp:
I'll make a release today. People wanna get the Excel feature 🫣

PR #1159, opened 2025-02-28
Add new io type: `.xlsx`

lkstrp:
- Adds excel support in a similar way than existing csv dir support (sheets instead of single files)
- It is slow
- I'll add another IO PR, so no doc update in this one

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,7 +73,7 @@ jobs:
     - name: Install package and dependencies
       run: |
         python -m pip install uv
-        uv pip install --compile --system "$(ls dist/*.whl)[dev,hdf5]"
+        uv pip install --compile --system "$(ls dist/*.whl)[dev,hdf5,excel]"
 
 
     - name: Run unit tests # Also includes doctests
--- a/doc/api/networks.rst
+++ b/doc/api/networks.rst
@@ -59,6 +59,8 @@ Input and output methods
     ~pypsa.Network.merge
     ~pypsa.Network.import_from_csv_folder
     ~pypsa.Network.export_to_csv_folder
+    ~pypsa.Network.import_from_excel
+    ~pypsa.Network.export_to_excel
     ~pypsa.Network.import_from_hdf5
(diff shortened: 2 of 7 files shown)


fneum:
Cool! Did you check if loading all sheets first is faster (e.g. in `__init__()`)?
```py
df_dict = pd.read_excel(self.excel_file_path)
```

lkstrp:
No, and it probably is. But we would lose our general IO interface, with all the processing, type checking and other things that are done there (`_import_from_importer`). I don't think we want to do that. But I have not dipped deeper, as #1128 will change a lot there.

coroa:
> . But we would lose our general IO interface, with all the processing, type checking and other things that are done there (`_import_from_importer`). I don't think we want to do that. But I have not dipped deeper, as #1128 will change a lot there.

Not necessarily, why don't you load everything into a dict in `ExcelImporter.__init__` and then only serve from that dict in the methods?

lkstrp:
I fixed a few things. The main problem was the use of the context writer. I'm not sure we still need `calamine` as it can't replace `openpyxl` for writing to a file. But you can now pass an engine and both dependencies are optional

fneum (approved):
Excellent, I tested it on a couple of larger networks as well and it is actually decently fast! 100 regions, 3-hourly resolution, sector-coupled was written in 1 minute.

PriyeshGosai:
@lkstrp 

There appears to be a bug when exporting the carriers table. The resulting Excel sheet includes an extra row at the top—specifically, the index label 'name'—which causes issues during re-import. When the file is loaded again, it incorrectly interprets 'name' as a new carrier entry.

lkstrp:
@PriyeshGosai  I think I stumbled across that. Could you please check whether that is still an issue in the current master branch?

PriyeshGosai:
> @PriyeshGosai I think I stumbled across that. Could you please check whether that is still an issue in the current master branch?

I checked the main branch, and the error only pops up when you list carriers without any changes to the carrier attributes.

PR #1160, opened 2025-03-02
Xarray view 2

FabianHofmann:
I am playing around a little with #1076 and see whether I can get somewhere

## Checklist


FabianHofmann:
### clearify

* would we already iterate over components instead of items from the lookup table in `create_model`?
* can we move the lookup table content to components definition ?
* potentially rename `Component.as_xarray` to something different as `Component.dataarray`?
* would we like to have a `Component.variables` for quickly accessing variables?

@Irieo I saw you too are working in a similary direction in #1154, I think we can align here. I haven't touched `constraints.py`. With using the `as_xarray` function the code should now smoothly run with the additional `scenario` dimension.

Irieo:
> @Irieo I saw you too are working in a similary direction in https://github.com/PyPSA/PyPSA/pull/1154, I think we can align here. I haven't touched constraints.py. With using the as_xarray function the code should now smoothly run with the additional scenario dimension.

yes, `as_xarray` helps to switch to xarray math operations and handle scenario dimension properly  (the long awaited thing). I work with constraints.py and objective in that PR, certainly worth aligning on that! @FabianHofmann

PR #1166, opened 2025-03-05
Refactor optimization code

Irieo:
This PR refactors `/optimization` module to (i) align code with the new component class functionality and (ii) to use xarray operations for mathematical expressions (a precursor for https://github.com/PyPSA/PyPSA/pull/1037)

## Checklist


pypsa/optimization/constraints.py
```diff
@@ -52,29 +53,35 @@ def define_operational_constraints_for_non_extendables(
...
+    transmission_losses : int
+        Whether to consider transmission losses
     """
-    dispatch_lower: DataArray | tuple
-    dispatch_upper: DataArray | tuple
-
-    fix_i = n.get_non_extendable_i(c)
-    fix_i = fix_i.difference(n.get_committable_i(c)).rename(fix_i.name)
+    component = as_components(n, c)
+    fix_i = component.fixed
```
FabianHofmann, line 60: let's use the `.get_non_extendable_i` for now. With that we can easily resolve https://github.com/PyPSA/PyPSA/issues/848 with the additional argument `rename_index`

pypsa/optimization/constraints.py
```diff
@@ -52,29 +53,35 @@ def define_operational_constraints_for_non_extendables(
...
     if fix_i.empty:
         return
 
-    nominal_fix = n.static(c)[nominal_attrs[c]].reindex(fix_i)
-    min_pu, max_pu = get_bounds_pu(n, c, sns, fix_i, attr)
-    lower = min_pu.mul(nominal_fix)
-    upper = max_pu.mul(nominal_fix)
+    nominal_fix = component.as_xarray(component.nominal_attr).rename(
+        {component.name: fix_i.name}
+    )
```
FabianHofmann, line 68:
```suggestion
    nominal_fix = component.as_xarray(component.nominal_attr, inds=fix_i)
```
this already renames the dimension

pypsa/components/descriptors.py
```diff
@@ -117,11 +118,32 @@ def get_activity_mask(
...
+    as_xarray: Literal[False] = False,
+) -> tuple[pd.DataFrame, pd.DataFrame]: ...
+
+
+def get_bounds_pu(
+    c: Components,
+    sns: Sequence,
+    index: pd.Index | None = None,
+    attr: str | None = None,
+    as_xarray: bool = False,
```
FabianHofmann, line 146: I am still thinking that this should also be moved to the component classes in order to avoid the if cases to differentiate between components here

FabianHofmann (commented):
just some comments when scrolling through

pypsa/optimization/constraints.py
Irieo: alright. NB I keep .rename(fix_i.name) as it seems .difference() call can drop the index name

pypsa/optimization/constraints.py
Irieo: ok

pypsa/components/array.py
```diff
@@ -62,6 +64,10 @@ def as_dynamic(
     2015-01-01 01:00:00         0.485748             1.0     0.481290         1.0        0.752910            1.0
 
     """
+    # Check if we are in a power flow calculation
+    stack = inspect.stack()
+    in_pf = any(os.path.basename(frame.filename) == "pf.py" for frame in stack)
```
lkstrp, line 69: As discussed and to keep it noted: This would need to be handled in `pf.py` and not here

pypsa/optimization/constraints.py
lkstrp: Just a note here that this will rename the dimension to whatever `fix_i` had as an index name. But we want to be stricter here and have fixed dimension names, so always return the same, no matter what your passed selection index is

Irieo:
NB unit tests are passing up until [c6055eb](https://github.com/PyPSA/PyPSA/pull/1166/commits/c6055eba19b82eada910af55e7d29dddb7d0f687) and break in [4991b55](https://github.com/PyPSA/PyPSA/pull/1166/commits/4991b552bdbe03f5b37355ada6f8041bd2f7f348)

Interesting that 4991b55 is shown as "passed" on windows-latest, even though some tests have failed (and marked accordingly) and the workflow is stopped with out-of-memory error

Should we pass `--strict-exit-code` to pytest command, or something along this idea, so that if internal errors occur (that do not propagate to exit code), we get red CI?

lkstrp:
True, this needs to be fixed. `--strict-exit-code` is just a flag made up by the LLM of your choice though. I'll fix that in the main branch

PR #1171, opened 2025-03-11
fix(expressions): use the absolute efficiency for capacity w/o bus_carrier

coroa:

`n.optimize.expressions.capacity` needs to use the absolute efficiency, unless a bus_carrier chooses a non-standard viewpoint.

the change is in strict analogy to statistics.installed_capacity (ref. https://github.com/PyPSA/pypsa/blob/0566423ff32c309008129b134e323e00d0bf9b1e/pypsa/statistics/expressions.py#L514-L521 )

I guess we should add a warning to the release notes, since that is in principle a breaking change? I am a bit fuzzy about the intricate details of `at_port` and `bus_carrier`. Do you think you can add the release notes, @FabianHofmann ?

## Checklist

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -29,6 +29,12 @@ Features
   expression (regex), enabling more flexible filtering options. 
   (https://github.com/PyPSA/PyPSA/pull/1155)
 
+
+* The expressions function `n.optimize.expressions.capacity` now uses the absolute efficiency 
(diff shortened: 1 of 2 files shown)


FabianHofmann (approved):
good catch!

FabianHofmann:
I think we can argue that this is a bugfix to be consistent with the statistics module, hence we can skip the warning(?)

coroa:
> I think we can argue that this is a bugfix to be consistent with the statistics module, hence we can skip the warning(?)

Sure, i think that is fine.

PR #1173, opened 2025-03-12
feat: generic nested options and set default params

lkstrp:
- Improvements to the module options. Now allows arbitray nested options, better Exceptions etc. 
- Deprecates `statistics.parameters` and moves them to `pypsa.options.statistics` instead:

``` python
>>> pypsa.options.params.statistics.describe_options()
PyPSA Options
===========
drop_zero:
    Default: True
    Description: Default value for the 'drop_zero' parameter in statistics module.
nice_names:
    Default: True
    Description: Default value for the 'nice_names' parameter in statistics module.
round:
    Default: 5
    Description: Default value for the 'round' parameter in statistics module.
```

**Logic**
The idea is to be able to change the parameter defaults at runtime without having to pass them to all methods. The hierarchy would be
1. Function-level parameters
2. `pypsa.options` default
3. (could be added), system wide setting, either via environment variables or `.rc` file, similar to matplotlib

To set options you could do
``` python
pypsa.options.params.statistics.nice_names = False
```

or via context manager to be more granular
``` python
with pypsa.option_context("params.statistics.nice_names", False):
    pass
```

This could be of course be extended to any other set of parameters in any method/ function. 

**To discuss**
What remains to be discussed is how to handle default kwargs in the method signature. Currently `nice_names` is included, but `round` and `drop_zero` are not. This needs to be resolved. We can

1. add them all to the signature
``` python
    def capex(
        self,
        ...
        nice_names: bool | None = None,
        round: int | None = None
        drop_zero: bool | None = None,
        ...
    ) -> LinearExpression:
```
This might bloats up the signature, specially regarding future additions and #1156.

2. pass them as `**kwargs`
``` python
    def capex(
        self,
        ...
        **kwargs: Any
        ...
    ) -> LinearExpression:
```
This would be similar to matplotlib, but would result in a less readable docstring. See "Other parameters" [here](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html). And it also complicates things if `**kwargs` is already being used.

## Checklist


FabianHofmann:
> I vote for option 1.

agree, I think this make the api more transparent

PR #1176, opened 2025-03-17
feat: use statistics handler as drop in method replacement

lkstrp:

To reduce the diff of #1156 and for easier review, as this interferes with the existing API:
- Wraps each statistic method with a handler to allow the suggested plotting syntax (`n.statistics.energy_balance.plot()`), while only extending the API and not changing it.
- Adds `carrier` argument to statistics functions

I would also like to refactor our entire accessor design. This is very different from module to module, and our `StatisticsAccessor' is not really just an accessor. But I think it's best to move this after the plotting library is merged.

## Checklist

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -29,6 +29,8 @@ Features
   expression (regex), enabling more flexible filtering options. 
   (https://github.com/PyPSA/PyPSA/pull/1155)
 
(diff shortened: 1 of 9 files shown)


FabianHofmann:
that looks neat!

FabianHofmann (approved):

PR #1182, opened 2025-03-21
feat: add "location" attribute to buses

fneum:


coroa:
Does this mean the included `country` grouper is not supported any longer? I have been using it quite liberally, though.

fneum:
I think it can stay. It's additional. In PyPSA-Eur, we use the grouping by `n.buses.location` a lot, although it is not a standard Bus attribute. Same for the country attribute the country grouper is based on. But "country" is maybe a too specific attribute?!

lkstrp (commented):
But right now you are deprecating the country grouper without any warning?

FabianHofmann:
I would strong vote for keeping the county aggregator and just add a new location aggregator

fneum:
My bad, this is exactly what I thought I had done. Just seemed to have forgotten copying the country function first. Sorry!

lkstrp (approved):

PR #1184, opened 2025-03-21
fix: replace deprecated empty() function with property

FabianHofmann:


fneum:
Then can be merged with one of the following adjustments:
1. Bump minimum version of `linopy`.
2. Version-dependent check of `.empty` or `.empty()` to support older linopy versions.

fneum (approved):

PR #1189, opened 2025-03-26
Interactive Plotting 

### LINKED ISSUE #292: Extend frontend plotting functions
It would be useful to have simple frontend plotting functions to display results and input data beyond the `network.plot()`. Like:

- [ ] dispatch plot (area chart)
- [ ] bar chart system cost, capacity, energy
- ... (please extend)

One could use `hvplot` or `plotly` for interactive plots, or one sticks with `matplotlib` static plots.

One could also leverage `pyam` plotting functionalities (esp. if one creates a PyPSA IAMC compatible output function).

We could get inspiration from the `pandas` plotting functions like `network.plot.dispatch(time=..., nodes=...)`.

Could also draw on https://github.com/martavp/PyPSA-plots and https://github.com/fneum/pypsa-dashboard.

FabianHofmann:
Closes #292 

This PR introduces mirror of the recently added plotting library for interactive **chart** plots based on plotly. 

The accessor is now chosen to follow 

`n.statistics.energy_balance.iplot.x`

For map plots the iplot function in the maps.py would have to be updated.  

## Checklist

- [ ] make sure, capacity units are not mixed #1191 
- [x] think about plotly being a fixed deps
- [x] hovering: show all values in parallel
- [ ] time series in MWh on y-axis
- [ ] revenue and curtailment plots unclear


fneum (changes requested):
Just tested the same plots as I did in #1156. Especially the energy balance area charts stand out.

I have a few comments:

1. One of the largest potentials for the energy balance area charts is to quickly read out values from the hovering information. This does not work great at the moment, as the hover information often falls to technologies that are 0. Perhaps the default could activate hovering for all techs in the marked-up snapshot? A bit like: https://www.energy-charts.info/charts/power/chart.htm?c=DE.

2. We should make `plotly` an optional dependency.

3. Floating an idea: Instead of `iplot`, we could also follow the `pandas` approach by setting a plotting backend as an option: 

https://plotly.com/python/pandas-backend/

```py
import pandas as pd
pd.options.plotting.backend = "plotly"
pd.DataFrame(dict(a=[1,3,2], b=[3,2,1])).plot()
``` 

5. Some housekeeping to do, like adding new functions to API, docs and release notes.

6. Perhaps rename `StatisticsIPlotter` to `StatisticsInteractivePlotter`?

FabianHofmann:
@fneum thanks, 

1. yes, will cover that, might be a tricky one, but doable
2. do you think so? afaik plotly is very compatible with other libraries, not like geo dependencies. I think the previous constraint came from times where installation was not as stable. also with packages as `uv` it is just a matter of milliseconds
3. thought about that too, however the interface is not as flexible as in the plotly express module
4. yes, will do
5. yes, will do

lkstrp:
2. @FabianHofmann I think you are right. Plotly is way smaller than I thought
3. This was meant as an API alternative to iplot. We can use the same options as done in pandas and keep `.plot` for both interactive and static plots. I am just not sure what this means for different method signatures

fneum:
Great, so we agree on `plotly` as a standard dependency.

I don't have strong feelings about `.iplot()` versus `pd.options.plotting.backend`. We can also keep `.iplot()` if the alternative is challenging.

Otherwise, good to go?

FabianHofmann:
@fneum I would say we could merge as soon as tests pass. I would prepare a follow up on refining the units. this seems to be a bigger issue also touching the statistics module

fneum (approved):

PR #1191, opened 2025-03-27
feat: exclude stores from capacities by default

FabianHofmann:
The capacity statistics functions `optimal_capacity`, `installed_capacity`, and `expanded_capacity` now exclude storage capacities by default, i.e., when the `storage` parameter is set to False. This ensures that the Store component is not included in the statistics, preventing the mixing of different units of capacity. As a result, power production capacities are consistently reported in MW, while energy storage capacities are reported in MWh.

## Checklist


fneum (approved):
LGTM, only the docstring of ‘storage’ also needs to be adjusted.

lkstrp (approved):
How careful do we need to be about changing the statistics output? Also regarding #1185? `storage` defaults to False here

fneum:
Yes, careful. At least any other statistics function that uses these functions should be double-checked. At first sight I only see `capacity_factor` that depends on it.

I don't think CAPEX or OPEX are affected by it. OPEX only considers the dispatch. CAPEX calculates costs not based on capacity statistics: https://github.com/PyPSA/pypsa/blob/cf9fb120ae9c1c76fb84db123f0d64317ace7eb5/pypsa/statistics/expressions.py#L456

FabianHofmann:
@fneum how would the capacity factor of a store component be defined? by the average energy level divided by the energy storage capacity?

lkstrp:
What's the status here @FabianHofmann ?

fneum (changes requested):
Revisiting this PR, I no longer think it's a good idea to just drop the Stores from the capacity statistics. My main concern now is that it is not backward compatible (someone running the code from previous analysis with a new PyPSA version would get different results than before).

Sorry for the change of heart!

lkstrp:
I am closing this as stale. Reopen if you wanna restart the discussion.

PR #1201, opened 2025-04-03
fix: maps for statistics module

FabianHofmann:
my bad, the plots seemed to be broken. should work now. @lkstrp I think we forgot unit tests for these right?

## Checklist


lkstrp:
@FabianHofmann They have unit tests, probably the tolerance is just too high now. Different behavior on different OS is fixed, they are only run on Python 3.13 and Ubuntu

PR #1212, opened 2025-04-15
Introduce `NetworkCollection`

FabianHofmann:
This PR adds a `Networks` class, a container of `pypsa.Network` objects that allows to run statistics, plot and ease the comparison. 

@lkstrp if you want to take a first look

## To be done:
- [x] ~Plot statistic maps with harmonized scaling~ Disable maps plots for NetworkBundle
- [ ] Create tutorial notebook 
 

## Checklist


pypsa/networks.py
```diff
@@ -2109,3 +2121,122 @@ def iterate_components(
...
         )
+
+
+class NetworkCollection:
+    """
+    Container for multiple Network objects indexed by a pandas Index or MultiIndex.
+    """
+
+    def __init__(
+        self, networks: Sequence[Network] | pd.Series, index: pd.Index | None = None
```
lkstrp, line 2132: What about using `*args, **kwargs` here?

With `*args` being a sequence and with `kwargs` having the ability to assign index names? This would be more collection like (similar to `list`, `dict` etc).

E.g.:
`NetworkCollection(n1=n, n2=nn2)`
and
`NetworkCollection(n, n2)`

And instead of an integer, I would use the network name as the default index

pypsa/networks.py
```diff
@@ -2109,3 +2121,122 @@ def iterate_components(
...
+        """
+        all_carriers = [n.carriers for n in self._networks]
+        combined_carriers = pd.concat(all_carriers)
+        # Keep the first occurrence of each carrier based on the index
+        unique_carriers = combined_carriers[
+            ~combined_carriers.index.duplicated(keep="first")
+        ]
+        return unique_carriers.sort_index()
+
+    def apply(self, func: Callable[[Network], Any]) -> pd.Series:
```
lkstrp, line 2196: I am not a fan of `apply` to run network functions. Can we not mirror `Network` behavior?

E.g.
`nc = NetworkCollection(n1=n, n2=n2)`
and
`n.plot()`
`nc.plot()`

We probably don't wanna inherit, but this should be possible with either `__getattribute__` or meta classes. I can also look into that if you want. But I would prefer the API that way

pypsa/statistics/expressions.py
```diff
@@ -2193,3 +2190,117 @@ def market_value(
         df.attrs["name"] = "Market Value"
         df.attrs["unit"] = "currency / MWh"
         return df
+
+
+class StatisticsAccessorMulti:
```
lkstrp, line 2195: The above approach might also make the `StatisticsAccessorMulti` unnecessary

lkstrp (commented):
I have still a few more API concerns that could also simplify the implementation.

In my opinion, the best case would be that `NetworkCollection` behaves exactly in the same way as a network, while many methods cannot be called and raise a error.
But the logic would simply be that any method is passed through the equivalent for each network with some optional processing on the output. This could be a bit similar to the `StatisticsPlotHandler` and we wouldn't need to implement custom accessors. And `.apply` would only be needed on functions which are not already existing Network methods.

We can leave the example notebook for now, or push it directly to #1211 in another PR. But this needs some finetuned docs anyway.

And how do we wanna abbreviate `NetworkCollection`? We have `n` for `network`, `c for `Components, `nc` for `NetworkCollection`? Not sure

FabianHofmann:
> I have still a few more API concerns that could also simplify the implementation.
> 
> In my opinion, the best case would be that `NetworkCollection` behaves exactly in the same way as a network, while many methods cannot be called and raise a error. But the logic would simply be that any method is passed through the equivalent for each network with some optional processing on the output. This could be a bit similar to the `StatisticsPlotHandler` and we wouldn't need to implement custom accessors. And `.apply` would only be needed on functions which are not already existing Network methods.
> 
> We can leave the example notebook for now, or push it directly to #1211 in another PR. But this needs some finetuned docs anyway.
> 
> And how do we wanna abbreviate `NetworkCollection`? We have `n` for `network`, `c for `Components, `nc` for `NetworkCollection`? Not sure

I like `nc`, the issue I see with mirroring the `Network` functionality is that as soon as a new method is added to the `Network` class which does not work automatically with `NetworkCollection` you have to explicitly disable it or?

FabianHofmann:
@lkstrp the pr is 98% ready to be reviewed. however I am still facing the weird behavior that the `drop_zero` option for statistics in the pypsa.options handler is set to False in the CI even though that is not the case in the local tests. do you have an idea where this could come from?

lkstrp:
Looks beautiful 💃
All fixed and good to go now. We can still improve a couple of things, e.g. add a validation if people try to put wild Networks in a Collection and some parts are very verbose. But I think it is good for now and let's get this in.

@fneum If you also want to have a look

pypsa/statistics/expressions.py
```diff
@@ -1363,7 +1505,13 @@ def system_cost(  # noqa: D417
...
             round=round,
         )
-        df = capex.add(opex, fill_value=0)
+        # TODO It would be better if the empty series return has index names
+        if not capex.empty and not opex.empty:
+            df = capex.add(opex, fill_value=0)
+        elif not capex.empty:
+            df = capex
+        else:
+            df = capex
```
fneum, line 1514: The else should be `df = opex`!

pypsa/statistics/expressions.py
```diff
@@ -1213,7 +1354,7 @@ def func(n: Network, c: str, port: str) -> pd.Series:
                 "spill_cost",
             ]:
                 if cost_type in cost_types_ and cost_type in n.static(c):
-                    attr = lookup.query(cost_type).loc[c].index.item()
+                    attr = lookup.query(cost_type).loc[c].index.item() + port
```
fneum, line 1357: Good catch! That's for links with `p0`, right? The important thing is that any `p1` or so would make this calculation invalid. Is that caught?  That's the difference between `statistics/expressions.py` and `optimization/expressions.py` :)

pypsa/statistics/expressions.py
```diff
@@ -1199,7 +1340,7 @@ def opex(  # noqa: D417
         else:
             cost_types_ = list(cost_types)
 
-        @pass_empty_series_if_keyerror
+        # @pass_empty_series_if_keyerror
```
fneum, line 1343: This error should still be handled somewhere, shouldn't it?

pypsa/plot/statistics/plotter.py
```diff
@@ -18,7 +21,7 @@
     from matplotlib.axes import Axes
     from matplotlib.figure import Figure, SubFigure
 
-    from pypsa import Network
```
fneum, line 21: Wouldn't it make sense to keep `pypsa.Network` accessible?

pyproject.toml
```diff
@@ -64,7 +64,8 @@ dev = [
     "scikit-learn",
     "pre-commit", 
     "ruff",
-    "mypy"
+    "mypy",
+    "jupyter>=1.1.1",
```
fneum, line 68: What is that dependency needed for, do we really want it?

fneum (commented):
Nice concept, looking forward to trying it in the future.

I found one major typo which should be fixed before the next release in the system cost calculation in the statistics module.

There are nice examples in the docstrings but none made it to an example notebook how to use it.

The API reference was also not added to the documentation (I guess that has to do with the upcoming doc change?).

I will hold you to adding the documentation, and example and API reference when the new docs are there :)

pypsa/plot/statistics/plotter.py
lkstrp: It is, this is just to prevent circular imports and only for type checking anyway (so not actually runtime code)

pyproject.toml
lkstrp: It's not needed for anything other than convenience if people use Jupyter notebooks. But it's an optional dev dependency and, especially with `uv`, bloated environments aren't really an issue

lkstrp:
Good catch, thank you!

Yes as you say, User Guide and API reference will go in the 1.0 docs. I will mark it as experimental for 0.35. There may also be smaller changes to it with the stochastic networks.

PR #1229, opened 2025-05-02
Add stochastic programming

### LINKED ISSUE #575: Stochastic programming formulation

lkstrp:
Closes #575 
Supersedes #814 
Base/ diff shown to #1154 

tbd

## Checklist

### DIFF
--- a/test/test_scenarios.py
+++ b/test/test_scenarios.py
@@ -1,64 +1,64 @@
-# from collections import Counter
+from collections import Counter
 
-# import numpy as np
-# import pandas as pd
-# import pytest
+import numpy as np
+import pandas as pd
+import pytest
 
-# rng = np.random.default_rng()
-
-
-# @pytest.fixture(params=["ac_dc_network", "ac_dc_network_mi"], scope="function")
-# def n(request):
-#     # Here we return the appropriate fixture based on the parameters.
-#     return request.getfixturevalue(request.param)
+rng = np.random.default_rng()
 
 
-# @pytest.mark.parametrize("add_scenarios", [True, False])
-# def test_xarray_cast(n, add_scenarios):
-#     # ---
-#     # TODO: Only temporary until
-#     # - scenarios are implemented
-#     # - index names are uniform
+@pytest.fixture(params=["ac_dc_network", "ac_dc_network_mi"], scope="function")
+def n(request):
+    # Here we return the appropriate fixture based on the parameters.
+    return request.getfixturevalue(request.param)
 
-#     def cast_to_multiindex(df, level_name, level_values, axis):
-#         current_names = getattr(df, axis).names
-#         new_columns = pd.MultiIndex.from_product([level_values, getattr(df, axis)])
-#         new_columns.names = [level_name] + current_names
-#         new_df = pd.concat([df] * 3, axis=axis, keys=level_values)
-#         setattr(new_df, axis, new_columns)
-#         return new_df
 
-#     n = n.copy()
-#     n.snapshots.name = "snapshot"
-#     c = n.components.generators
+@pytest.mark.parametrize("add_scenarios", [True, False])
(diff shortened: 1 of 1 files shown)


FabianHofmann:
@lkstrp @Irieo the pypsa-eur tests are failing due to the change in the index name  "Bus" -> "name". there is no chance to fix that here. should we create a new branch in pypsa-eur where we adjust the code and run the ci on that? or ignore it?

lkstrp:
We need to do that anyway, otherwise we won't be able to test any pre-releases. So I guess it doesn't hurt to do that already. Do you think we can make PyPSA-Eur compatible with both `>=1.0` and `<1.0`? Also note that it is currently still `'component'`, I'll add a commit to rename it again to `'name'`

FabianHofmann:
> Do you think we can make PyPSA-Eur compatible with both >=1.0 and <1.0?

I think that would be good. the question is how can we differentiate the pre-release version in pypsa-eur. did you already git tag alphas for v1.0 so we have the specifier in `pypsa.__version__` somewhere?

lkstrp:
We don't have a tag yet, and I don't really want to add one before we're on the master branch. Otherwise, we'll have different commit histories (since it's not squashed here). 

But we can use something like ` bool(re.match("^0\.35\.0\.post1\.dev\d{3}" pypsa.version))`. This is not ideal, but it will match only this branch (and any other branch with distance >= 100 to 0.35), but not the master branch. We can change it to a proper `pypsa.__version_semver_tuple__ >= (1, 0, 0)` once we have the tag.

FabianHofmann:
:rocket:

PR #1242, opened 2025-05-30
Identify buses (e.g. bus0, bus1, bus2, ...) by regex

SpamAndEgg:

In my PyPSA workflow I use custom component parameters for postprocessing or custom constraints. I encountered a problem when adding many links with a custom parameter name containing the string "bus", e.g.  `carbon_emission_by_bus`. 

The problem lies in the function `get_strongly_meshed_buses`, where parameter names are checked for the keyword "bus" like this: `c.static.filter(like="bus")`. This will hold true for all parameter names that have "bus" somewhere in the name. 
https://github.com/PyPSA/PyPSA/blob/407950422b7b1986f44141167aaf1ea2c3c43ce7/pypsa/optimization/common.py#L60

If now there are more than 45 instances (which is the current threshold) of a parameter with "bus" in it with the same value, e.g. 46+ times `carbon_emission_by_bus=0.0`, then `0.0` will be identified as a bus name. This later on leads to an error. 

Here is an example of failing code:
```
from pypsa import Network
import pandas as pd

N_OUTPUT_BUSES = 50


def main():
    snapshots = pd.date_range("2025-01-01", freq="h", periods=5)
    n = Network(snapshots=snapshots)
    n.add("Carrier", "dummy_carrier")

    for i_bus in range(N_OUTPUT_BUSES):
        bus_input = f"bus_input_{i_bus}"
        n.add("Bus", name=bus_input, carrier="dummy_carrier")
        n.add("Generator", name=f"grid_{i_bus}", bus=bus_input, p_nom=N_OUTPUT_BUSES, carrier="dummy_carrier")
        bus_output = f"bus_output_{i_bus}"
        n.add("Bus", name=bus_output)
        n.add(
            "Link",
            name=f"link_{i_bus}",
            bus0=bus_input,
            bus1=bus_output,
            carbon_emission_by_bus=0.0,
            p_nom=1,
            carrier="dummy_carrier",
        )
        n.add(
            "Generator",
            name=f"market_{i_bus}",
            bus=bus_output,
            p_min_pu=-1,
            marginal_cost=1,
            p_nom=1,
        )

    n.optimize()


if __name__ == "__main__":
    main()
```

I propose to change checks like `like` or `startswith` with regex checks that only match parameter names that start with "bus" and then have digits after that (depending on the check, a digit might be required or not)

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,11 +73,12 @@ jobs:
     - name: Install package and dependencies
       run: |
         python -m pip install uv
-        uv pip install --compile --system "$(ls dist/*.whl)[dev,hdf5,excel,cartopy]"
+        uv sync --all-extras --no-install-project
+        uv pip install "$(ls dist/*.whl)"
 
     - name: Run unit tests
       run: |
-        pytest --mpl --cov=pypsa && coverage xml
+        uv run pytest --mpl --cov=pypsa && uv run coverage xml
 
     - name: Upload code coverage report
       uses: codecov/codecov-action@v5
@@ -108,13 +109,13 @@ jobs:
     - name: Install package and dependencies
       run: |
         python -m pip install uv
-        uv pip install --system nbmake requests
-        uv pip install --system "$(ls dist/*.whl)[dev, cartopy, docs]"
+        uv sync --all-extras --no-install-project
+        uv pip install "$(ls dist/*.whl)"
 
     - name: Test docs build
       run: |
         sudo apt-get install -y pandoc
-        pytest test/test_docs.py --test-docs-build
+        uv run pytest test/test_docs.py --test-docs-build
 
   check-types:
     name: Check types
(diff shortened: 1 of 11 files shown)


lkstrp (commented):
Hey @SpamAndEgg,
thanks for the PR. This is indeed better.

Could we use a compiled regex and define the patterns only once? It would be bit cleaner even if we stop using the deprecated methods soon.

Something like
```
PORTS_PATTERN = re.compile(r"^bus\d*$")
ADDITIONAL_PORTS_PATTERN = ..
```
within `constants.py`

SpamAndEgg:
Hey @lkstrp, thanks for the review, I have implemented your suggestions. `filter(regex=...)` does not allow a compiled regex as an input, thus the regex string is given. 

I am a bit confused about the mypy failures, they don't seem to be related to my changes. 

Is a release_notes.rst expected from me? Or is it fine as changes are small and not noticable for the user?

lkstrp:
Okay, thank you!
`mypy` failures are not related and come from the master branch because of some dependency updates. I will fix those. The release notes are also not needed.

PR #1249, opened 2025-06-10
chore: expose dynamic network propertys

lkstrp:
- Uses internal properties for network data attribute
   -  `n.model`/`n.objective`/`n.objective_constant`/`n.pypsa_version` are read-only without a setter
   - `n.name`is also wrapped around but can be set
   - This is one more step to a `NetworkData` class which will separate logic from data, which will make things more stable
- Allows to expose them to API/ docs and they are not just hidden attributes
- I guess there is no use case why someone would need to set objective or objective_constant manually?
- Removes `io` backwards compatibility for `v0.13.1` and `v0.18.0`

## Checklist

### DIFF
--- a/pypsa/networks.py
+++ b/pypsa/networks.py
@@ -550,6 +550,17 @@ class Network(
             raise ValueError(msg)
         return self._objective
 
+    @objective.setter
+    def objective(self, new: float) -> None:
+        """Set the objective value of the network."""
+        warnings.warn(
+            "Setting the objective value via `n.objective = ...` is deprecated in 0.35 "
+            "and will be removed in 1.0. Use `n.model.objective.value = ...` instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        self._objective = new
+
     @property
     def objective_constant(self) -> float:
         """Objective constant of the network.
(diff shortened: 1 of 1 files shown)


fneum:
All good about exposure of network attributes.

> This is one more step to a NetworkData class which will separate logic from data, which will make things more stable.

Not quite sure what a NetworkData class would improve. Let's discuss when I'm back.

> I guess there is no use case why someone would need to set objective or objective_constant manually?

No, it wouldn't be good style.

> Removes io backwards compatibility for v0.13.1 and v0.18.0

Does it hurt much to keep the compatibility? Would allow import of archived networks with newer PyPSA versions.

lkstrp:
> Does it hurt much to keep the compatibility? Would allow import of archived networks with newer PyPSA versions.

It does not, I added `0.18.0` again. `0.13.1` is only relevant for HDF5, too. We don't test on old network versions, so we can't be sure if this still works with the current dependencies. With this release, we save the nc and csv versions for every release, so we can test them from here. But if there is a need to load older versions from an archive, we should add tests for those as well

Irieo:
(@lkstrp I’m afraid I’m thinking about this suggestion without enough context.) That said, in general there are plenty of use cases where it’s useful to work with the objective term (and more broadly n.model) manually:

- You might want to solve an optimization problem that isn’t (yet) supported by PyPSA, but still benefit from everything PyPSA provides. For example, a stochastic problem: https://resilient-project.github.io/pypsa-workshop-202505/pypsa-stochastic.html

- You may want to implement a specific subclass of stochastic problems. For example, assigning higher weights to costly branches for risk aversion. In such cases, you would need to formulate your own objective.

- In other cases, users might rely on PyPSA to formulate the objective for a complex, multi-component system, but then patch it. e.g. by appending it with a penalty term for res curtailment.

- similar argument extends to constraints in n.model

PR #1257, opened 2025-06-20
refactor: remove `v1.0.0` deprecations

lkstrp:
- Removes all deprecations to be removed in `v1.0.0`


fneum (approved):
Looks reasonable!

I have a small petition to keep `n.plot()` and `n.explore()` on the top level for quick access as aliases. What do you think about that?

You could also do `n.apply_line_types()` and `n.apply_transformer_types()`.

Since quite a few functions are (re-)moved, there should be a migration guide for v1.0 documentation. Since we have been relatively quick to introduce and remove deprecations, people may miss deprecation warnings if they skip recent versions (e.g., v0.35). A good hint would be to upgrade to 0.35 first and address all deprecations.

PR #1250, opened 2025-06-11
`v1.0.0` Documentation 

lkstrp:
This branch is the base branch for all version 1.0 documentation PRs. Do not commit directly on it, since it will be rebased on the main branch when merging. Create a PR with `v1-docs` as base instead.

## How to contribute

1. Checkout `v1-docs` branch (`git checkout v1-docs`)
2. Create a new branch for your changes (`git switch -c 'docs-changes'`)
3. Run `uv sync --all-extras` to install the dependencies
4. Run `mkdocs serve` to see docs locally at `http://127.0.0.1:8000/` with live reload
5. Commit changes, push to remote and create a PR with `v1-docs` as base

## Todos

- [ ] Resolve all `#TODO` comments
- [ ] api: add optimization expressions
- [ ] api: add network collection
- [ ] api: make internal what should not be exposed
- [ ] user guide: create user guide for all api sections with corresponding reference
- [ ] change all `.rst` references to `.md`
- [ ] change all `.rst` callouts to `.md` (also `> **Note**`)
- [ ] Add caption to all images
- [ ] Add broken links checker
- [ ] doctests for user guide
- [ ] Close all open documentation issues: see [here](https://github.com/PyPSA/PyPSA/issues?q=type%3A%22Docs%22%20state%3Aopen)

PR #1259, opened 2025-06-22
feat: add warnings for typos and setting unintended attributes

lkstrp:
There have been several examples where attributes were added to a network with the belief that they would do something, but they did not. Either because the specific default attribute did not exist for the component and only for others or because of a typo.

Although the opt-in components class could help here, this PR adds additional warnings for these cases.

For typos:

```python
>>> n.add("Generator", "my-gen", bus="bus1", comittable=True)
WARNING:pypsa.network.transform:Attribute 'comittable' is likely a typo of standard attribute 'committable' of generators. Please check the attribute name. See https://link.pypsa.org/warning-attr-typo.
```

For unintended custom attributes:
```python
n.add("Store", "my-store", p_nom=100)
WARNING:pypsa.network.transform:The attributes ['p_nom'] are default attributes for other components, but not for stores. This is most likely unintended and these attributes should be renamed. See https://link.pypsa.org/warning-attr-misleading.
```

Also pinging @Irieo to add the discussed docs changes 

## Checklist


Irieo:
Very useful feature. We have caught ourselves making silly things often without this system, and indeed many folks were asking for something in this direction.

Some quick feedback:

- There are cases that are not captured by warnings, although looking through the code I feel like they should be: n.add("Generator", "gen2", bus="London", P_nom=100) -> distance is 1. Why does it not trigger a warning?

- There are cases of plausible (yet wrong) attribute names with higher distance. I see there is a trade-off between letting folks create custom attributes for pandas operations and stuff like this...
n.add("Load", "load1", bus="London", demand=50) -> should be p_set
n.add("Generator", "gen1", bus="London", p_nominal=100) -> should be p_nom
n.add("Generator", "new5", bus="London", p_nom=100, unit_commitment=True) -> should be committable

- Levenshtein is now a PyPSA dependency, right?

lkstrp:
> * There are cases that are not captured by warnings, although looking through the code I feel like they should be: n.add("Generator", "gen2", bus="London", P_nom=100) -> distance is 1. Why does it not trigger a warning?

Because with a different "base" variable it is also distance 1. `f_nom` and `s_nom` for transformers etc. But now when I think about it we can also add a check with the total sets of attributes for the typo check. Let me adjust that.

> * There are cases of plausible (yet wrong) attribute names with higher distance. I see there is a trade-off between letting folks create custom attributes for pandas operations and stuff like this...
>       n.add("Load", "load1", bus="London", demand=50) -> should be p_set
>       n.add("Generator", "gen1", bus="London", p_nominal=100) -> should be p_nom
>       n.add("Generator", "new5", bus="London", p_nom=100, unit_commitment=True) -> should be committable

If you provide me a list of those, we can easily add that. That would just be a simple look up. I'll start here, just edit the comment.
Alias warnings:
`committable` <- `unit_commitment`
`p_set` <- `demand` 
`x_nom` <- `x_nominal`

> * Levenshtein is now a PyPSA dependency, right?

Yeah there is no way around that. But written in C and very small (12kb).

Irieo:
Regarding the ping above: required doc change we discussed  (no p_nom for Store) is added to the new docs branch
https://github.com/PyPSA/PyPSA/pull/1250/commits/8227dd6632903f85fe7855a8a01e50802f88b640

PR #1269, opened 2025-07-03
Add additional MGA functionality and example

koen-vg:
Introduces new functions in optimization/abstract.py to facilitate the analysis of trade-offs between multiple objectives/dimensions using near-optimal methods. Allows for solving a network in a direction given in the coordinate space of user-specified dimensions, and also introduces a parallelized function to solve in multiple directions at once. Three different functions generating (random) directions are also provided. Finally, an example notebook showcasing the new functionality is included.

I might still add an additional commit with some more documentation other than the example notebook. (Release note also outstanding.)

Still need to add a unit test, but I thought I'd submit this as WIP so others can have a look and comment.

## Checklist

- [-] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.


lkstrp:
Hey @koen-vg,

thank you, we will review this after #1153 is merged. Also, please note that we no longer merge the existing docs and only merge in #1250. This is fine for now, since we only have to split the example notebook. But it would probably be better if this were a dedicated section in the user guide, which would require a change in #1250

koen-vg:
> thank you, we will review this after #1153 is merged.

No problem, there's no hurry to get this merged for my part :)

> Also, please note that we no longer merge the existing docs and only merge in #1250. This is fine for now, since we only have to split the example notebook. But it would probably be better if this were a dedicated section in the user guide, which would require a change in #1250

I was thinking about writing something more about MGA in the user guide. Would you prefer a separate PR with #1250 as a base? It would obviously refer extensively to functions introduced in this PR.

lkstrp:
> I was thinking about writing something more about MGA in the user guide. Would you prefer a separate PR with https://github.com/PyPSA/PyPSA/pull/1250 as a base? It would obviously refer extensively to functions introduced in this PR.

That would be great! Contributions on #1250 are very welcome and another PR with #1250 as base is probably the easiest setup.

In general, we don't want examples of PyPSA functionality without a dedicated user guide anymore. So examples should only accompany them (e.g. [Stochastic Optimization - User Guide](https://pypsa--1250.org.readthedocs.build/en/1250/user-guide/stochastic-optimization/#overview) and [Stochastic Optimization - Example](https://pypsa--1250.org.readthedocs.build/en/1250/examples/stochastic-optimization/). In this case, the example could probably be referenced in the existing MGA user guide, along with some additional details (I'm not too sure what we have there right now)

koen-vg:
Okay I've removed the example notebook from this PR (since it's now in #1272), and also added test coverage. Let's see what the tests say

koen-vg:
I'm basically fine with the implementation as it is, but I have to say I'm completely mystified as to why the tests fail. They work on my machine, and I don't see why this ValueError shouldn't be raised?? How does an unsolved network suddenly have an objective?? Or am I misunderstanding something?

This does by the way point to the slightly awkward issue that I don't think there's a particularly good way of checking whether a given network has been solved or not. The `hasattr(n, "objective")` check that is already used in `optimize_mga` is a bit weird, since with the current implementation of the "objective" property, doing this check either returns true or raises a ValueError itself (not the True/False outcome one would expect). Is there a better way? Would it even make sense to add a boolean flag to the Network class (i.e. `solved` / `is_solved` / `solved_p` / `optimized` or something similar) to make this simpler? Though I can see that maintaining this state could be a bit tricky in edge cases.

lkstrp:
Ah yeah sorry, that's because of #1265 and #1249. `n.is_solved` was just merged in #1265, so use that instead

And the PyPSA-DE test is broken right now, so ignore that one

koen-vg:
Okay, great, maybe now we are good.

Actually one last think which I meant to comment on is that I did something slightly hacky which I was wondering if could be improved. Specifically, in the parallelisation in `optimize_mga_in_multiple_directions`, we delete the Highs instance (or other solver instance, as it were) of the Linopy model instance:
```python
if hasattr(self._n.model, "solver_model"):
    del self._n.model.solver_model
```
This is because the Highs object happens to be the only thing that can't be pickled; necessary for the multiprocessing business.

Could it be worth considering doing something about that? If so, what? I don't know much about the topic of pickling/serialisation. Maybe it's also just fine like this.

pypsa/optimization/abstract.py
```diff
@@ -694,11 +821,97 @@ def optimize_with_rolling_horizon(
...
+        n.model.add_constraints(
+            objective + fixed_cost <= (1 + slack) * optimal_cost, name="budget"
+        )
+
     def optimize_mga(
         self,
         snapshots: Sequence | None = None,
         multi_investment_periods: bool = False,
-        weights: dict | None = None,
+        weights: dict | LinearExpression | None = None,
```
fneum: Also add to docstring description.

pypsa/optimization/abstract.py
```diff
@@ -806,42 +1000,201 @@ def optimize_mga(
...
-                return {k: convert_to_dict(v) for k, v in obj.items()}
-            return obj
+        return status, condition
 
-        self._n.meta["weights"] = convert_to_dict(weights)
+    def project_solved(
+        self,
+        dimensions: dict,
+    ) -> dict:
+        """Project solved model onto low-dimensional space."""
```
fneum: Copy over dimensions argument description into docstring.

pypsa/optimization/abstract.py
```diff
@@ -806,42 +1000,201 @@ def optimize_mga(
...
+        slack : float
+            Cost slack for budget constraint. Defaults to 0.05.
+        model_kwargs: dict
+            Keyword arguments used by `linopy.Model`, such as `solver_dir` or
+            `chunk`.
+        **kwargs:
+            Keyword argument used by `linopy.Model.solve`, such as `solver_name`,
+
+        Returns
+        -------
```
fneum: The new (third) returned item is missing in docstring.

pypsa/optimization/abstract.py
```diff
@@ -806,42 +1000,201 @@ def optimize_mga(
...
+        # Build objective as linear combination of direction and
+        # dimensions. Flip the sign in order to maximize in the given
+        # direction.
+        m.objective = -sum(
+            direction[key] * self.build_linexpr_from_weights(dimensions[key], model=m)
+            for key in direction.keys()
+        )
+
+        status, condition = self._n.optimize.solve_model(**kwargs)
+        coordinates = self.project_solved(dimensions) if status == "ok" else None
```
fneum: Up for discussion, but the coordinates could also be returned as pandas.Series, especially since multiple directions return a pandas.DataFrame?

fneum (approved):
Very nice! This makes near-optimal space exploration very accessible. I only found some small gaps in the docstrings.

@lkstrp are you happy with where the functions are placed?

I'm fine with the explicit removal of the highs model for multiprocessing.

pypsa/optimization/abstract.py
koen-vg: Good catch; this was actually a mistake, left from an initial draft which I forgot to clean up. Will remove.

pypsa/optimization/abstract.py
koen-vg: Will add a proper docstring describing argument and return value

pypsa/optimization/abstract.py
koen-vg: Will fix

pypsa/optimization/abstract.py
koen-vg: Sure, I we can let both `project_solved` and `optimize_mga_in_direction` return pd.Series.

fneum:
Good to go! Thanks @koen-vg!

lkstrp:
Thanks @koen-vg ! Will merge as soon as we have #1153 in

pypsa/optimization/abstract.py
```diff
@@ -283,6 +287,129 @@ def optimize_and_run_non_linear_powerflow(
...
+    seed : int | None, optional
+        A seed for the random number generator to ensure reproducibility.
+        If None, a random seed will be used. Defaults to None.
+
+    Returns
+    -------
+    pd.DataFrame
+        A DataFrame containing the generated random directions, where each row
+        represents a direction and each column corresponds to a key from `keys`.
+    """
```
lkstrp: Add "See Also" with link to `optimize_mga..` methods to all docstrings

pypsa/optimization/abstract.py
```diff
@@ -694,6 +821,92 @@ def optimize_with_rolling_horizon(
                 )
         return self._n
 
+    def build_linexpr_from_weights(
```
lkstrp: If the user will never want to use this (?), this should be internal
```suggestion
    def _build_linexpr_from_weights(
```

pypsa/optimization/abstract.py
```diff
@@ -806,42 +1000,225 @@ def optimize_mga(
...
-                return obj.to_dict(orient="list")
-            if isinstance(obj, pd.Series):
-                return obj.to_dict()
-            if isinstance(obj, dict):
-                return {k: convert_to_dict(v) for k, v in obj.items()}
-            return obj
+        return status, condition
 
-        self._n.meta["weights"] = convert_to_dict(weights)
+    def project_solved(
```
lkstrp: Same here, let's make it internal if the user should not use it directly

```suggestion
    def _project_solved(
```

lkstrp (commented):
Ok, I take back my last comment. Let's merge this now, otherwise, the docs dev will become too chaotic, and I'll resolve the merge conflicts in #1154.

Some minor things:
- There are a couple of MGA methods now. I think we want to move them to pypsa/optimization/mga.py and take them out of the `abstract.py`.
- I'm not sure if all methods need to be exposed. Anything that a user should not use should be internal (e.g. '_function_name'). See review comments
- The generate directions helper should be available via `pypsa.optimization.generate_directions` and linked to in the docstrings

I can also apply those changes quickly if you want @koen-vg

koen-vg:
>     * There are a couple of MGA methods now. I think we want to move them to pypsa/optimization/mga.py and take them out of the `abstract.py`.
Sounds good to me!

>     * I'm not sure if all methods need to be exposed. Anything that a user should not use should be internal (e.g. '_function_name'). See review comments
The `build_linexpr_from_weights` and `project_solved` functions aren't exactly user-facing, but I can imagine some situations where users might want to use them, though I guess 95% of the time the main mga functions are good enough. But to keep things simple I think it's okay to make these internal, and in the 5% case the user can find or copy these internal functions as needed anyway; they aren't too complicated either.

>     * The generate directions helper should be available via `pypsa.optimization.generate_directions` and linked to in the docstrings
> 
> 
> I can also apply those changes quickly if you want @koen-vg
If you want to and have the time, that would be great. Thanks for the review! If you won't get around to it, I can probably do it in a couple of days.

pypsa/optimization/abstract.py
```diff
@@ -806,42 +1000,225 @@ def optimize_mga(
...
+        snapshots: Sequence | None = None,
+        multi_investment_periods: bool = False,
+        slack: float = 0.05,
+        model_kwargs: dict | None = None,
+        max_parallel: int = 4,
+        **kwargs: Any,
+    ) -> tuple[pd.DataFrame, pd.DataFrame]:
+        # Remove solver model stored by linopy if present; it cannot be pickled
+        if hasattr(self._n.model, "solver_model"):
+            del self._n.model.solver_model
```
koen-vg: I just noticed that linopy 0.5.3 supports pickling models. So maybe if we increase the minimum linopy version, we won't need this slightly hacky trick anymore?

lkstrp: But a pickle can't be put in an `.nc` PyPSA file. We need some general improved io features for that

koen-vg: Just in case I quickly tested this, but linopy 0.5.3 indeed doesn't make this redundant; it's still needed.

lkstrp:
Hey @koen-vg,

I merged diff back and moved all the methods to `mga.py`. Could I ask you to have a look again and fix the last issues?

koen-vg:
Alright here you go! I did a little more than was strictly speaking necessary to get this running, trying to add some guardrails around the parallel code to hopefully handle errors and hanging processes better. Not extensively tested under adverse conditions, but better safe than sorry. Was also missing a docstring. Now at least ruff is also happy about this I think.

Parallelism is now handled to writing the network to a temp file and re-importing in worker processes. Seems the easiest. In order to use the Network(fn) function I had to import pypsa.Network in the middle of the script somewhere; at the top this leads to circular imports. Ruff didn't like it so I added an exception. But if you know of a better way of getting access to the Network() function with a top-level import and without circular imports, feel free to implement that are remove the ruff ignore.

koen-vg:
Hmm I don't know what's wrong with the tests, maybe it was too ambitious to include an example if I don't know what to do to make it work.

lkstrp:
It's never too ambitious to have good docstrings!

Example sections in the docstrings are tested using doctests. This means that the written output must match the actual output. I also replaced it with ac_dc_meshed, which is a global so it doesn't need to be defined again runs quicker.

In general, to avoid circular imports, import from the module rather than the package (e.g. `from pypsa.networks import Network`). In this case, that is not enough and we need (future) structural changes

Thanks a lot @koen-vg !

Issue #1287, opened 2025-07-10
Test statistics functionality with multiple investment periods


fneum:
- add unit tests
- add functionality to group by investment period

nick-gorman:
Hi @fneum,

Following up on the issues and pull requests (#1200, #1247, #1289) which raise the issue of snapshot weightings not being applied properly in the statistics. 

I think a similar issue is occurring with the investment period weightings.

In the example below, there are investment periods of uneven length, something which we can account for by giving a greater weight to the longer investment period (2021). The weighting is reflected in the objective function but not the n.statistics.capex()

pypsa=0.35.1
python=3.12

```Python
import io

import pandas as pd
import pypsa

def csv_str_to_df(csv_str):
    """Helper function to convert a CSV string to a DataFrame."""
    # Remove spaces and tabs that have been included for readability.
    csv_str = csv_str.replace(" ", "").replace("\t", "").replace("__", " ")
    return pd.read_csv(io.StringIO(csv_str))


snapshots = """
period, timestep
2020,   2020-01-01__00:30:00
2021,   2021-01-01__00:30:00
2021,   2022-01-01__00:30:00
2023,   2023-01-01__00:30:00
"""
snapshots = csv_str_to_df(snapshots)
snapshots = snapshots.set_index(["period", "timestep"])

load_var = pd.Series(
    [0, 100, 100, 100],
    index=snapshots.index,
    name="load"
)

n = pypsa.Network()
n.snapshots = snapshots.index
n.investment_periods = [2020, 2021, 2023]

# Account for the fact that the 2021 snapshot is 2 years long.
n.investment_period_weightings.at[2021, "objective"] = 2.0
n.investment_period_weightings.at[2021, "years"] = 2.0

n.add("Bus", "bus_one")

n.add(
    "Load",
    "load_one",
    bus="bus_one",
    p_set=load_var,

)

n.add(
    "Generator",
    "gen_one",
    bus="bus_one",
    p_nom=0,
    p_nom_extendable=True,
    capital_cost=1,
    marginal_cost=0,
    build_year=2021,
)

n.optimize(multi_investment_periods=True)

print(n.objective)
# 300.0

print(n.investment_period_weightings)
#         objective  years
# period
# 2020          1.0    1.0
# 2021          2.0    2.0
# 2023          1.0    1.0

print(n.statistics.capex())
#                    2020   2021   2023
# component carrier
# Generator -         NaN  100.0  100.0
```

PR #1289, opened 2025-07-11
Bugfix: correct use of snapshot weighting columns in statistics

fneum:
The use of weightings in the optimisation module was correct.

The use of weightings in the statistics module was **not** correct.

The misconception was that the "stores" and "generator" weightings are not component-specific.

Rules:

- You want to calculate some cost? Take `objective` weightings.
- You want to calculate storage state of charge? Take `stores` weightings.
- You want to calculate total energy balances? Take `generators` weightings.

The initial naming choice was maybe not ideal.


lkstrp:
Great @fneum, that's a minor change if it resolves the issue/

As just discussed with @FabianHofmann and @Irieo:

Since this took a while now to find out, the naming choice is not ideal and the vast majority of users will only use the same weight in all columns, I propose a backwards-compatible simplification. Also since we have the same confusion with `n.investment_period_weightings` and we wanna align with the upcoming `scenario_weightings`.

With `1.0`, the default entry for all of `n.snapshot_weightings`, `n.investment_period_weightings` and (new) `n.scenario_weightings` is a single `pd.Series` with the name `weight`. If you then want to distinguish between each weight column (as possible right now), you can set them via `n.set_xx` to get the current structure. But only if you pass different weights for each column. This can be fully backwards compatible with old networks. Any objections to that?

lkstrp:
Follow up on the mentioned points above:
- Mixing `pd.Series` with `pd.DataFrame` could lead to issues, also there are potential future developments which could need multiple scenario weights for `n.scenarios`
- So we keep the `pd.DataFrame`'s as they are and improve docs instead to make this more clear

I'll merge but we also need to fix #1287

PR #1294, opened 2025-07-24
Deprecate define_nominal_constraints_per_bus_carrier

fneum:
Suggestion to deprecate this function as it has the same functionality as the global constraint of type 'tech_capacity_expansion_limit'. There should be one recommended way to set such constraints up and they are most relevant for networks with multiple investment periods.


pypsa/optimization/global_constraints.py
```diff
@@ -85,6 +86,11 @@ def define_tech_capacity_expansion_limit(n: Network, sns: Sequence) -> None:
             )
 
 
+@deprecated(
+    deprecated_in="1.0",
+    removed_in="1.1",
```
lkstrp: We wanna follow semver now, which means everything what is in `1.0` needs to stay until `2.0`
```suggestion
    removed_in="2.0",
```

lkstrp (commented):
This also needs to be replaced in `create_model` then as well

fneum:
> This also needs to be replaced in create_model then as well

Yes, but it stays until 2.0 then, doesn't it?

Ah, but you want to pull the check whether you actually need this function out of the function itself to avoid unnecessary deprecation warnings.

lkstrp:
Ah, so you mean the same functionality via a different implementation. In that case, we can't achieve the same results based on `define_tech_capacity_expansion_limit`. So yes, it needs to stay until v2.0, and we need to check whether the constraint is even doing something based on network attributes and only then raise the warning

PR #1295, opened 2025-07-24
Add utility function to compute annuity

fneum:
Add utility function to calculate annuity factor (capital recovery factor) from given discount rate and lifetime. This can be useful for calculating `capital_cost` attributes from technology-data given in upfront cost, lifetime and discount rate; as well as getting back to the upfront cost from annuitized investment costs.

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -17,6 +17,17 @@ Bug Fixes
 * Correct use of snapshot weighting columns in statistics module. The
   doscstring for ``n.snapshot_weightings`` was clarified.
 
+* Added utility function ``pypsa.common.annuity`` to calculate the annuity
+  factor for a given discount rate and lifetime. Also known as capital recovery
+  factor, it is used to convert a capital cost into an annualized cost. The
+  formula is: 
+
+  .. math::
+  
+      \frac{r}{1 - (1 + r)^{-n}}
+
+  where :math:`r` is the discount rate and :math:`n` is the lifetime in years.
+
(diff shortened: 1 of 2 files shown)


euronion:
You want to do this in PyPSA? I have it on the agenda to implement it in `technologydata`.

fneum:
I think it's good to have it in PyPSA, since everyone doing capacity expansion needs this, but only a fraction will use `technology-data`.

fneum:
We would also need it for reconstructing overnight investments from the `capital_cost` attribute:

https://github.com/PyPSA/PyPSA/issues/371#issuecomment-3051888434

euronion:
ok!

doc/references/release-notes.rst
```diff
@@ -17,6 +17,17 @@ Bug Fixes
...
+* Added utility function ``pypsa.common.annuity`` to calculate the annuity
+  factor for a given discount rate and lifetime. Also known as capital recovery
+  factor, it is used to convert a capital cost into an annualized cost. The
+  formula is: 
+
+  .. math::
+  
+      \frac{r}{1 - (1 + r)^{-n}}
+
+  where :math:`r` is the discount rate and :math:`n` is the lifetime in years.
```
lkstrp, line 29: I still have a comment: The details here should rather (also) go in the docstring

fneum: Done!

FabianHofmann:
I guess we could think about a cost module which supports basic interconnection to technology-data or?

PR #1299, opened 2025-07-31
New docs example: Negative prices due to KVL

bobbyxng:
- Added new example in documentation: "Negative locational marginal prices from line congestion in DCOPF"
- This notebook demonstrates how negative locational marginal prices (LMPs) can occur in electricity systems due to line congestion. Using a simple 3-bus linearised DC power flow model, we reproduce the phenomenon implemented in [Kyrie Baker's "3bus_LMP" example](https://github.com/kyribaker/3bus_LMPs). When cheap generation is trapped behind a congested line, LMPs can drop below zero as the system redistributes power flows to meet demand. This behaviour is a direct result of the duality of the DC-OPF problem, where LMPs emerge as the shadow prices of nodal power balance. With this example, we illustrates how network topology, generator costs, and constraints jointly shape prices in modern electricity markets.

## Checklist


bobbyxng:
Thanks @fneum for the quick review and the corrections! Also fixed another small typo.

Irieo:
Nice @bobbyxng, such a great short example! my only comment would only be about
> All lines have the same **susceptance** of x=1

since both in general conventions and in PyPSA standards (`x` /  `x_pu`) stand for reactance.
I suppose Kyrie Baker example uses susceptance as it relies on B*Theta formulation for linear power flow

Irieo (approved):

bobbyxng:
@Irieo Thanks! You are right with both, I will correct it to reactance, as PyPSA uses x/x_pu instead of the susceptance. With X=1=-1/B, the results are the same though :)

juliusmeier:
Thank you adding this great example!
@bobbyxng , do you know why pypsa does not produce the exact same LMPs for the case on the right in Kyri Baker's example?
I re-built it a while ago with pypsa but obtained different LMPs than she did. 

For clarity, I am refering to this case in her example:
image

bobbyxng:
@juliusmeier : Very interesting, that made me also test the second example both in PyPSA but also in Kyri Baker's notebook. I did three runs, 1x PyPSA and 2x Kyri Baker's notebook with different solvers. All three runs yield the same primal optimal solution, but indeed very different marginal costs. Even when using `highs` solver once in the PyPSA formulation and once in the original notebook, I get the same primal solutions but different dual values.

So very likely, the primal solution is unique, but the dual solution clearly is not, likely because the dual problem is degenerate, i.e., multiple solutions satisfy the dual problem.

PR #1307, opened 2025-08-02
make start-up and shut-down variables continuous

### LINKED ISSUE #1303: Define start-up and shut-down variables as continuous
The start-up and shut-down variables are defined as binary variables and calculated from the difference of two adjacent status variables, which are also binary. It would be sufficient to define the start-up and shut-down variables as continuous, as they are bound to integer variables through the status variables. This would help the branch-and-bound algorithm that solves the MILP, as it has fewer integer variables to consider.

fneum:
Start up and shut down variables are now initialised as continuous variables
  with bounds [0,1] instead of as binary variables. They can already only take
  the values 0 or 1 as they depend on the difference of the binary status
  variables of adjacent snapshots. This should speed up the branch-and-bound
  algorithm as it reduces the number of binary variables in the model.

closes #1303

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -11,6 +11,12 @@ Release Notes
 ..    To use the features already you have to install the ``master`` branch, e.g. 
 ..    ``pip install git+https://github.com/pypsa/pypsa``.
 
+* Start up and shut down variables are now initialised as continuous variables
+  with bounds [0,1] instead of as binary variables. They can already only take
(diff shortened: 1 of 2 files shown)


fneum:
Should only be a mini merge conflict with https://github.com/PyPSA/PyPSA/pull/1154

Irieo:
Did you test that this actually speeds up the branch-and-bound alg?

fneum:
No, I did not (yet). Just theory so far.

PR #1312, opened 2025-08-04
Extending n.explore()

bobbyxng:
- Improving `n.explore()` and extending its functionalities with interactive plots
- Changing backend from folium to pydeck
- Work in progress (draft)

## Checklist

### DIFF
--- a/examples/notebooks/explore.ipynb
+++ b/examples/notebooks/explore.ipynb
@@ -1,0 +1,33 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "8ff5544d",
+   "metadata": {},
+   "source": [
+    "# Interactive exploration of networks"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a103fda6",
+   "metadata": {},
+   "source": [
+    "This notebook demonstrates the interactive exploration capabilities of PyPSA networks."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "30dff83f",
+   "metadata": {},
+   "source": []
+  }
+ ],
+ "metadata": {
+  "language_info": {
+   "name": "python"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 5
+}


bobbyxng:
Update: Apart from generating pie charts based on multi-index pd.Series, all functionality from static n.plot() are now implemented in n.explore, with identical API wherever applicable:

```
Create an interactive map of the PyPSA network using Pydeck.

    Parameters
    ----------
    n : Network
        The PyPSA network to plot.
    bus_sizes : float/dict/pandas.Series
        Sizes of bus points in meters, defaults to 5000.
    bus_colors : str/dict/pandas.Series
        Colors for the buses, defaults to "cadetblue".
    bus_alpha : float/dict/pandas.Series
        Add alpha channel to buses, defaults to 0.7.
    bus_columns : list, default None
        List of bus columns to include. If None, only the bus index and x, y coordinates are used.
        Specify additional columns to include in the tooltip.
    line_flow : float/dict/pandas.Series/None
        Series of line flows indexed by line names, defaults to None. If None, no arrows will be created.
        If a float is provided, it will be used as a constant flow for all lines.
    line_colors : str/dict/pandas.Series
        Colors for the lines, defaults to 'rosybrown'.
    line_alpha : float/dict/pandas.Series
        Add alpha channel to lines, defaults to 0.7.
    line_widths : float/dict/pandas.Series
        Widths of lines in meters, defaults to 1500.
    line_columns : list, default None
        List of line columns to include. If None, only the bus0 and bus1 columns are used.
        Specify additional columns to include in the tooltip.
    link_flow : float/dict/pandas.Series/None
        Series of link flows indexed by link names, defaults to None. If None, no arrows will be created.
        If a float is provided, it will be used as a constant flow for all links.
    link_colors : str/dict/pandas.Series
        Colors for the links, defaults to 'darkseagreen'.
    link_alpha : float/dict/pandas.Series
        Add alpha channel to links, defaults to 0.7.
    link_widths : float/dict/pandas.Series
        Widths of links in meters, defaults to 1500.
    link_columns : list, default None
        List of link columns to include. If None, only the bus0 and bus1 columns are used.
        Specify additional columns to include in the tooltip.
    transformer_flow : float/dict/pandas.Series/None
        Series of transformer flows indexed by transformer names, defaults to None. If None, no arrows will be created.
        If a float is provided, it will be used as a constant flow for all transformers.
    transformer_colors : str/dict/pandas.Series
        Colors for the transformers, defaults to 'orange'.
    transformer_alpha : float/dict/pandas.Series
        Add alpha channel to transformers, defaults to 0.7.
    transformer_widths : float/dict/pandas.Series
        Widths of transformers in meters, defaults to 1500.
    transformer_columns : list, default None
        List of transformer columns to include. If None, only the bus0 and bus1 columns are used.
        Specify additional columns to include in the tooltip.
    arrow_size_factor : float, default 2.5
        Factor to scale the arrow size in relation to line_flow. A value of 1 denotes a multiplier of 1xline_width.
    arrow_colors : str/dict/pandas.Series | None, default None
        Colors for the arrows. If not specified, defaults to the same colors as the respective branch component.
    arrow_alpha : float/dict/pandas.Series, default 1.0
        Add alpha channel to arrows, defaults to 1.0.
    map_style : str
        Map style to use for the plot. One of 'light', 'dark', 'road', 'satellite', 'dark_no_labels', and 'light_no_labels'.
    tooltip : bool, default True
        Whether to add a tooltip to the bus layer.

    Returns
    -------
    pdk.Deck
        The interactive map as a Pydeck Deck object.
```

Currently, n.explore() would directly return a functioning pydeck object which could be directly manipulated using pydecks built-in methods. It also allows for stacking with other already generated pydeck object, very similar to how matplotlib.pyplot handles stacking.

However, by returning a pydeck object directly, subsequent manipulations to the generated object can only be done on the pydeck object level, not on the PydeckPlotter class (which is introduced in pypsa.plot.maps.interactive).

For example, not implemented though, the PydeckPlotter class can have additional properties like `buses_without_coordinates`, `branch_components_dropped`, etc. (tbd. whether this is useful at all).

Long story short, here are the options I see:
a) Keep as it is now, only use PydeckPlotter class for structuring methods, return a pydeck object with n.explore()
b) Return a PydeckPlotter object with n.explore(). Then a user would do the following operation to get the pydeck object: `map = n.explore()`, followed by `map.deck()`
c) PydeckPlotter inherits from pydeck class and is extended by additional methods and properties. Requires some refactoring of code.

Do you have any thoughts @lkstrp, @fneum, @FabianHofmann ?

bobbyxng:
Sneak preview:
image

pypsa/common.py
bobbyxng, line 1: @lkstrp Can you check if this is fine or needed at all?
If removed here, I will add `.reindex()` in the PydeckPlotter methods where needed.

bobbyxng:
I attached a simple notebook using the SciGrid network. There you can try out `n.explore()`

image

pypsa/common.py
bobbyxng: Reverted to old function, as other functions depend on the old behaviour (issues arise with multi-indices)

bobbyxng:
- Fixed bug where arrow heads pointed into wrong direction and where offset when the flow was negative
- Merged arrow_layer and branch_layer creation into one function, as they depend on the same attributes.

bobbyxng:
- Added initial Piechart functionality including `bus_split_circles=True`
- See notebook https://github.com/PyPSA/PyPSA/blob/new-explore/examples/notebooks/explore.ipynb for a minimal example
- Ready for (initial) review @fneum @lkstrp :)
- There are probably still some places in the code with refactoring/speed-up potential

## Screenshot of notebook output
image

bobbyxng:
Static equivalent (note the differences in semi circle scaling, open to do):
image

~~Edit: Noticed a bug, negative values are not colored correctly.~~ Fixed with https://github.com/PyPSA/PyPSA/pull/1312/commits/525e7dcb11f2c42db64d952eab776be852166ba9
- [x] ~~Fix coloring of negative semi circles~~

bobbyxng:
With https://github.com/PyPSA/PyPSA/pull/1312/commits/525e7dcb11f2c42db64d952eab776be852166ba9:
-  Bugfix: Correct mapping of color indexes with bus_split_circles. 
- Refactored, simplified code for bus_split_circles handling.

fneum:
Excellent work! This will be extremely useful to many people!

Here are some suggestions that should be quick to implement.

- Example: Update bus size factors and carrier colors.
- Is there a way to separate the hover information by component (buses, lines, links, transformers)?
- For mapping of all relevant keyword arguments of `n.plot()`, I would add `branch_components`, `jitter` and the `*_cmap` and `*_norm` and `geomap` arguments (see below).
- Default to `map_style='road'`? Looks nice.
- Default to higher `*_alpha=0.9`.
- Possible to have an interactive plot without a map? I.e. `geomap=False`?
- Support for float values in `*_colors`, e.g. `n.explore(line_colors=n.lines.s_nom, line_cmap='viridis')`
- No warning is thrown if unknown function arguments are passed (e.g. `mapstyle` instead of `map_style`.
- `map_style='satellite'` did not work for me.
- Speed in the browser is fast, but slow in Jupyter / VS Code. Do you know why?
- Perhaps give `bus_sizes` in square kilometres or just by radius in meters to avoid big numbers that square meters give?
- Documentation in `v1-docs`. API Lukas will do.
- We should probably keep the `folium` variant somewhere (`n.explore(backend='folium')` or so?) so this PR is not hard breaking?

Adding new layers to the `Deck` object works well!

```py 
import pydeck as pdk
import geopandas as gpd
import json

map = n.explore(map_style='road', line_alpha=0.9, bus_alpha=0.9, bus_sizes=1e6)

fn = "/home/fneum/pr/pypsa-eur/resources/distribution-grid-experimental/regions_onshore.geojson"

gdf = gpd.read_file(fn).query("country in ['GB', 'IE']")
geojson = json.loads(gdf.to_json())

layer = pdk.Layer(
    "GeoJsonLayer",
    geojson,
    pickable=True,
    stroked=True,
    filled=True,
    get_fill_color=[100, 150, 255, 50],
    get_line_color=[0, 0, 0, 200],
    get_line_width=200,
    autoHighlight=True,
)

map.layers = [layer, *map.layers]

map.to_html("map.html", offline=True);
```

pypsa/plot/maps/interactive.py
```diff
@@ -323,279 +320,875 @@ def iplot(
...
         )
-        components_present.append("Link")
 
-    if not n.c.buses.static.empty and "Bus" in components:
-        gdf_buses = gpd.GeoDataFrame(
-            n.c.buses.static,
-            geometry=gpd.points_from_xy(n.c.buses.static.x, n.c.buses.static.y),
-            crs=crs,
+        # Map bus sizes
+        bus_data["radius"] = bus_data["value"] ** 0.5
```
fneum: It could be dangerous if using the rounded values from above (for some edge cases with very low values).

pypsa/plot/maps/interactive.py
```diff
@@ -323,279 +320,875 @@ def iplot(
...
+            "PathLayer",
+            data=c_data.reset_index(),
+            get_path="path",
+            get_width="width",
+            get_color="rgba",
+            pickable=True,
+            auto_highlight=True,
+            parameters={
+                "depthTest": False
+            },  # To prevent z-fighting issues/flickering in 3D space
```
fneum, line 1187: Will it be possible here to take a `geometry` column for lines?

fneum (changes requested):
I'll let @lkstrp judge the code quality. Looks good from a quick glance!

bobbyxng:
@fneum Thank you very much for your review and comments: Some quick replies to a few of your points (generally we should be able to address all of them):
- For different hover informations ("tooltips" in pydeck/deck.gl) per layer/component: Unfortunately pydeck only supports a single tooltip definition for all layers. This means a value "undefined" will be shown in case the column selected does not exist in the PyPSA component. I have found a workaround on the official deck.gl repo, which I will experiment with: https://github.com/visgl/deck.gl/issues/6799 . The workaround specifies a common "tooltip" column which contains individually defined html tables. Note that with this workaround the filesize will increase as values shown in the tooltip are basically formatted duplicates of values already encoded in the embedded json (duplicates only for geometry information).
- `geomap=False`: Probably also just as a workaround by specifying a map_style that does not exist. In addition to this, I wanted to add a toggle for PyPSA's country shapes as optional overlay
- regarding keeping folium alive: my personal preference would be to remove it entirely, as it was the original test and by far does not support all the features that are implemented with the pydeck-based approach, many which are not even possible to implement into folium (not GPU accelerated, no scaling of elements with zoom level, etc.). I don't think it will be a critical breaking change, as most people likely only ran `n.explore()`, which still works only returning a different object. But I am also fine with keeping it as a legacy backend. :) 
- Speed differences: Yes I reproduced this on Ubuntu 24.04 LTS in combination with latest VS Code and Jupyter Notebook extension. I could not manage to fix the issue. My current understanding is that somehow the (i)GPU of some chips are not correctly passed/enabled in VS Code, so it's an issue outside of our hands. Any extension that is GPU "heavy" will run into the same issues (e.g., Geo Data Viewer). As soon as you change one of the two, either Ubuntu 24.04 LTS (to 22.04 LTS) or switch to a jupyter notebook environment outside of VS Code, everything is as fast as in the exported html file. At least in my tests on macOS and Fedora + VS Code everything is working fine
- I will start implementing everything mentioned and your proposals :)

pypsa/plot/maps/interactive.py
```diff
@@ -315,279 +327,969 @@ def iplot(
...
+        deck = pdk.Deck(
+            layers=layers,
+            map_style=self._map_style,
+            tooltip=tooltip_content,
+            initial_view_state=self.view_state,
+            # set 3d view
         )
+        return deck
+
+    def build_layers(
```
lkstrp: Just move this into `def explore`. Then you have nice, logical separation for the `PydeckPlotter`, and no need to wrap things

examples/notebooks/explore.ipynb
```diff
@@ -0,0 +1,298 @@
+{
```
lkstrp: Remove and move to `v1-docs`

pypsa/plot/maps/common.py
```diff
@@ -111,3 +113,452 @@ def apply_layouter(
...
+    p0_m : tuple of float
+        (lon0, lat0) reference point in degrees.
+
+    Returns
+    -------
+    np.ndarray
+        Nx2 array of polygon vertices in (lon, lat) degrees.
+
+    """
+    R = 6378137.0  # equitorial radius in meters
```
lkstrp: Move to `constants.py`

bobbyxng:
New features added:
- [x] `geomap`
- [x] `jitter`
- [x] `_cmap`, `_cmap_norm`
- [x] default `map_style="road"`
- [x] Errors are thrown now

image

@fneum

bobbyxng:
- [X] Added geomap support for all resolutions `{"10m","50m", "110m"}` for ocean and land feature.
- [X] Added support for passing a geomap_colors dict.

Note that pydeck cannot handle shapes stretching beyond or to the poles. This is why I a supporting clipping function to remove marginal areas at lat and lon 90/-90 minus EPS (1e-6).

pypsa/plot/maps/interactive.py
bobbyxng: As discussed, for now remain with separated architecture to support future extendability

examples/notebooks/explore.ipynb
bobbyxng: done!

bobbyxng:
- [X] Added `auto_scale` based on `bus_size_max` and `branch_width_max`
- [X] Added support for `geometry` column in branch components

bobbyxng:
- [X] Finalised unit tests
- [X] Added release nots.
- Demo: https://bxio.ng/files/pypsa-explore-demo

From my side it's also ready to merge :)
Thank you both @lkstrp and @fneum !

PR #1322, opened 2025-08-12
Contingency finder helper function

Cellophil:
Hi,

I updated two instances where the branch index name was not consistent. In network it's "component" and "name" (for "Line", "bla"); in sub_network it's sometimes nothing and sometimes "type" and "name". It might break things somewhere possibly.

The graph object has a very convenient fast way of finding system splitting contingencies (finding "bridges"). The BODF object will always contain columns even if the contingency is system splitting, which is then ill defined. The columns can have entries from nan to inf and possibly just numbers > 1 or even innocent looking numbers [0,1]. It might make sense to

- shout a warning in calculate BODF when there are system splitting contingencies
- ignore those columns
- overwrite all with nan

Maybe rather only the last one; many people wont need this and system splitting contingencies are world-wide not uncommon.


lkstrp:
Hey @Cellophil,
This may already have been fixed (for sure changed) with #1153. We will take a look once that PR is in. You can check out the behavior in the `new-opt` branch already

Cellophil:
Hi @lkstrp ,

I checked the branch, at least the function calculate BODF is unchanged and should have the same problem as before (accepting system splitting branches as contingencies). Functions like lpf_contingency rely on nothing being weird about contingencies, so I think it should be checked or excluded. The problem is that system splitting contingencies in the BODF have ill defined behavior, you can get nan or inf (good, at least it's clear) or any float [0, inf] for weird numerical reasons. The BODF can not itself filter these columns for that reason.
There's also still "component" and "type" being used for branch indices, but I don't know the intended nomenclature.

For the BODF it's maybe sufficent to see that the graph object has this convenient and very fast nx.bridges(sn.graph()) that points out the system splitting branches. So no need to use my code, maybe there's a more elegant solution.

Cellophil:
Minimal example:

```
import pypsa
import pandas as pd

n = pypsa.Network()

for i in range(10):
    n.add('Bus', name=f'Bus_left_{i}')
    for j in range(i-1):
        n.add('Line', name=f'Line_left_{i}{j}', bus0=f'Bus_left_{i}', bus1=f'Bus_left_{j}', s_nom=100, x=1)

for i in range(10):
    n.add('Bus', name=f'Bus_right_{i}')
    for j in range(i-1):
        n.add('Line', name=f'Line_right_{i}{j}', bus0=f'Bus_right_{i}', bus1=f'Bus_right_{j}', s_nom=100, x=1)

# connecting line, aka bridge, should never be contingency
n.add('Line', name='Bridge', bus0='Bus_left_0', bus1='Bus_right_0', s_nom=100, x=10)

n.determine_network_topology()
sn = n.sub_networks.loc["0", "obj"]
sn.calculate_BODF()
bodf = pd.DataFrame(data=sn.BODF, index=sn.branches_i(), columns=sn.branches_i())

# this should all be nan:
bodf.loc[:, ('Line', 'Bridge')].sort_values()
```

fneum:
The function is quite helpful! Thanks for contributing it!

Now, the remaining question is how to use it. Perhaps it could be called at the beginning of `calculate_BODF()` and even throw an error (rather than just a warning) if system splitting contingencies are present, a la:

"System splitting contingency detected in lines {lines}. These are currently not supported."

That would require an adjustment to `calculate_BODF()` such that one dimension, only contains a list of `branch_outages`, e.g.

```py
branch_outages = n.lines[:10]
n.determine_network_topology()
sn = n.sub_networks.loc["0", "obj"]
sn.calculate_BODF(branch_outages=branch_outages)
```

which could be passed from `n.optimize.optimize_security_constrained(branch_outages=branch_outages)`.

I would not just silently replace with NaN (as people would expect the system splitting contingency would be considered). At least warning plus stating it's ignored.

Cellophil:
Ok great, sounds good - I would also attach the branch MultiIndex (both index / columns) and covert the numpy matrix to a pandas Dataframe? I can try to set that up and update the pull request.

fneum:
Yes, let's do that. Thanks!

Cellophil:
Hi,

This code works on my end but some comments...
The PTDF index hangs in the air a bit. The code is
```
        z = np.concatenate(
            [
                (c.static.loc[c.static.query("active").index, attribute]).values
                for c in self.iterate_components(n.passive_branch_components)
            ]
        )
```
and is identical to branches_i code as far as I've checked. The index is therefore branches_i, buses_o. It would presumably be cleaner to derive this directly from branches_i.

I created BODF_df and PTDF_df but maybe it's time to switch completely over to dataframes and deprecate numpy arrays without indices...

I haven't updated the security constrained optimization; it'll fail if there's a bridge element currently (actually, as it should), but the user will not be informed. It would be an option to add a quick warning in the BODF calculation; I didn't because I wasn't sure it's the right spot to do it. Possibly BODF is recalculated in some places and then it would throw the warning every time. Also, I might be aware of it and I can choose to run the security constrained optimization with a subset of branches.

Also, the security constrained optimization multiplies out the outage with every branch in the system and adds the respective constraint. That is quite a number of constraints... It is the only way to get a full n-1 optimization without iteration though.
With iteration, subselecting violated branch x outages (CNEC) works quite well. It's however a manual process I think.

PR #1329, opened 2025-08-14
feat: full support for current and new components API

lkstrp:
Follow up on #1235

Let's hope this is one of the last PRs with quite a diff. But this time not very intrusive.

To support both the new and old components APIs, this PR introduces to not using either of them, but rather the never changing class reference in the ComponentsStore.

E.g. PyPSA and all tests work now with
`n.generators` -> Components class
as well as 
`n.generators` -> static df
by just using everywhere:
`n.c.generators` 

Also another Ubuntu CI job is added which runs all tests with the new API

### DIFF
--- a/pypsa/plot/maps/interactive.py
+++ b/pypsa/plot/maps/interactive.py
@@ -542,7 +542,7 @@ def explore(
 
     if not n.c.loads.static.empty and "Load" in components:
         loads = n.c.loads.static.copy()
-        loads["p_set_sum"] = n.c.c.loads.dynamic.p_set.sum(axis=0).round(1)
+        loads["p_set_sum"] = n.c.loads.dynamic.p_set.sum(axis=0).round(1)
         gdf_loads = gpd.GeoDataFrame(
             loads,
             geometry=gpd.points_from_xy(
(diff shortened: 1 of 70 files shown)


lkstrp:
> lovely, I just scrolled through, any particular places where I should watch closer?

No, it's more for you to be aware for what I am doing

FabianHofmann:
> No, it's more for you to be aware for what I am doing

you do magic, I already know ;)

pypsa/plot/maps/interactive.py
```diff
@@ -537,13 +540,13 @@ def explore(
         )
         components_present.append("Generator")
 
-    if not n.loads.empty and "Load" in components:
-        loads = n.loads.copy()
-        loads["p_set_sum"] = n.loads_t.p_set.sum(axis=0).round(1)
+    if not n.c.loads.static.empty and "Load" in components:
+        loads = n.c.loads.static.copy()
+        loads["p_set_sum"] = n.c.c.loads.dynamic.p_set.sum(axis=0).round(1)
```
fneum:
```suggestion
        loads["p_set_sum"] = n.c.loads.dynamic.p_set.sum(axis=0).round(1)
```

fneum (approved):
It's great that we can test both simultaneously!

I found one double `n.c.c...` :)

pypsa/plot/maps/interactive.py
lkstrp: Still the sharpest reviewer eye 🚀

PR #1333, opened 2025-08-15
Add processes components as multi-link alternative

### LINKED ISSUE #549: Adaption of multilink implementation
Q: **Should we refactor the multilink representation from wide table format to a long table format?**

Some discussion points:

Not or partially supported functionalities and problematic extensions with current multilinks implementation:

## IO

Atm we have to `use override_component_attrs` to load any pypsa-eur-sec network. There are also overrides not related to multilinks (<https://github.com/PyPSA/pypsa-eur-sec/tree/master/data/override_component_attrs>) but these are actually not necessarily needed as tracked by the IO of pypsa anyway (custom static columns are exported and imported). We discussed how to automatically support overrides when loading from disk <https://github.com/PyPSA/PyPSA/issues/321> but it is tedious and requires deep dive into the `io.py` module. For each data format we would need a `parse_override` function to be called in the `__init__` of the network before the override section and a `write_override` for each export.
This becomes problematic as soon as we want to publish sector networks on Zenodo as stand alone files.

## Clustering

Clustering does not support multilinks. This would need to be done when adding multilinks at an earlier stage in PyPSA-EUR.

## Plotting

It is not possible to plot secondary links in the plotting routine. One had to create a new network in order to show intersector links and the sub topology of a location. Hiding links in the plot on the other hand is easy.

## Statics

The statistics module is not aware of multilinks. Biggest hurdles are groupbys and negative efficiency's and emtpy rows. This would require quite some case distinctions and unnecessarily bloated code.

## Power flow

The PF calculation does not support p_set of multilinks. In our workflows, we do not have electricity as secondary outputs, but users might rely on it.

## Time-dependent efficiency's

This will be quiet tricky to generalize. Efficiency's would also be needed to go into the time series dict eventually with a subset of the link index. The `get_switchable_as_dense` function would need to be adapted quite clevery.

---

Possible advantages of the alternative data structure using a `coupling` column:

- Nearly all of the above issues are resolved. The only thing I see is the clustering which needs to ignore secondary links (`n.links[n.links.coupling.isnull()]`) and after the clustering a mapping of the `coupling` column to the linkmap.
- It is arbitrarily extendable.
- Carriers can be added to secondary links, e.g. it would be possible to actually use the carrier for a carrier specification and not a technology specification.
- Negative efficiencies could be avoided by switching the direction of the secondary link. Actually generalized couplings accross a network could be created.
- Marginal cost could be added to secondary links and would be automatically considered in the objective function: `n.links.groupby(n.links.coupling.fillna(n.links.index)).marginal_cost.sum()`

Possible disadvantages:

- `p0` has a subset of the link index in presence of secondary links.
- some columns are obselete for secondary links, e.g. `capital_cost` and need warning messages when used.
- code in prepare_sector_network.py needs to be adapted. Instead of

```python
n.madd("Link",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    bus2=nodes + " H2",
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    efficiency=1 / (cf_industry["MWh_elec_per_tNH3_electrolysis"] / cf_industry["MWh_NH3_per_tNH3"]), # output: MW_NH3 per MW_elec
    efficiency2=-cf_industry["MWh_H2_per_tNH3_electrolysis"] / cf_industry["MWh_elec_per_tNH3_electrolysis"], # input: MW_H2 per MW_elec
    capital_cost=costs.at["Haber-Bosch synthesis", "fixed"],
    lifetime=costs.at["Haber-Bosch synthesis", 'lifetime']
)
```

the new patters would be

```python
n.madd("Link",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    efficiency=1 / (cf_industry["MWh_elec_per_tNH3_electrolysis"] / cf_industry["MWh_NH3_per_tNH3"]), # output: MW_NH3 per MW_elec
    capital_cost=costs.at["Haber-Bosch synthesis", "fixed"],
    lifetime=costs.at["Haber-Bosch synthesis", 'lifetime']
)

n.madd("Link",
    nodes,
    suffix=" Haber-Bosch H2",
    bus0=nodes,
    bus1=nodes + " H2",
    coupling=nodes + " Haber-Bosch",
    carrier="Haber-Bosch",
    efficiency=-cf_industry["MWh_H2_per_tNH3_electrolysis"] / cf_industry["MWh_elec_per_tNH3_electrolysis"], # input: MW_H2 per MW_elec
)
```

@fneum @nworbmot @lisazeyen

coroa:
Closes #549 .

Adds a new `processes` component as a multi-link alternative.

The difference to links is that `efficiency`, `efficiency2`, ... are replaced by `rate0`, `rate1`, `rate2`, ... to determine the energy output (always positive sign) at `bus0`. As well as a consistent sign convention, in that the power output at busX of the process is always `rateX * p` where p is the internal optimization variable.

This allows implicitly to choose the reference bus, by setting any `rate` to 1 or -1.

Take the example of the Haber-Bosch process, which is written with a (Multi-)Link as:
```python
n.add(
    "Link",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    bus2=nodes + " H2",
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    efficiency=1 / costs.at["Haber-Bosch", "electricity-input"],
    efficiency2=-costs.at["Haber-Bosch", "hydrogen-input"]
    / costs.at["Haber-Bosch", "electricity-input"],
    capital_cost=costs.at["Haber-Bosch", "capital_cost"]
    / costs.at["Haber-Bosch", "electricity-input"],
    marginal_cost=costs.at["Haber-Bosch", "VOM"]
    / costs.at["Haber-Bosch", "electricity-input"],
    lifetime=costs.at["Haber-Bosch", "lifetime"],
)
```

with the `Process` component this becomes:
```python
n.add(
    "Process",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    bus2=nodes + " H2",
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    rate0=-costs.at["Haber-Bosch", "electricity-input"],  # negative rate consumes
    rate1=1,   # positive rate produces; 1 sets the reference bus
    rate2=-costs.at["Haber-Bosch", "hydrogen-input"],
    capital_cost=costs.at["Haber-Bosch", "capital_cost"],
    marginal_cost=costs.at["Haber-Bosch", "VOM"],
    lifetime=costs.at["Haber-Bosch", "lifetime"],
)
```

See also discussion in discord thread: https://discord.com/channels/914472852571426846/1405839054305624156 .

The last commit 4210200d5477d5b0e397aab2b4b0a5b8d4670b80 adds a variant of the sector coupling notebook replacing all Links (incl. a CHP Multilink) with Processes. The results are identical.

## Todo

- [ ] Couple of TODOs in the code, @lkstrp Would you mind reviewing already, whether i made obvious mistakes (or fucked up any of your excellent component work)
- [ ] Documentation needs to be updated
- [ ] Plotting functions need to work with Process too

## Checklist

- [ ] Documentation updates
- [ ] 


coroa:
Rebased to the actual master now. Had an old state. The test failures are deprecations in the master branch.

euronion:
Instead of setting `rateX=1` to specify the reference bus, how about specifying the reference bus explicitly, e.g. with 

```python
reference_bus="bus1"
```

That way, the rates have no implied meaning and users do not need to calculate the rates relative to one particular rate (one less calculation step that can be erroneous).

Say in your example you have the inputs and outputs of the process in absolute values, then you could plug these values directly (dummy values, with approx 5 MWh/t ammonia)

```python
n.add(
    "Process",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    bus2=nodes + " H2",
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    rate0=-1.2,
    rate1=5,
    rate2=-5.5,
    reference_bus="bus1",
    capital_cost=costs.at["Haber-Bosch", "capital_cost"],
    marginal_cost=costs.at["Haber-Bosch", "VOM"],
    lifetime=costs.at["Haber-Bosch", "lifetime"],
)
```

coroa:
Hi, @euronion, the reference_bus would then be for fixing the units for capital and marginal costs?

Efficiencies are already fully determined by the ratios of the rates. Not sure what this would then do? I think i fail to understand what would be the input data that would lead you to set
```
   rate0=-1.2,
   rate1=5,
   rate2=-5.5
```
do factsheets of a process not always report efficiencies as ratios between ports?

In your example, the process is producing 5 t of ammonia, by consuming 1.2 MWh of electricity and 5.5t H2. Did you mean that? And marginal costs are expected as Eur/5t ammonia or Eur/1.2MWhel.

Setting the reference bus to bus1 would then allow you to specify it as Eur/t ammonia, i guess.

euronion:
> Hi, @euronion, the reference_bus would then be for fixing the units for capital and marginal costs?
> 
> Efficiencies are already fully determined by the ratios of the rates. Not sure what this would then do? I think i fail to understand what would be the input data that would lead you to set
> 
> ```
>    rate0=-1.2,
>    rate1=5,
>    rate2=-5.5
> ```
> 
> do factsheets of a process not always report efficiencies as ratios between ports?
> 
> In your example, the process is producing 5 t of ammonia, by consuming 1.2 MWh of electricity and 5.5t H2. Did you mean that? And marginal costs are expected as Eur/5t ammonia or Eur/1.2MWhel.
> 
> Setting the reference bus to bus1 would then allow you to specify it as Eur/t ammonia, i guess.

Yes, you understood that correctly.
While there are some cases where technology data sheets are not normed, my suggestion primarily stems from the urge to not hide and imply the "reference_bus" information based on the `rate_X=1` setting, but to make it explicit.

coroa:
I'll wait for #1349 to land and then rebase on that.

pypsa/descriptors.py
```diff
@@ -256,32 +262,50 @@ def _update_linkports_component_attrs(
...
+
+    attrs_index = n.components[c]["attrs"].index
+    for i, attr in product(ports, ["bus", *varying_attrs]):
         target = f"{attr}{i}"
         if target in n.components[c]["attrs"].index:
             continue
         j = "1" if attr != "efficiency" else ""
         base_attr = attr + j
-        base_attr_index = n.components[c]["attrs"].index.get_loc(base_attr)
-        n.components[c]["attrs"].index.insert(base_attr_index + 1, target)
```
coroa: I already want to raise this to you, @lkstrp .

The logic here tries to insert efficiency2+ and bus2+ after their respective `bus1`, `efficiency` attributes, but that does not work since `pd.Index.insert` is not in-place, on the contrary, `pd.Index` is immutable.

lkstrp: Which means this is an existing bug, and already not updated? `attrs` here is based on component types which need an overhaul anyway. The current implementation mainly came from custom components, which we fully deprecated

pypsa/descriptors.py
```diff
@@ -256,32 +262,50 @@ def _update_linkports_component_attrs(
...
             )
             n.dynamic(c)[target] = df
         elif attr == "bus" and target not in n.static(c).columns:
             n.static(c)[target] = n.components[c]["attrs"].loc[target, "default"]
+
+    # re-order correctly
+    # TODO: there is no in-place reordering or in-place inserting at the right row (as far as i am aware),
+    # so we would have to update the attrs for that to work, but there i am hitting some special components
+    # setters, @lkstrp
+    # n.components[c]["attrs"] = n.components[c]["attrs"].loc[attrs_index]
```
lkstrp: Yes this is the component type. Needs to be handled there. So put a TODO here for now. If it's already broken, it's also not much of an issue

pypsa/descriptors.py
coroa: Yes that is exactly what it means. bus2, efficiency2, ... are added at the end of all columns.

lkstrp:
>  Couple of TODOs in the code, @lkstrp Would you mind reviewing already, whether i made obvious mistakes (or fucked up any of your excellent component work)

I only skimmed this, not questioning the processes component, but you are adhering to the component architecture wonderfully 👍

coroa:
Instead of `reference_bus="bus0"`, we introduce `relative_to="bus1"` to make it explicit, to be provided as required argument.

FabianHofmann:
> Instead of `reference_bus="bus0"`, we introduce `relative_to="bus1"` to make it explicit, to be provided as required argument.

@coroa I thought about it again: I imagine use cases where you have a process component with a reference unit, however the sector of that reference unit is not in the system (or not any more). For example you want to model haber-bosch without the explicit ammonia sector, just model the hydrogen and electricity consumption, potentially keeping the same dispatch of the HB asset from a previous run. In that case you would loose the flexibility to use the original rate data and have to convert all units to either hydrogen or electricity. 

Therefore, the way I see it, most flexibility you gain with discarding the reference bus and allow for arbitrary rate values. The output of the process would contain a `p` (along with `p0`, `p1`, ...) which itself indicates the internal, reference operation of the process. If `rate<n>=1` then  `p==p<n>` . however  the  latter is not a requirement. Like this you allow for arbitrary selections of in- and outputs of your process while keeping the same input data. Does that make sense?

coroa:
> > Instead of `reference_bus="bus0"`, we introduce `relative_to="bus1"` to make it explicit, to be provided as required argument.
> 
> @coroa I thought about it again: I imagine use cases where you have a process component with a reference unit, however the sector of that reference unit is not in the system (or not any more). For example you want to model haber-bosch without the explicit ammonia sector, just model the hydrogen and electricity consumption, potentially keeping the same dispatch of the HB asset from a previous run. In that case you would loose the flexibility to use the original rate data and have to convert all units to either hydrogen or electricity.
> 
> Therefore, the way I see it, most flexibility you gain with discarding the reference bus and allow for arbitrary rate values. The output of the process would contain a `p` (along with `p0`, `p1`, ...) which itself indicates the internal, reference operation of the process. If `rate<n>=1` then `p==p<n>` . however the latter is not a requirement. Like this you allow for arbitrary selections of in- and outputs of your process while keeping the same input data. Does that make sense?

well I agree with you, and this internal dispatch is also already in the code, since it made sense anyway, but that is not what was decided.

FabianHofmann:
> well I agree with you, and this internal dispatch is also already in the code, since it made sense anyway, but that is not what was decided.

I think this was not fully clear to the other (at least it wasn't to me). The flexibility would be very helpful and we should go for it. @lkstrp @brynpickering @euronion let's reconsider? perhaps @fneum also has a view on this

euronion:
:no_entry_sign:  I'm still strongly in favour of making this information explicit, not implicit.

:heavy_check_mark: I do see the point that you are making for cases where the reference is not available or part of the system/process.

An option could be a compromise, where we allow overloading of `relative_to` to be either the reference bus (to avoid the user having to specify redundant values) or the reference rate explicitly:

```python
relative_to: str | float = "bus1"
# or
relative_to: str | float = <p>
```

coroa:
I took a stab at an implementation of `relative_to` today,

```python
n.add(
    "Process",
    nodes,
    suffix=" Haber-Bosch",
    bus0=nodes,
    bus1=spatial.ammonia.nodes,
    bus2=nodes + " H2",
    p_nom_extendable=True,
    carrier="Haber-Bosch",
    rate0=-costs.at["Haber-Bosch", "electricity-input"],  # negative rate consumes
    rate1=1,   # positive rate produces; 1 sets the reference bus
    rate2=-costs.at["Haber-Bosch", "hydrogen-input"],
    capital_cost=costs.at["Haber-Bosch", "capital_cost"],
    marginal_cost=costs.at["Haber-Bosch", "VOM"],
    lifetime=costs.at["Haber-Bosch", "lifetime"],

    relative_to="bus2",
)
```

Ie. the internal dispatch variable is synced with the ammonia output; the `relative_to` is choosing the hydrogen input as the reference.

I think this means, that the `marginal_cost` should be in `- x Eur/t hydrogen` (minus sign, because of an input).

Does this mean that p_nom_opt is then expected in `t hydrogen`?

if so, i think that means i need to actually rescale all the rate variables internally by that `rate2`:
`rateX_effective = rateX_nom / rate2`.

coroa:
> An option could be a compromise, where we allow overloading of `relative_to` to be either the reference bus (to avoid the user having to specify redundant values) or the reference rate explicitly:
> 
> ```python
> relative_to: str | float = "bus1"
> # or
> relative_to: str | float = <p>
> ```

That's a bit complicated by the fact that we are starting to allow also just the port numbers as integers as valid port signifiers, as introduced in #1386 .

If we tried to be consistent, then this would mean `relative_to=2` would mean relative to `rate2` and `relative_to=2.` would mean relative to a rate of `2.0`.

euronion:
I didn't know about the port and discussion in the other PR. Is it a good idea to allow just the port numbers as integers? I understand it is convenient, at the same time it reduces our degrees of freedom for future development (maybe?).

Having `relative_to=2` and `relative_to=2.0` behave differently seems like a bad idea to me.
If `port` is being used already anyways, what about `reference_port = str | int`?

Shall we revisit this discussion next Wednesday and collect the options and ideas again by then?

lkstrp:
> I think this was not fully clear to the other (at least it wasn't to me). The flexibility would be very helpful and we should go for it. @lkstrp @brynpickering @euronion let's reconsider? perhaps @fneum also has a view on this

I talked with both Fabians again, and @fneum is also very much in favor of keeping this implicit without the need for an extra attribute. Just bring one of your rates to 1 and all good. I think this is also more intuitive and simpler. I am also against any of those mixed solutions where reference_to could be optional. Since @coroa and @FabianHofmann see it the same way I guess the vote would have changed already anyway. Could still discuss that next week (again) if you think it's needed

euronion:
After the arguments that @brynpickering shared about making it explict I was under the impression last time that explicit should be preferred.

But feel free to go ahead if that is the current majority vote.

FabianHofmann:
thanks for being flexible @euronion on this (it's always a back and forth, isn't it). if we see there the interpretation is too complicated we still can add explicit things later

FabianHofmann:
### Code review

Found 1 issue:

1. Sign convention for Process port powers is inconsistent between power flow and optimization. In power flow, `p{i} = p_set * rate{i}`, but in optimization `assign_solution`, `p{i} = -p * rate{i}`. With the default `rate0 = -1`, power flow gives `p0 = -p_set` (negative when consuming) while optimization gives `p0 = +p` (positive when consuming). This means a round-trip (optimize then run PF) will produce opposite signs for `p0`. The TODO comment at line 879 also flags uncertainty about the negative sign.

Power flow:
https://github.com/PyPSA/PyPSA/blob/1e5926ef86fc2709425743e64e91cfed35e9f17b/pypsa/network/power_flow.py#L184-L187

Optimization:
https://github.com/PyPSA/PyPSA/blob/1e5926ef86fc2709425743e64e91cfed35e9f17b/pypsa/optimization/optimize.py#L877-L881

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

FabianHofmann:
@coroa sorry, did not intend to push here directly, but now it's there. added a test and a fix for the issue reaised above. power flow calculation is now consistent with the process component

FabianHofmann:
@lkstrp could you have a look at the `multiports.py`. I had to pull in some types to collect the shared functions, don't know how clean that is. also I am importing the `Multiports` from `_types` in `descriptors.py` (better avoided?)

docs/api/components/types/processes.md
```diff
@@ -0,0 +1,7 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+::: pypsa.components.Processes
```
lkstrp, line 7: Multiport methods and properties are currently not shown in API

docs/examples/sector-coupling-single-node-process.ipynb
```diff
@@ -0,0 +1,1076 @@
+{
+ "cells": [
```
lkstrp: This is currently a nearly duplicate for sector-coupling-single-node.ipynb. Whats the idea here? Update to processes or show both?

docs/user-guide/components/processes.md
```diff
@@ -0,0 +1,103 @@
...
+
+!!! note "Comparison with [`Link`][pypsa.components.Links]"
+
+    The key difference to [`Link`][pypsa.components.Links] is that `efficiency`,
+    `efficiency2`, ... are replaced by `rate0`, `rate1`, `rate2`, ... with a
+    single sign convention: the power output at `busX` is always `rateX * p`.
+    This avoids the implicit sign convention of Links where `bus0` is always the
+    input and efficiencies for additional inputs must be negative.
+
+!!! example "[`Process`][pypsa.components.Processes] for an electrolyser"
```
lkstrp, line 44: Can we align here the examples with Links? Either choose the ones from Links or use the ones here also for Links (not sure what I like more)

docs/user-guide/components/processes.md
```diff
@@ -0,0 +1,103 @@
...
+        rate0=-costs.at["Haber-Bosch", "electricity-input"],
+        rate1=1,           # reference bus
+        rate2=-costs.at["Haber-Bosch", "hydrogen-input"], # in tH2/tNH3
+        p_nom_extendable=True,
+        capital_cost=costs.at["Haber-Bosch", "capital_cost"], # in Eur/(tNH3/h)
+        marginal_cost=costs.at["Haber-Bosch", "VOM"], # in Eur/tNH3
+    )
+    ```
+
+{{ read_csv('../../../pypsa/data/component_attrs/processes.csv', disable_numparse=True) }}
```
lkstrp, line 104: Links has a way to specific delay section (which is out of consistency, since all other Link features are not as detailed explained) and that section is fully ignored here but also fully supported which doesn't make sense. I would strip it, we have a dedicated delay section somewhere else anyway isn't it?

pypsa/components/_types/shared_layer/multiports.py
```diff
@@ -0,0 +1,272 @@
...
+    if TYPE_CHECKING:
+        ctype: ComponentType
+        static: pd.DataFrame
+        dynamic: Dict
+        da: _XarrayAccessor
+        ports: list[str]
+        empty: bool
+        active_assets: pd.Index
+        defaults: pd.DataFrame
+        _unsuffixed_attrs: set[str]
```
lkstrp: Looks like a Claude invention which shouldn't be patched like this. You wanna inherit from Components instead

pypsa/components/_types/mixin/multiports.py
```diff
@@ -0,0 +1,272 @@
...
+        ['2']
+
+        """
+        return [
+            match.group(1)
+            for col in self.static.columns
+            if (match := RE_PORTS_GE_2.search(col))
+        ]
+
+    @property
```
lkstrp, line 61: What's going on here? I guess this is for different behavior for `Links` and `Processes`? Could this be made symmetric when adding breaking change or not?

In any case this would then need to be an abstractmethod here and defined in `Links` and `Processes` on its own. Even better would be to make it symmetric

pypsa/components/_types/shared_layer/multiports.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def _port_suffix(self, port: str) -> str:
+        return port
+
+    @property
+    @abstractmethod
+    def _coefficient_attr(self) -> str: ...
+
+    @property
+    def _delay_config(self) -> dict[str, tuple[pd.Series | int, pd.Series | bool]]:
```
lkstrp: I would rather move this and `_iter_balance_args` to the optimization code. As already discussed in the Links PR. `_delay_positions` is only here because it's used in `get_delay_source_indexer` which is exposed

pypsa/components/_types/mixin/multiports.py
```diff
@@ -0,0 +1,272 @@
...
+            for period in snapshots.unique("period"):
+                w = weightings[snapshots.get_loc(period)].to_numpy().astype(float)
+                p_src, p_valid = Multiport._delay_positions(w, delay, is_cyclic)
+                src[offset : offset + len(w)] = p_src + offset
+                valid[offset : offset + len(w)] = p_valid
+                offset += len(w)
+            return src, valid
+
+        w = weightings.reindex(snapshots).astype(float).to_numpy()
+        return Multiport._delay_positions(w, delay, is_cyclic)
```
lkstrp: `__init__` is missing in `shared_layer` and I would rename it to `mixin`, we already use that concept for accessors

pypsa/components/_types/links.py
```diff
@@ -6,27 +6,22 @@
...
 
+    import pandas as pd
     import xarray as xr
 
-logger = logging.getLogger(__name__)
-
 
 @patch_add_docstring
-class Links(Components):
+class Links(Multiport, Components):
```
lkstrp: As discussed above, Multiport should already inherit from Components

pypsa/components/_types/links.py
lkstrp, line 32: Add a See Also Processes

pypsa/components/_types/processes.py
```diff
@@ -0,0 +1,89 @@
...
+class Processes(Multiport, Components):
+    """Processes components class.
+
+    This class is used for process components. All functionality specific to
+    processes is implemented here. Functionality for all components is implemented in
+    the abstract base class.
+
+    See Also
+    --------
+    [pypsa.Components][]
```
lkstrp, line 32: See also Links

pypsa/components/_types/processes.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    _operational_variables = ["p"]
+    _unsuffixed_attrs: set[str] = set()
+
+    @property
+    def _coefficient_attr(self) -> str:
+        return "rate"
+
+    def get_bounds_pu(
```
lkstrp, line 56: This is the same for `Links` isnt it? So can be moved to `Multiport`

pypsa/plot/maps/static.py
```diff
@@ -1791,6 +1791,8 @@ def add_legend_arrows(
 
 def round_to_significant_digits(x: float, n: int = 2) -> int | float:
     """Round a number to n significant figures."""
+    if np.isnan(x):
```
lkstrp, line 1794: Just a off topic bug fix?

lkstrp (changes requested):
A couple of comments. `Multiport` wants to inherit from Components and be a Mixin, that will also clean up all the patched type stuff

docs/examples/sector-coupling-single-node-process.ipynb
coroa: This has been updated to processes to demonstrate that equivalent results come out. Maybe should be merged into the original sector-coupling-single-node under a different top-level entry or only add the alternative as a comment?

pypsa/components/_types/shared_layer/multiports.py
FabianHofmann: yes, this was my question, thanks. I wasn't sure if a inheritance from Components in MultiPort would be an overkill and wanted to keep it light-weight. but thinking about it further, it makes sense to follow the inheritance chain Components > MultiPort > Links/Process (and the latter don't need the super class Components anymore)

docs/user-guide/components/processes.md
FabianHofmann: not sure what you mean. aligning with the example in links.md?

pypsa/plot/maps/static.py
FabianHofmann: yes

pypsa/components/_types/processes.py
FabianHofmann: no, it's different and needs to stay here

docs/examples/sector-coupling-single-node-process.ipynb
FabianHofmann: I merged them and added a note + alternative settings for processes

FabianHofmann:
@lkstrp @coroa I addressed all the open comments and closed them after addressing (otherwise I would have lost track). I will look out for the CI now, but otherwise this seems good to go

FabianHofmann:
I would add one more test and "bug fix" that columns from attributes from additional port are appearing after there predecessor

fneum:
The `Process` component looks very neat and elegant. Excellent!

I have one request for the finishing touches before merging:

The `Process` component is currently absent from the formulations in the user guide (i.e. subsequent sections of https://pypsa--1333.org.readthedocs.build/1333/user-guide/optimization/overview/). Could you complete this with the `Process` component everywhere? Some elements can probably be handled similarly jointly as `Line` and `Transformer` component.

docs/api/components/types/multiports.md
```diff
@@ -0,0 +1,7 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+::: pypsa.components.Multiport
```
lkstrp: No, we don't wanna expose this, methods here should be inherited to Links and Processes and shown there. I can just fix that

lkstrp (approved):

FabianHofmann:
> The `Process` component looks very neat and elegant. Excellent!
> 
> I have one request for the finishing touches before merging:
> 
> The `Process` component is currently absent from the formulations in the user guide (i.e. subsequent sections of https://pypsa--1333.org.readthedocs.build/1333/user-guide/optimization/overview/). Could you complete this with the `Process` component everywhere? Some elements can probably be handled similarly jointly as `Line` and `Transformer` component.

good point! done

PR #1338, opened 2025-08-19
Bugfix: Handle alphanumeric versions (like "1.0.0rc1")

bobbyxng:
- Making __version_semver__ compatible with alphanumeric versions like `1.0.0rc1` as introduced with latest master.
- Previously, some built-in URLs did not resolve correctly, e.g. `n.examples.scigrid_de()` would be pointing to a URL based on `1.0.0` (which does not exist - yet) instead of `1.0.0rc1` where the actual resources reside.
- Extended unit tests for `test_version.py`

## Checklist

### DIFF
--- a/pypsa/version.py
+++ b/pypsa/version.py
@@ -31,17 +31,58 @@ def check_pypsa_version(version_string: str) -> None:
         )
 
 
+def parse_version_tuple(version_str: str) -> tuple[int | str, ...]:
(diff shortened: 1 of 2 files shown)


bobbyxng:
@lkstrp Propose refactoring the code to use `packaging.version` everywhere instead of version tuples, as it is way more robust and has alphanumeric versioning including "rc1", "rc2", etc. built-in

PR #1345, opened 2025-09-01
Add risk-averse optimization

Irieo:
# Add CVaR-based risk-averse optimization

we build here on top of https://github.com/PyPSA/PyPSA/pull/1154

This PR integrates a linearized Conditional Value-at-Risk (CVaR) formulation as a new feature for PyPSA’s stochastic optimization. It augments the objective with a convex risk term and adds auxiliary variables and constraints to represent CVaR over scenario-dependent operating costs.

- New risk preference API on the Network: `set_risk_preference(alpha: float, omega: float)`.
- Works seamlessly for stochastic networks; falls back to the original objective for deterministic networks or when risk preference is not set.

[image](https://iriepin.com/publication/p_2023_enecon_riskav/featured_hu9270cf9246e3d35b8fe4176a21cd9617_71931_720x2500_fit_q75_h2_lanczos_3.webp)

## Mathematical formulation

This formulation was introduced by Rockafellar & Uryasev (2002), and later adopted by folks from the energy economics & modelling fields, eg Munoz et al. (2017) is a great example!

Let $s \in S$ index scenarios with probability $p_s$, and let $\mathrm{OPEX}(s)$ denote the operating cost in scenario $s$. We introduce auxiliary decision variables:
- `a_s ≥ 0` (scenario excess costs)
- `θ` (VaR level)
- `CVaR` (the CVaR value)

The risk-augmented objective is

$$
\min\ OBJ := \mathrm{CAPEX} + (1-\omega)\*\mathbb{E}[\mathrm{OPEX}] + \omega\*\mathrm{CVaR}_{\alpha},
$$

subject to the standard PyPSA constraints and the following auxiliary constraints for CVaR linearization:

$$
	\theta + \frac{1}{1-\alpha} \sum_{s\in S} p_s\* a_s \le \mathrm{CVaR}.
$$

$$
a_s \ge \mathrm{OPEX}(s) - \theta.
$$

Here,
- $\mathbb{E}[\mathrm{OPEX}] = \sum_{s\in S} p_s\ * \mathrm{OPEX}(s)$
- $\alpha$ is the CVaR confidence level, so $1-\alpha$ is the tail mass
- $\omega$ is the risk preference weight

## Variables and constraints added
- Variables
  - `CVaR-a[scenario]` (nonnegative): excess loss $a_s$ for each scenario $s$.
  - `CVaR-theta` (free): VaR level $\theta$.
  - `CVaR` (free): CVaR value.
- Constraints
  - `CVaR-excess-<scenario>`: $a_s - \mathrm{OPEX}(s) + \theta \ge 0$.
  - `CVaR-def`: $\theta + \tfrac{1}{1-\alpha} \sum_s p_s a_s \le \mathrm{CVaR}$.

Notes
- The auxiliary variables are not written back to component tables; they are internal to the optimization model.
- Scenario probabilities come from `Network.scenario_weightings` and are used both in `E[OPEX]` and the CVaR-def constraint.
- Alpha (CVaR confidence level) has bounds of $0 < \alpha < 1$. At $\alpha = 1$ the denominator is zero (undefined). At $\alpha = 0$, the confidence level covers the entire distribution and the formulation degenerates (equivalent to expectation).
- Omega (risk weight) has bounds of $0 \le \omega \le 1$. The objective uses a convex combination $(1-\omega)\*\mathbb{E}[\mathrm{OPEX}] + \omega\*\mathrm{CVaR}$ Allowing the closed interval includes meaningful endpoints: $\omega=0$ yields the risk‑neutral expected‑cost objective; $\omega=1$ yields pure CVaR minimization.
- Our CVaR constraint uses tail $1/(1 − alpha)$, meaning it averages losses over the worst (1 − alpha) mass.
- For a case with quadratic marginal costs, CVaR constraint becomes a convex quadratic inequality, which is not supported by linopy. So `marginal_cost_quadratic` is the only option we're not supporting with CVaR. I add a guard for that.

## API for users
- `Network.set_risk_preference(alpha: float, omega: float)`
  - Sets the CVaR level `α` and risk weight `ω` on the network. When present and the network has scenarios, the optimization model automatically includes CVaR variables and constraints and augments the objective.
  - If not set, or if the network has no scenarios, the model behaves as before (risk‑neutral expected cost for stochastic; plain deterministic for non-stochastic).

## Backward compatibility
- Deterministic networks: unchanged objective and constraints.
- Stochastic networks without a configured risk preference: unchanged (risk‑neutral expected cost).

## How to use
```
import pypsa
n = pypsa.examples.model_energy()
n.set_scenarios({"volcano": 0.1, "no_volcano": 0.9})
n.generators_t.p_max_pu.loc[:, ("volcano", "solar")] *= 0.3

# Risk-neutral baseline (expected cost)
n.optimize()

# CVaR with moderate risk aversion
n.set_risk_preference(alpha=0.9, omega=0.5)
n.optimize()

# Edge case: CVaR with omega=0 equals risk-neutral
n.set_risk_preference(alpha=0.9, omega=0.0)
n.optimize()

# Edge case: CVaR capturing the results of the worst-case scenario (omega=1, alpha so only worst scenario is in tail)
p_worst = float(n.scenario_weightings.loc["volcano", "weight"])  # 0.1
n.set_risk_preference(alpha=1 - p_worst, omega=1.0)
n.optimize()
```

Example of results to expect
```
Capacity diff (omega=0.5 - neutral) [GW]:
component    carrier         
Generator    load shedding          0.000000
             solar                 10.882582
             wind                   4.636405
Link         electrolysis           0.594093
             turbine                0.364935
StorageUnit  battery storage       -1.313774
Store        hydrogen storage    1579.986342
```

CVaR penalizes expensive tail operations (worst-scenario OPEX like load shedding and costly peakers). It trades more CAPEX now to avoid high OPEX in the tail. In this system, the cheapest hedge is to overbuild cheap VRE (solar) even if it is hit in “volcano” scenario. Here “more solar + long-duration storage” is the least-cost way to suppress worst-case costs; however, what hedge risk-averse optimization yields depends on cost assumptions and degrees of freedom available. In short: **risk aversion hedges against the risky scenario by shifting costs from tail OPEX to upfront CAPEX**. 

A dedicated PR with a notebooks and user guide will go v1-docs branch https://github.com/PyPSA/PyPSA/pull/1250

## References
- Rockafellar & Uryasev (2002). Conditional Value‑at‑Risk for General Loss Distributions. Journal of Banking & Finance, 26(7), 1443–1471. DOI: https://doi.org/10.1016/S0378-4266(02)00271-6
- Munoz et al. (2017). Does risk aversion affect transmission and generation planning? A Western North America case study. Energy Economics, Volume 64. DOI: https://doi.org/10.1016/j.eneco.2017.03.025

## TODOs
- [x] add Network methods for risk preference (setter, getter, bool)
- [x] prototype first working version
- [x] prototype yields right solution for edge cases
- [x] make it nice
- [x] existing tests passing
- [x] add tests for risk aversion: API (index.py)
- [x] add tests for risk aversion: optimization (optimize.py and constraints.py)
- [x] release notes
- [x] make mypy happy
- [x] make codecov happier
- [x] think through CVaR constraints form if marginal costs are quadratic (if needed add a guard)

Other
-> [into v1-docs] add user guide and example notebook to docs (and patch where docs saying risk aversion will be in future releases)
-> is there a better way for a workaround in constraints.py to avoid circular import with optimize.py?


fneum:
I copy your user guide suggestions @fneum, thanks. I hide it here to keep this PR discussion shorter, and open a new PR to v1-docs branch with docs & examples.

doc/references/release-notes.rst
```diff
@@ -15,6 +15,25 @@ Upcoming Release
 Features
 --------
 
+* New optimization mode: Stochastic optimization problems
```
fneum: Match style with previous release notes (no extra links, shorter).

pypsa/network/index.py
```diff
@@ -757,3 +759,84 @@ def scenario_weightings(self) -> pd.DataFrame:
...
+    # -----------
+
+    def set_risk_preference(self, alpha: float, omega: float) -> None:
+        """Set risk aversion preferences for stochastic optimization using CVaR formulation.
+
+        Parameters
+        ----------
+        alpha : float
+            Confidence level in (0, 1). CVaR averages losses over the worst
+            (1 - alpha) probability mass (the "tail"). For "worst 10%",
```
fneum: remove `"`?

pypsa/network/index.py
```diff
@@ -757,3 +759,84 @@ def scenario_weightings(self) -> pd.DataFrame:
...
+            Risk preference parameter (risk aversion weight). Must be between 0 and 1.
+            - omega = 0: Risk-neutral optimization
+            - omega > 0: Risk-averse optimization (more focus on the tail risk)
+            - omega = 1: Maximum risk aversion (optimize for the tail risk only)
+            Higher values indicate more risk aversion.
+
+        Raises
+        ------
+        ValueError
+            If alpha is not between 0 and 1, or if omega is not between 0 and 1.
```
fneum:
```suggestion
            If alpha is not in (0,1), or if omega is not in [0,1].
```
To emphasize closed versus open interval.

pypsa/network/index.py
```diff
@@ -757,3 +759,84 @@ def scenario_weightings(self) -> pd.DataFrame:
...
+
+        Examples
+        --------
+        >>> n = pypsa.Network()
+        >>> n.set_scenarios({"low": 0.3, "medium": 0.4, "high": 0.3})
+        >>> n.set_risk_preference(alpha=0.95, omega=0.1)  # 5% tail CVaR (1 - 0.05)
+
+        Notes
+        -----
+        This method must be called after `set_scenarios()` as CVaR formulation
```
fneum, line 795: Do you really require the order? When calling `set_scenarios` or `set_risk_preferences` you're not defining any variables just yet. Instead of failing when calling `set_risk_preferences` before defining a stochastic network, you could just throw a warning when calling `n.optimize()` and risk preferences are set but we don't have a stochastic network. This would give more flexibility when building the network.

pypsa/optimization/constraints.py
```diff
@@ -28,6 +29,13 @@
 
 logger = logging.getLogger(__name__)
 
+# Local copy of the optimization variable lookup to avoid importing optimize (creates a cycle)
+# TODO make it nicer
+lookup = pd.read_csv(
+    Path(__file__).parent / ".." / "data" / "variables.csv",
+    index_col=["component", "variable"],
+)
+
```
fneum: This is something for @lkstrp to look into.

pypsa/optimization/constraints.py
```diff
@@ -1659,3 +1667,169 @@ def _extract_names(index: pd.Index) -> pd.Index:
...
+    Requirements
+    ------------
+    - n.has_scenarios is True and n.scenarios defined
+    - n.has_risk_preference is True with keys 'alpha' and 'omega'
+
+    Notes
+    -----
+    Uses the same OPEX construction as in the objective (incl. marginal costs, spill costs,
+    storage marginal costs, quadratic marginal costs, stand-by costs, and unit-commitment
+    costs) all weighted by snapshot durations (and investment period weights if relevant).
```
fneum: Can remove. This is obvious.

pypsa/optimization/constraints.py
```diff
@@ -1659,3 +1667,169 @@ def _extract_names(index: pd.Index) -> pd.Index:
...
+
+    Adds the standard linear CVaR constraints following Rockafellar & Uryasev (2000):
+
+    - For each scenario s: a(s) >= OPEX(s) - theta
+    - Theta + 1/(1-alpha) * sum_s p_s * a(s) <= CVaR
+
+    Requirements
+    ------------
+    - n.has_scenarios is True and n.scenarios defined
+    - n.has_risk_preference is True with keys 'alpha' and 'omega'
```
fneum: But `n.has_risk_preference` is bool. You mean `n.risk_preference`?

pypsa/optimization/constraints.py
```diff
@@ -1659,3 +1667,169 @@ def _extract_names(index: pd.Index) -> pd.Index:
...
+        weighting = n.snapshot_weightings.objective.loc[sns]
+
+    # Convert to xarray for reliable alignment with DataArrays below
+    weight = DataArray(weighting.values, coords={"snapshot": sns}, dims=["snapshot"])
+
+    # Assemble OPEX terms as in objective
+    opex_terms = []
+    is_quadratic = False
+
+    # marginal costs, storage marginal cost, and spill cost
```
fneum: There's quite a bit of code duplication with OPEX objective definition. Perhaps we can modularise the OPEX elements as functions and call them here again?

```
def marginal_cost(n, cost_types=["marginal_cost", "marginal_cost_storage", "spill_cost"]):
   ...

def stand_by_cost(n):
   ...

def start_up_shut_down_cost(n):
   ...
```

pypsa/optimization/constraints.py
```diff
@@ -1659,3 +1667,169 @@ def _extract_names(index: pd.Index) -> pd.Index:
...
+
+    # Retrieve CVaR variables
+    a = m["CVaR-a"]
+    theta = m["CVaR-theta"]
+    cvar = m["CVaR"]
+
+    # Define excess costs per scenario: a(s) - OPEX(s) + theta >= 0
+    for s in n.scenarios:
+        lhs = a.sel(scenario=s) - scen_opex_exprs[s] + theta
+        m.add_constraints(lhs, ">=", 0, name=f"CVaR-excess-{s}")
```
fneum: It would be nicer if scenarios were a dimension of the CVaR constraint. Could be performance bottleneck for many scenarios.

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,52 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+            _msg_rp = "Risk preference must be set when has_risk_preference is True"
+            raise RuntimeError(_msg_rp)
+        omega = rp.get("omega")
+        if not (isinstance(omega, int | float) and 0 <= float(omega) <= 1):
+            _msg_omega = f"omega must be a number between 0 and 1, got {omega}"
+            raise ValueError(_msg_omega)
+        cvar = m["CVaR"]
+        # Final objective: CAPEX + (1-omega) * E[OPEX] + omega * CVaR
+        obj_expr = (
+            expected_capex + (1 - float(omega)) * expected_opex + float(omega) * cvar
```
fneum: Is the typecast `float(omega)` necessary? If yes, could it be done earlier instead of here twice?

fneum (changes requested):

fneum:
Functionality-wise all excellent! Played a bit with the following code to calculate insurance premium, tail risk reduction, risk-hedging cost ratio and risk-cost frontiers. These would make excellent functions like planned for expected value of information.

```py
import pandas as pd

import pypsa

n = pypsa.examples.model_energy()
n.add(
    "Generator",
    "gas",
    carrier="gas",
    bus="electricity",
    p_nom=5000,
    efficiency=0.5,
    marginal_cost=100,
)

n.set_scenarios({"volcano": 0.1, "no_volcano": 0.9})
n.generators_t.p_max_pu.loc[:, ("volcano", "solar")] *= 0.3

n.set_risk_preference(alpha=0.9, omega=0)
n.optimize(log_to_console=False, solver_name="gurobi")

cvar_risk_neutral = float(n.model["CVaR"].solution) / 1e9

etc_risk_neutral = (
    n.statistics.capex().div(1e9).sum()
    + n.statistics.opex().mul(n.scenario_weightings.squeeze()).div(1e9).sum()
)
etc_risk_neutral

n.set_risk_preference(alpha=0.9, omega=0.5)
n.optimize(log_to_console=False, solver_name="gurobi")

cvar_risk_averse = float(n.model["CVaR"].solution) / 1e9

etc_risk_averse = (
    n.statistics.capex().div(1e9).sum()
    + n.statistics.opex().mul(n.scenario_weightings.squeeze()).div(1e9).sum()
)
etc_risk_averse

insurance_premium = etc_risk_averse - etc_risk_neutral
insurance_premium

tail_risk_reduction = cvar_risk_neutral - cvar_risk_averse
tail_risk_reduction

risk_hedging_cost = insurance_premium / tail_risk_reduction
risk_hedging_cost

cvar = {}
etc = {}
for omega in [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]:
    n.set_risk_preference(0.9, omega)
    n.optimize(log_to_console=False, solver_name="gurobi")
    cvar[omega] = float(n.model["CVaR"].solution) / 1e9
    etc[omega] = (
        n.statistics.capex().div(1e9).sum()
        + n.statistics.opex().mul(n.scenario_weightings.squeeze()).div(1e9).sum()
    )

pd.Series(etc).plot(label="expected total cost")
pd.Series(cvar).plot(label="conditional value at risk", ylim=(0, 12))
plt.legend()
plt.grid()

plt.scatter(etc.values(), cvar.values(), c=cvar.keys())
plt.grid()
plt.ylabel("conditional value at risk")
plt.xlabel("expected total cost")
plt.ylim(0, 3)
plt.xlim(9, 11)
```

image

image

doc/references/release-notes.rst
Irieo: Done

pypsa/network/index.py
Irieo: ok

pypsa/network/index.py
Irieo: Alright, this makes sense! I relaxed it, ie one can set risk preference on any network. Also I added a warning that shows up if optimization is called with risk prefs AND w/o scenarios saying risk prefs are ignored.

pypsa/optimization/constraints.py
Irieo: patched

pypsa/optimization/constraints.py
Irieo: ok

pypsa/optimization/optimize.py
Irieo: Hmm, it seems that it is indeed redundant since above I have a check `isinstance(omega, int | float)`, but it was there for mypy. I reworked it.

pypsa/optimization/constraints.py
Irieo: Right, I thought about it and mentioned in commit string here https://github.com/PyPSA/PyPSA/pull/1345/commits/48a1895f3a1fb466b3e926b2946f68d43bbaf2d2

I see three ways of doing 
A) keeping aux constraints in objective (the way it was in https://github.com/PyPSA/PyPSA/commit/48a1895f3a1fb466b3e926b2946f68d43bbaf2d2)
B) keeping aux constraints in contraints and re-type OPEX terms there (the way it is now)
C)  keeping aux constraints in contraints and introduce a shared OPEX builder functions

A pros: less code | A cons: aux constraints in objective function
B pros: clear and simple | B cons: duplication of code for OPEX 
C pros: remove duplication
C cons:
-  we will need flags there (eg CVaR must reject quadratic OPEX)
- how to avoid circular import? we would need smth like `optimization/costs.py`
- Right now one can see exactly which terms go into objective/CVaR. A shared builder will hide that, one will need to go here-and-back. Not a problem per se but why going this route.

Put short I don't like C, will chat with @lkstrp

Irieo: Resolution: we go A.

pypsa/optimization/constraints.py
Irieo: We keep it here as is, and sub lookups with constants.py import in a separate PR

pypsa/optimization/constraints.py
Irieo: Right. I also like how short and easy to follow this version is. I'd keep this so unless there is a good argument to wrap the dim inside.

pypsa/network/index.py
```diff
@@ -757,3 +759,67 @@ def scenario_weightings(self) -> pd.DataFrame:
...
+            are alpha in {0.90, 0.95, 0.99}. Higher alpha focuses on
+            rarer, more extreme tails; lower alpha considers a broader tail.
+        omega : float
+            Risk preference parameter (risk aversion weight). Must be between 0 and 1.
+            - omega = 0: Risk-neutral optimization
+            - omega > 0: Risk-averse optimization (more focus on the tail risk)
+            - omega = 1: Maximum risk aversion (optimize for the tail risk only)
+            Higher values indicate more risk aversion.
+
+        Raises
```
lkstrp:
```suggestion
```
Redundant

pypsa/network/index.py
```diff
@@ -757,3 +759,67 @@ def scenario_weightings(self) -> pd.DataFrame:
...
+        ------
+        ValueError
+            If alpha is not in (0,1), or if omega is not in [0,1].
+
+        Examples
+        --------
+        >>> n = pypsa.Network()
+        >>> n.set_scenarios({"low": 0.3, "medium": 0.4, "high": 0.3})
+        >>> n.set_risk_preference(alpha=0.95, omega=0.1)  # 5% tail CVaR (1 - 0.05)
```
lkstrp, line 789:
```suggestion
        >>> n.set_risk_preference(alpha=0.95, omega=0.1)  # 5% tail CVaR (1 - 0.05)
        >>> n.risk_preference
        {"alpha": 0.95, "omega": 0.1}
```

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+                # If quadratic terms exist, avoid merge (which is linear-only) and sum instead
+                merged = sum(selected) if is_quadratic else merge(selected)
+                terms.append(merged * p)
+            return sum(terms) if is_quadratic else merge(terms)
+        return sum(exprs) if is_quadratic else merge(exprs)
+
+    expected_capex = _expected(capex_terms)
+    expected_opex = _expected(opex_terms)
+
+    """
```
lkstrp: Is this a comment for artifact from moving things around?

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+    CVaR augmentation if enabled
+    Define auxiliary CVaR constraints for stochastic risk-averse optimization
+       - For each scenario s: a(s) >= OPEX(s) - theta
+       - Theta + 1/(1-alpha) * sum_s p_s * a(s) <= CVaR
+    """
+    if getattr(n, "has_scenarios", False) and getattr(n, "has_risk_preference", False):
```
lkstrp:
```suggestion
    if n.has_scenarios and n.has_risk_preference:
```
or even
```
if if n.has_risk_preference:
```
see below

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+        rp = n.risk_preference
```
lkstrp, line 311: Personal tase, but I would not assign it to another variable and just use `n.risk_preference`.

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+        if rp is None:
+            _msg_rp = "Risk preference must be set when has_risk_preference is True"
+            raise RuntimeError(_msg_rp)
```
lkstrp:
```suggestion
            raise ValueError(_msg_rp)
```

Actually, this can be removed completely, since we already check this in the setter/getter.

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+                f"got alpha={rp.get('alpha')}, omega={rp.get('omega')}"
+            )
+            raise ValueError(_msg) from None
+
+        if not (0.0 < alpha < 1.0):
+            _msg = f"alpha must be a number between 0 and 1, got {alpha}"
+            raise ValueError(_msg)
+        if not (0.0 <= omega <= 1.0):
+            _msg = f"omega must be a number between 0 and 1, got {omega}"
+            raise ValueError(_msg)
```
lkstrp:
```suggestion
```

Same here, just use rp["alpha"] or n.risk_preference["alpha"] and no checks needed

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+
+        # Guard: quadratic OPEX would make CVaR constraints quadratic
+        if is_quadratic:
+            msg_q = (
+                "CVaR with quadratic operational costs yields quadratic constraints. "
+                "So a(s) >= OPEX(s) - theta becomes a quadratic inequality. "
+                "Remove/approximate quadratic costs (e.g. set 'marginal_cost_quadratic=0' "
+                "or use a piecewise-linear approximation)."
+            )
+            raise RuntimeError(msg_q)
```
lkstrp:
```suggestion
            raise ValueError(msg_q)
```

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+        # Retrieve CVaR auxiliary variables
+        a = m["CVaR-a"]
+        theta = m["CVaR-theta"]
+        cvar = m["CVaR"]
+
+        for s in n.scenarios:
+            lhs = a.sel(scenario=s) - scen_opex_exprs[s] + theta
+            m.add_constraints(lhs, ">=", 0, name=f"CVaR-excess-{s}")
+
+        inv_tail = 1.0 / (1.0 - alpha)
```
lkstrp, line 345: If I understand correctly, no risk preference is the same as having `alpha=0` and `omega=0`. I am not sure if I then prefer to just have those as defaults in the risk_preference dict. Then we don't need to mix types there.

pypsa/optimization/optimize.py
```diff
@@ -275,27 +280,112 @@ def define_objective(n: Network, sns: pd.Index) -> None:
...
+            term = a.sel(scenario=s) * float(p)
+            weighted_a = term if weighted_a is None else weighted_a + term
+        m.add_constraints(theta + inv_tail * weighted_a, "<=", cvar, name="CVaR-def")
+
+        # Final objective: CAPEX + (1-omega) * E[OPEX] + omega * CVaR
+        obj_expr = expected_capex + (1 - omega) * expected_opex + omega * cvar
     else:
-        terms = objective
+        # Warn if risk preference was set but there are no scenarios
+        if getattr(n, "has_risk_preference", False) and not getattr(
```
lkstrp: Should we just not allow to set risk_preference when no scenarios are set?

pypsa/optimization/variables.py
```diff
@@ -234,3 +234,33 @@ def define_loss_variables(n: Network, sns: Sequence, c_name: str) -> None:
...
+
+def define_cvar_variables(n: Network) -> None:
+    """Define auxiliary variables for CVaR risk formulation.
+
+    Adds the following variables when stochastic optimization with CVaR is enabled:
+    - CVaR-a(scenario): auxiliary variable per scenario
+    - CVaR-theta: VaR level (scalar)
+    - CVaR: Conditional Value at Risk (scalar)
+
+    Requirements
```
lkstrp:
```suggestion
    Requirements:
    - n.has_scenarios must be True
    - n.has_risk_preference must be True
```
This is not according to our docstrings style
https://numpydoc.readthedocs.io/en/latest/format.html#sections

lkstrp (changes requested):
A couple of minor stuff and:

> If I understand correctly, no risk preference is the same as having alpha=0 and omega=0. I am not sure if I then prefer to just have those as defaults in the risk_preference dict. Then we don't need to mix types there.

What do you think?

Also, if we will never have `risk_preference` without stochastic networks, just disallow setting it on non stochastic networks.

fneum:
> Also, if we will never have risk_preference without stochastic networks, just disallow setting it on non stochastic networks.

Yes, it's correct that `risk_preference` should only be set for stochastic networks.

> If I understand correctly, no risk preference is the same as having alpha=0 and omega=0. I am not sure if I then prefer to just have those as defaults in the risk_preference dict. Then we don't need to mix types there.

I think that works. Actually omega=0 would be enough regardless of alpha (not that it matters for the default).

pypsa/optimization/optimize.py
Irieo: it is a comment (not an artifact); I move it up and merge into docstrings.

pypsa/optimization/optimize.py
Irieo: alright!

pypsa/optimization/optimize.py
Irieo: I am not sure if you imply here an invariant (has_risk_preference ⇔ risk_preference is not None) or checks for alpha and omega values. Both can be seen as redundant if everyone always uses API as intended, but these work as second guard if user somehow rewrites the values between setter and optimization. These checks are cheap also. I leave it for your judgement.

pypsa/optimization/optimize.py
Irieo: see above

pypsa/optimization/optimize.py
Irieo: `omega=0 ` means no weight is put to CVaR tail, which means no risk preference (we show it in example), it is fine.

for alpha it is more complicated: standard definition of CVaR (aka expected shortfall) is defined for alpha ∈ (0,1), exactly how I constrain the setter.  `alpha=1` is invalid (division by 1−alpha blows up and the tail has zero mass). `alpha=0` is definable but degenerate (ie no risk preference). However, keeping `alpha=0` as default is bad, because we add aux variables that don't have an effect on objective. 

So I'd not set these defaults; but happy for thoughts on this @lkstrp @fneum

pypsa/optimization/optimize.py
Irieo: I had this setup like you suggest initially, then I relaxed it in line with @fneum comment https://github.com/PyPSA/PyPSA/pull/1345#discussion_r2319529692

Irieo:
> > Also, if we will never have risk_preference without stochastic networks, just disallow setting it on non stochastic networks.
> 
> Yes, it's correct that `risk_preference` should only be set for stochastic networks.
> 
> > If I understand correctly, no risk preference is the same as having alpha=0 and omega=0. I am not sure if I then prefer to just have those as defaults in the risk_preference dict. Then we don't need to mix types there.
> 
> I think that works. Actually omega=0 would be enough regardless of alpha (not that it matters for the default).

regarding the point whether risk_preference should only be set for stochastic networks, or not (ie for any but ignore if not stochastic), it might be this point contradicts this https://github.com/PyPSA/PyPSA/pull/1345#discussion_r2319529692

@lkstrp what do you think?

upd: we decided for risk_preference only for stoch network way.

pypsa/optimization/variables.py
Irieo: fixed

pypsa/network/index.py
Irieo: ok!

pypsa/optimization/optimize.py
fneum: Maybe no defaults is better then.

If defaults, then for `alpha` something common like 0.9, 0.95, 0.99 perhaps.

pypsa/optimization/optimize.py
fneum: Either way is fine eventually. I generally think it's weird to require that things be done in a particular order -- or at least that it should be avoided as much as possible. We already have one "layer" of this to create stochastic networks.

pypsa/optimization/optimize.py
Irieo: Upd: decided for no defaults

pypsa/network/index.py
Irieo: fixed

Irieo:
@lkstrp all your points are addressed now, I think.

PR #1335, opened 2025-08-16
fix: address deprecations in network mixins

coroa:
As a proper fix instead of #1334 .

## Changes proposed in this Pull Request

Tries to address the deprecations raised with the new component system.

Not done.

## Checklist

- [ ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ ] Unit tests for new features were added (if applicable).
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [ ] I consent to the release of this PR's code under the MIT license.

PR #1349, opened 2025-09-05
refactor: resolve deprecations and make components iter compatible

lkstrp:
Supersedes #1335
Reverts #1334

In addition to the deprecated API changes, this PR fixes a bug whereby iterating over n.components also yielded empty components. This was incompatible with n.iterate_components(). This is a step towards having optional components, which are ignored when empty. Selecting specific component types from the component store comes in another PR.

@coroa Changes indeed blew up, so it was easier to start from scratch. I haven't checked if you added any other changes besides resolving the deprecations in #1335. Could you double check?

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,8 +98,7 @@ include = ["pypsa"]
 
 [tool.pytest.ini_options]
 filterwarnings = [
-    # Activate again with final 1.0 release
-    # "error::DeprecationWarning", # Raise all DeprecationWarnings as errors
+    "error::DeprecationWarning", # Raise all DeprecationWarnings as errors
     # "error::FutureWarning",      # Raise all FutureWarnings as errors # TODO Add again
     # "error:RuntimeWarning",
     # "error:UserWarning",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,12 +98,16 @@ include = ["pypsa"]
 
 [tool.pytest.ini_options]
 filterwarnings = [
-    "error::DeprecationWarning", # Raise all DeprecationWarnings as errors
-    # "error::FutureWarning",      # Raise all FutureWarnings as errors # TODO Add again
+    # Raise warnings as errors
+    "error::DeprecationWarning", 
+    # "error::FutureWarning", 
     # "error:RuntimeWarning",
     # "error:UserWarning",
+    
     "ignore::matplotlib._api.deprecation.MatplotlibDeprecationWarning:pydev",
     # See https://github.com/microsoft/debugpy/issues/1623
+    # "ignore:Using merge without specifying the class is deprecated:DeprecationWarning:linopy.*",
+    "ignore::DeprecationWarning:linopy.*",
 ]
 markers = [
     "mpl_image_compare",
--- a/pypsa/clustering/spatial.py
+++ b/pypsa/clustering/spatial.py
@@ -237,7 +237,7 @@ def aggregateoneport(
     if custom_strategies is None:
         custom_strategies = {}
     c = component
-    static = n.static(c)
+    static = n.c[c].static
     attrs = n.components[c]["attrs"]
     if "carrier" in static.columns:
         if carriers is None:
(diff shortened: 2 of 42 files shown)


fneum (approved):

coroa:
Well, i added quite a few also slightly unrelated changes and generalizations, since i was looking basically at each code site separately. So, i do not think starting from scratch was smart, but i will not go through all the code here now and do that again, either, so feel free to close the other one. Let's please discuss next time first.

Not that self.iterate_components(...) is equivalent to:
```
for c in self.components[...]:
    if c.empty:
        continue
```
due to `skip_empty = True` by default. Scrolling through the changes here, i did not see that.

lkstrp:
Okay, I just merged it back in. A couple of good changes, I agree (see https://github.com/PyPSA/PyPSA/pull/1349/commits/482f605c3a40f7c4379375c669a1d35038a4bb88).

> Not that self.iterate_components(...) is equivalent to:
> 
> ```
> for c in self.components[...]:
>     if c.empty:
>         continue
> ```

I made iterating over `self.components` the same as using `iterate_components` (e.g. by dropping empty components). I am also considering adding the same functionality to a custom `__getitem__`, with some additional selections. This could allow you to use something like `self.components['branch_components']`.

The downside of this custom behavior is that it might be a bit confusing. But I don't think it's anything that the documentation can't help with, and adding this functionality now, is a first step in allowing fully optional components in the future. Let me know what you think, @coroa.

coroa:
Great thank you, i spent quite some time on it. So happy it was not all wasted.

coroa:
> I made iterating over self.components the same as using iterate_components (e.g. by dropping empty components). I am also considering adding the same functionality to a custom __getitem__, with some additional selections. This could allow you to use something like self.components['branch_components'].

Hmm ... i thought i tested that to make sure. But i don't fully remember.

Good to know it is not needed.

lkstrp:
> Hmm ... i thought i tested that to make sure. But i don't fully remember.
> 
> Good to know it is not needed.

No, I am not clear.

This is correct `__getitem__`:
> Note that self.iterate_components(...) is equivalent to:
> ```
> for c in self.components[...]:
>     if c.empty:
>         continue
> ```

Same goes for `__iter__`:
`self.iterate_components()` is equivalent to:
```
for c in self.components:
    if c.empty:
        continue
```

The second one is changed for in this PR, so iter does not yield empty components anymore:

`self.iterate_components(...)` is equivalent to:
```
for c in self.components:
  ...
```

I plan to do the same for `__getitem__`, which will mean
`self.iterate_components(...)` is equivalent to:
```
for c in self.components[..]:
  ...
```

As described above. The issue of this approach is that it is custom behavior. I think this is fine, but iterating over self.components will not yield the same components that are displayed when `self.components` is simply printed. This needs to be explained in the documentation.

The advantage of adding this now is that it brings us closer to enabling fully optional components. This means that we could then drop components entirely without changing the behavior.

Maybe this needs also more opinions @fneum @FabianHofmann

coroa:
Thanks for the explanation. Ok, great. I like that behaviour, and i think it is fine if it is properly documented. It's a bit sad that we cannot add a kwarg to getitem like we had for iterate_components, but it is easy to work around that, so no big loss.

I am not enthusiastic about strings like `branch_components`  to identify what you want to iterate over, since strings already are expected to give you a concrete component. i don't like mixing based on concrete string names.

If you want something like this, i'd rather have that as part of an enum or maybe a type hierarchy or something like that.

To flesh out the type hierarchy idea, since you are now designing components as classes they could derive from BaseTypes like a `BranchComponent`. Then `__getitem__` could have something like:

```
def __getitem__(self, key):
    if issubclass(key, Component):
       key = {c for c in all_components if isinstance(c, key)}
    ...
```

And you then say
```
from pypsa.components import BranchComponent
n.components[BranchComponent]
```

Not sure, if we could then also get
```
n.components[BranchComponent | OnePortComponent]
```
to work.

lkstrp:
I very much agree with enums/ type hierarchy as mentioned above!
If I recall correctly, our types are already mutually exclusive. I'm not sure if there would ever be a need for other non-exclusive types, as this could make a hierarchy very chaotic.

But I'll finally merge this now. Let's move the discussion to another PR.

PR #1358, opened 2025-09-16
refactor: refine statistic arguments

### LINKED ISSUE #1253: `n.statistics` parameter redefinment
When I look at n.statistics in more detail, I think we should clean up and redefine the parameters, which have grown a bit over time. It's not just the documentation that could be improved, the signature is also not very consistent and intuitive.

If we have a look on all "Base" Parameters, which up until #1034, all statistic methods support:

<details>
<summary>Base Parameters</summary>

```
Parameters
----------
comps : str | Sequence[str] | None, default=None
    Components to include in the calculation. If None, includes all one-port
    and branch components. Available components are 'Generator', 'StorageUnit',
    'Store', 'Load', 'Line', 'Transformer' and'Link'.
aggregate_groups : Callable | str, default="sum"
    Function to aggregate groups when using the groupby parameter.
    Any pandas aggregation function can be used.
aggregate_across_components : bool, default=False
    Whether to aggregate across components. If there are different components
    which would be grouped together due to the same index, this is avoided.
groupby : str | Sequence[str] | Callable, default=["carrier", "bus_carrier"]
    How to group components:
    - str or list of str: Column names from component static DataFrames
    - callable: Function that takes network and component name as arguments
at_port : bool | str | Sequence[str], default=True
    Which ports to consider:
    - True: All ports of components
    - False: Exclude first port ("bus"/"bus0")
    - str or list of str: Specific ports to include
carrier : str | Sequence[str] | None, default=None
    Filter by carrier. If specified, only considers assets with given
    carrier(s).
bus_carrier : str | Sequence[str] | None, default=None
    Filter by carrier of connected buses. If specified, only considers assets
    connected to buses with the given carrier(s).
nice_names : bool | None, default=None
    Whether to use carrier nice names defined in n.carriers.nice_name. Defaults
    to module wide option (default: True).
    See `pypsa.options.params.statistics.describe()` for more information.
drop_zero : bool | None, default=None
    Whether to drop zero values from the result. Defaults to module wide option
    (default: True). See `pypsa.options.params.statistics.describe()` for more
    information.
round : int | None, default=None
    Number of decimal places to round the result to. Defaults to module wide
    option (default: 2). See `pypsa.options.params.statistics.describe()` for
    more information.
```

</details>

Then we can group them into filtering, aggregation, and general processing.

**Aggregation**
With `groupby`, `aggregate_groups`, `aggregate_across_components` and if we optionally add `aggregate_time`, there are 3 aggregation methods (`aggregate_groups` is just a addition to `groupby`), which are named and implemented very differently.

Unaggregated results would have data (optionally) `snapshots` as columns and a multi index out of `component` (component type) and `name`. Right now you would need to pass `..(groupby=False, aggregate_time=False)` which is not very intuitive. `groupby` would aggregate the index, `aggregate_time` would aggregate the columns.

For `aggregate_across_components`, which aggregates across index dimension, I don't even see a reason why it should be there. Why would you want to name different components with the same name? Specially since you can just pass `.groupby("name").sum()` on the output and achieve the results? Can we kick it?

So a couple of adjustments I would like to make/ discuss:
1. Kick `aggregate_across_components`
2. Rename `aggregate_groups` to `groupby_method`, to make clear that this is only relevant with `groupby`
3. Use similar, concise names for `groupby` and `aggregate_time`. It should be clear that they are both doing the same thing, just on different dimensions. Something like `groupby` and `groupby_time`, `aggregate_components` and `aggregate_time`, or `groupby_components` and `groupby_time`. I think, I prefer `groupby` and `groupby_time`, as it's a good compromise between being intuitive and not needing to rename everything, and also keeping in mind that not all statistics have a time dimension.
4. Remove `bool` out of the mix of `bools`/ `strings` and `Callable` which can right now be passed.
    - `bool` is ambigious. What is group by/ aggregate by `True`? `False` means don't group? Or group by something else? 
    - Simply use `None` for no aggregation and `str` or `Callable` for some form of aggregation. For time aggregation, rather than simply using the binary `True`, this also allows specific aggregation methods to be specified (e.g. `"all"`, `"D"`, `"M"`, `"daily"`, etc.). We could use those of `df.resample` for that.
 
So the only and new proposed aggregation methods would be:
`groupby`, `groupby_method`, `groupby_time`, `groupby_time_method`
 
**Filtering**
`comps`, `bus_carrier` and `carrier` filter data. While `bus_carrier` and `carrier`, need to be there, since they filter is applied before aggregation (e.g. `groupby="bus", carrier="AC"`), I don't see a reason to keep `comps`. We could keep it for convenience, but the same output can always be achieved by slicing the MultIndex, since it is always added: `.loc[idx["Generator", :]]` instead of `comps="Generator"`. 

Similar to above, we could make the category clear. Above we have `aggregate_x` or `groupby_x`, we could rename here to `filter_bus_carriers` and `filter_carriers`. But not sure what I think about it. It depends on whether we would find it useful to add filter functionality for any other assets? This feature could also be achieved via a general filter method where more kinds could be added via `filter={"carrier": "AC"}`. And `carrier="AC"` can be kept as fast-access/ convenience.

5. Remove `comps` completely or at least rename it to `components`. What does `comps` even mean.
6. Do we have potentially more assets we might wanna filter on? E.g. more than `carrier` and `bus_carrier`? But this could be added as a additional feature.

**Other processing**
Then there are the general processing parameters `nice_names`, `drop_zero` and `round`, which are already mirrored to the module options. I wouldn't change anything here. However, with regard to #1034, I would change `nice_names` to make it compatible with any `nice_names` column so that we can add it and it doesn't matter if the statistic can be grouped by 'carrier'.

@fneum @FabianHofmann @p-glaum 
Let me know what you think, and please feel free to suggest anything else you might have in mind. We should take the opportunity to make any changes and redefinitions now with version 1.0, since afterwards we can only add additional features.

lkstrp:
Closes #1253

See linked issue.

## Checklist

### DIFF
--- a/pypsa/plot/statistics/maps.py
+++ b/pypsa/plot/statistics/maps.py
@@ -1,5 +1,6 @@
 """Maps plots based on statistics functions."""
 
+import warnings
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Literal
 
@@ -99,14 +100,20 @@ class MapPlotGenerator(PlotsGenerator, MapPlotter):
         non_transmission_carriers = n.c.carriers.static.index.difference(trans_carriers)
 
         # Get bus sizes from statistics function
-        bus_sizes = func(
-            bus_carrier=bus_carrier,
-            groupby=["bus", "carrier"],
-            carrier=list(non_transmission_carriers),
-            nice_names=False,
-            aggregate_across_components=True,
-            **(stats_kwargs or {}),
-        )
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "default",
+                message=".*Passing `aggregate_across_components` was deprecated.*",
+                category=DeprecationWarning,
+            )
+            bus_sizes = func(
+                bus_carrier=bus_carrier,
+                groupby=["bus", "carrier"],
+                carrier=list(non_transmission_carriers),
+                nice_names=False,
+                aggregate_across_components=True,
+                **(stats_kwargs or {}),
+            )
         if bus_sizes.empty:
             # TODO: this fallback case should be handled in the statistics function
             bus_sizes = (
@@ -137,7 +144,7 @@ class MapPlotGenerator(PlotsGenerator, MapPlotter):
         else:
             branch_flow_scaled = {}
             branch_widths = func(
-                comps=n.branch_components,
+                components=n.branch_components,
                 bus_carrier=bus_carrier,
                 groupby=False,
                 carrier=list(trans_carriers),
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -41,9 +41,14 @@ Features
 * The ``Network.add()`` method now returns ``None`` by default. Use ``return_names=True`` 
   to get the previous behavior of returning component names which have been added.
 
-* Refactored version attributes: ``__version_semver__`` → ``__version_base__``, 
+* Refactored version attributes: ``__version_semver__`` → ``__version_base__``,
   ``__version_short__`` → ``__version_major_minor__``. Removed tuple versions.
   Old names raise ``DeprecationWarning``.
+
+* Refined statistic arguments across optimization expressions and plotting modules.
+  Renamed ``comps`` → ``components``, ``aggregate_groups`` → ``groupby_method``,
+  and ``aggregate_time`` → ``groupby_time`` for consistency. Old argument names
+  are deprecated and will be removed in v2.0.0.
 
 
 Bug Fixes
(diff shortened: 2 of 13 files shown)


pypsa/plot/statistics/maps.py
```diff
@@ -99,14 +100,20 @@ def plot(
...
+                "default",
+                message=".*Passing `aggregate_across_components` was deprecated.*",
+                category=DeprecationWarning,
+            )
+            bus_sizes = func(
+                bus_carrier=bus_carrier,
+                groupby=["bus", "carrier"],
+                carrier=list(non_transmission_carriers),
+                nice_names=False,
+                aggregate_across_components=True,
```
fneum, line 114: The alternative would be a groupby operation at the end.

lkstrp: True, this is the for now lazy solution without messing around with different col names

Issue #1360, opened 2025-09-16
`e_initial` applied per period even with `e_initial_per_period=False`


fneum:
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/pypsa/releases) of PyPSA.

- [x] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch of PyPSA.

### Issue Description

The stores get replenished with `e_initial` at the beginning of each investment period, even if `e_initial_per_period=False`. The expected behaviour for this combination is that `e_initial` is set at the start of the first investment period and continuously depleted.

It appears that replenishment is performed based on the "period" snapshot index, despite the `e_initial_per_period=False` setting. Because even if `n.optimize(multi_investment_periods=False)` but a multi-indexed `n.snapshots` is used, the store is replenished with every change of the "period" index level.

This potentially also applies to storage units (where I did not test it).

This is not a new problem; I can also reproduce it on v0.29.

### Reproducible Example

```python
import pypsa
import pandas as pd

display(pypsa.__version__)

# without multi_investment_periods and no multi-indexed snapshots:
# expected n.stores_t.e [9, 8, 7, 6] -> correct
n = pypsa.Network()
n.snapshots = [1,2,3,4]

n.add("Bus", "bus")
n.add("Load", "load", bus="bus", p_set=[1, 1, 1, 1])
n.add("Store", "store", bus="bus", e_nom=10, e_initial=10, marginal_cost=1)

n.optimize(multi_investment_periods=False, solver_name='highs')
display(n.stores_t.p)
display(n.stores_t.e)

# without multi_investment_periods and multi-indexed snapshots:
# expected n.stores_t.e [9, 8, 7, 6] -> observed [9, 8, 9, 8]
n = pypsa.Network()
years = [2030, 2040]
timesteps = [1, 2]

n.snapshots = pd.MultiIndex.from_product([years, timesteps])
n.investment_periods = years

n.add("Bus", "bus")
n.add("Load", "load", bus="bus", p_set=[1, 1, 1, 1])
n.add("Store", "store", bus="bus", e_nom=10, e_initial=10, marginal_cost=1, e_initial_per_period=False)


n.optimize(multi_investment_periods=False) # still no multi-period optimization
display(n.stores_t.p)
display(n.stores_t.e)

# with multi_investment_periods and e_initial_per_period=True:
[...]

PR #1361, opened 2025-09-16
fix: resolve inconsistent period weighting for non-cyclic storage units in global constraints

### LINKED ISSUE #765: Inconsistent multi-period weighting for the primary energy limit global constraint
## Checklist

- [x] I am using the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch or the latest [release](https://github.com/pypsa/pypsa/releases). Please indicate:

## Describe the Bug

I just realized that the primary energy limit global constraints apply the number of years each period on-top of the regular generator snapshot weightings, but **only** for `generators`. And **not** for `storage_units`/`stores`.

ie. for generators the underlying assumption is that the dispatch repeats in each year of the investment period and therefore needs to be counted multiple times, while for storage units and stores only the difference between the last and the first level is counted without any weighting, which is inconsistent.

I had originally expected that the primary energy limit for a particular investment period would bound the **yearly** emissions, but i must admit the period weighting is necessary for an overall budget, then it is also necessary for a non-cyclic store/storage unit though.

@lisazeyen 

## Error Message

*If applicable, paste any terminal output to help illustrating your problem.*
*In some cases it may also be useful to share your list of installed packages: `conda list`.*

```
<paste here>
```

fneum:
closes #765

see release note for description

I think it's safest to throw an error for the combination of non-cyclic storage units/stores with `{e,state_of_charge}_initial_per_period = False` and investment periods year weightings other than 1, since this creates an inconsistency between the continuous storage state of charge and the contribution to the global constraint. Happy to hear alternative thoughts.

What I fixed here is the handling of non-cyclic storage units/stores with `{e,state_of_charge}_initial_per_period = True`. 

Note that while working on this, I detected another bug #1360 regarding the handling of `e_initial_per_period`. But this PR can be merged before fixing #1360.

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -100,6 +100,14 @@ Bug Fixes
 * Fix ``get_transmission_carriers()`` to handle components without carrier attribute (e.g., Transformer).
   (https://github.com/PyPSA/PyPSA/issues/1321)
 
+* Fix storage state-of-charge handling in multi-investment period optimizations. The constraint 
+  logic incorrectly determined when to apply per-period cycling vs. continuous storage state tracking, causing storage to behave unexpectedly regardless of flag settings. The fix ensures storage units and stores correctly preserve or reset their state across investment periods based on the ``cyclic_state_of_charge_per_period``/``e_cyclic_per_period`` and ``state_of_charge_initial_per_period``/``e_initial_per_period`` flags.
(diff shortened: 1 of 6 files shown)


coroa (approved):
LGTM

coroa:
I had forgotten about this bug

PR #1362, opened 2025-09-16
Optimize _sort_attrs for already ordered axes

koen-vg:

This is a fairly simple change, adding a few fast-paths when reading networks from an .nc file.

In a slightly strange model I was working with, loading the model was taking minutes because `_sort_attrs` always ran a full reindex, even when the DataFrame’s axis already matched the target order. I will say that this was in a large network with lots and lots of components but not really any time series data; a fair atypical use-case but still.

In pypsa/network/io.py, `_sort_attrs` now:

- Returns immediately for empty data, missing attr lists, or when no columns/index overlap with attrs_list.
- Compares the current axis order with the desired order and skips the reindex if they’re already aligned.
- Falls back to a single df.loc[...] realignment only when the order actually differs.

The behaviour should be the same (and on my side all the tests passed), but in the example of my slightly weird model it cut the network loading time down from over 3 minutes to below 10 seconds (we're talking a 50MB .nc file size so IO was not a dominating factor).

Let me know if this is good as is or if you want more documentation / testing / release note :)

## Checklist

### DIFF
--- a/doc/references/release-notes.rst
+++ b/doc/references/release-notes.rst
@@ -56,6 +56,8 @@ Features
   ``__version_short__`` → ``__version_major_minor__``. Removed tuple versions.
   Old names raise ``DeprecationWarning``.
 
+* Improve performance of loading networks by avoiding re-ordering dataframe columns and indices where unnecessary; especially impactful for networks with large numbers of components.
+
 
 Bug Fixes
 ---------
--- a/pypsa/network/io.py
+++ b/pypsa/network/io.py
@@ -1046,13 +1046,25 @@ def _sort_attrs(df: pd.DataFrame, attrs_list: list[str], axis: int) -> pd.DataFr
         Sorted DataFrame
 
     """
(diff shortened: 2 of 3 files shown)


lkstrp (approved):
Thanks @koen-vg 

Referencing #1128 once again, which would improve those things by a margin.

Could you add a very quick release note?

koen-vg:
I've added one more commit which is a slightly less "in-place" change. But I had additional performance issues, this time with solved networks (which I guess somehow circumvented the fast-paths I added earlier) and which I now managed to fix. There were two instances where _sort_attrs was called on a newly allocated "mock" dataframe just to retrieve the list of columns in the end; I've refactored it so _sort_attrs just acts upon an index instead of a whole dataframe. Performance-wise this now works well for me, and again I don't think the functionality is affected at all.

Let me know what you think; if this is all somehow made redundant soon by #1128 then maybe there isn't much of a use in merging this and I'm happy to use a local fork for the time being for my purposes. Thought it was better to improve on this PR while I was at it though.

PR #1363, opened 2025-09-17
feat: new `n[..]`

### LINKED ISSUE #1332: Add function to reduce stochastic network to one of its scenarios
Suppose you have a stochastic network with a couple of scenarios.
You maybe want to extract one of the scenarios and solve it as deterministic problem.
A function for that would be useful:

```py
n_scenA = n_stoch.get_scenario("scenA")
assert n_scenA.has_scenarios == False
```

lkstrp:
Closes #1332 

Adds `n.get_collection`, `n.get_scenario` and `n.slice_network` and combines them all in `n[..]`.

@coroa You wrote the original `__getitem__` ages ago and it seems nobody knew about its existance. Could you review and confirm that this was never a "shallow" slice?

## Checklist

### DIFF
--- a/pypsa/network/index.py
+++ b/pypsa/network/index.py
@@ -11,9 +11,13 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Sequence
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
+import numpy as np
 import pandas as pd
+
+if TYPE_CHECKING:
+    from pypsa import Network
 
 from pypsa.network.abstract import _NetworkABC
 
@@ -760,9 +764,7 @@ class NetworkIndexMixin(_NetworkABC):
         """Boolean indicating if the network has scenarios defined."""
         return len(self._scenarios_data) > 0
 
-    # -----------
     # Risk Preferences (CVaR)
-    # -----------
 
     def set_risk_preference(self, alpha: float, omega: float) -> None:
         """Set risk aversion preferences for stochastic optimization using CVaR formulation.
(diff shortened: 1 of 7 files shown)


pypsa/network/index.py
```diff
@@ -835,3 +837,325 @@ def risk_preference(self) -> dict[str, float] | None:
...
+
+        # Allow single selection without losing a dimension
+        if np.isscalar(buses):
+            buses = pd.Index([buses])
+
+        # Handle scalar snapshot selection
+        selected_snapshots = self.snapshots[snapshots]
+        if not isinstance(selected_snapshots, pd.Index):
+            # Single snapshot was selected, wrap in Index
+            selected_snapshots = pd.Index([selected_snapshots])
```
coroa, line 1098: At some point previously we supported fancy datetime based indexing:

```python
n["2024-03", :]
```

This would not work any longer if you use `__getitem__` of the index.

You would have to use something like:

```
inds = self.snapshots.get_indexer(snapshots)
if (inds == -1).any():
    raise KeyError(...)
selected_snapshots = self.snapshots[inds]
```
instead, i think. but would have to try myself.

pypsa/network/index.py
```diff
@@ -835,3 +837,325 @@ def risk_preference(self) -> dict[str, float] | None:
...
+            self._scenarios_data.loc[scenario]
+        except KeyError as e:
+            msg = f"Scenario '{scenario}' not found in network scenarios."
+            raise KeyError(msg) from e
+
+        n = self.copy()
+
+        n.name = f"{n.name} - Scenario '{scenario}'"
+        # Remove
+        n._scenarios_data = n._scenarios_data.iloc[:0]
```
coroa, line 915: I am unsure what type `n._scenarios_data` has; so cannot check.

coroa:
This has always been as shallow as `df.loc[...]` was, which means not by much with the boolean indexers.

pypsa/network/index.py
lkstrp: `n["2024-03", :]` is already broken. We could reimplement it, but via the `n.slice_networks(snapshots=..)`, which is a bit cleaner. We could also make use of it in the `__getitem__`, but it would require to be more robust to determine which indexing actually to apply based on the input

pypsa/network/index.py
lkstrp: All internal `_.._data`  attributes hold all scenario/ snapshot/ periods data as a `pd.DataFrame`. Also for scenarios it is not a Series, even if we currently just have a single column

lkstrp:
Thanks @coroa !
I'll merge this. See my comment above. Let's document this properly, make it less niche, and we can consider extending it more robustly after that

pypsa/network/index.py
coroa: Sure, this was meant for `slice_networks`.

PR #1371, opened 2025-09-23
fix: correct store/storage_unit state-of-charge behavior

### LINKED ISSUE #1360: `e_initial` applied per period even with `e_initial_per_period=False`

lkstrp:
Closes #1360

One of the reasons for this issue is that our mix of attributes overwrites each other without handling the defaults. At some point, we should implement a schema to disallow combinations that are not allowed instead of overwriting them.

## Checklist


pypsa/optimization/constraints.py
```diff
@@ -1310,14 +1310,14 @@ def define_storage_unit_constraints(n: Network, sns: pd.Index) -> None:
...
-    if isinstance(sns, pd.MultiIndex):
+    if n._multi_invest:
         # If multi-horizon optimizing, we update the previous_soc and the rhs
         # for all assets which are cyclic/non-cyclic per period
         periods = soc.coords["period"]
         per_period = (
             c.da.cyclic_state_of_charge_per_period
-            | c.da.state_of_charge_initial_per_period
-        )
+            & c.da.state_of_charge_initial_per_period
```
Irieo: This solves the problem in https://github.com/PyPSA/PyPSA/issues/1360, and the fix is tested unambiguously. However, I *think* the new version changes semantics and adds new problems. 
Old version conceptually tells 
`per_period = (cycle within period) OR (reset initial each period)`
New version tells
`per_period = (cycle within period AND reset initial) OR (cycle within period AND globally cyclic)`
I elaborate in comment below the potential problem with that

Irieo:
### We have three knobs

So we have three the boolean flags for a storage unit (analogous names for stores):

- [lets call it CP] `cyclic_state_of_charge_per_period` -- within every investment period, storage state trajectory should form a closed loop (Zero net energy change over each period)
- [lets call it IP]  `state_of_charge_initial_per_period` -- the user-provided `state_of_charge_initial` should be applied at the start of every investment period
- [lets call it C] `cyclic_state_of_charge` -- over the entire multi-period horizon, final snapshot SoC = very first snapshot SoC (global horizon closure). It converges to the logic of cyclic condition for a single-year optimization.

### The old version was partially wrong

The bug in https://github.com/PyPSA/PyPSA/issues/1360 is fortunately spotted and fixed.

The is another potential problem I guess if we had: CP = True, IP = False, C = False. Here we intend: “Let each period net to zero (cycle), and don’t reset the initial value each period.” We actually got a case where inside “per-period mode” the existing code suppresses the carry-over from the previous period’s last snapshot for non-cyclic assets by excluding the first snapshot of each period from `include_previous_soc_pp`.

### Great, what's the problem now?

What we do now: per_period = CP & (IP | C)

https://github.com/lkstrp/PyPSA/blob/eef1ff4897100753309d6156278fcb5ef53d22be/pypsa/optimization/constraints.py#L1313-L1320

This throws out two legitimate combinations of knobs, or?

- IP=True (CP & C=False, although we are in `_multi_invest`): we want a fixed initial level each period without cyclic storage
- CP=True, IP=False, C=False: we want simply zero net change each period.

I'll push tests now to showcase missing combinations

Irieo:
### Logic of a proposed fix

`per_period = cp | ip`

Where:
- `cp` = `cyclic_state_of_charge_per_period` (or `e_cyclic_per_period` for Store)
- `ip` = `state_of_charge_initial_per_period` (or `e_initial_per_period` for Store)

**If NOT `per_period`** (global optimization):
- Global behavior controlled by `c` (`cyclic_state_of_charge` or `e_cyclic`):
  - If `c` is True:
    - `include_previous` at ALL snapshots (global wrap end → start)
  - If `c` is False:
    - Exclude `include_previous` at the very first snapshot
    - Use initial level at RHS once

**If `per_period`** (per-period optimization):
- Within each period:
  - `include_previous` at all snapshots except the first snapshot of a period
  - If `cp` is True:
    - Also `include_previous` at the first snapshot of each period (wrap last → first within that period)
  - If `cp` is False AND `ip` is True:
    - Exclude `include_previous` at the first snapshot of each period (reset to initial)

### Changes

**BREAKING**: Modified default values in component attribute CSV files:

- `cyclic_state_of_charge_per_period`: Changed from `True` → **`False`** (StorageUnit)
- `e_cyclic_per_period`: Changed from `True` → **`False`** (Store)

**Why**: The old default of `True` adds a per-period wrapping behavior even when users do not explicitly want it. The new default of `False` is more intuitive.

### These combinations now work as intended

**`cp=False, ip=False, c=False`: Single global optimization period without cyclic storage**
- Inject initial level once at the first snapshot
- The mask `include_previous_soc` is False at the very first active snapshot for non-cyclic assets
- The constraint injects `soc_init` set by user if provided

**`cp=False, ip=False, c=True`: Single global optimization period with global cyclic storage**
- Wrap end→start globally
- No initial level injection
- NB so this means `c` flag works for multi-period opt too, but only when `cp` is False and when `ip` is false. **If either of two is True, they overrule `c` flag.**

**`cp=True, ip=False, c=any`: Per-period cyclic storage**
- Each period wraps end→start
- No initial injection at period starts
- State of charge cycles within each investment period
- NB **if c=True it gets overruled by the cp=True** (so effectively c has no effect if CP=True). I will add warning about this too.

**`cp=False, ip=True, c=any`: Per-period non-cyclic storage**
- Each period starts from its initial state
- No wrap within periods
- State resets to `state_of_charge_initial` at the beginning of each period

**`cp=True, ip=True, c=any`: Both per-period cyclic AND per-period initial set**
- Cyclic takes precedence over user-defined initial level at period starts
- We do wrapping, no initial injection at period boundaries
- **A warning is issued** to inform users that `cyclic_state_of_charge_per_period=True` overrides their `state_of_charge_initial` values at period boundaries
- NB **if c=True it gets overruled by the cp=True** (so effectively c has no effect if CP=True). I will add warning about this too.

Irieo:
well, and one more thing I realise -- we have to reconcile what happens if user sets both `c` = True and `cp` = True, what is correct?
a) we cast two constraints wrapping ends in each period and the optimization horizon overall or 
b) we add one constraint for periods (and ignore overall). 
We do the latter one (historically and also follow doing so in this fix)
This does not mean c flag can be used only for one-period optimization, if  `cp` = False and `ip` = False, user can solve one concatenated optimization problem and wrap first/last periods with `c` = True.
We just have to be explicit about the overrules. I will a logger message for this as well (and can volunteer to patch docs...)

pypsa/optimization/constraints.py
```diff
@@ -1351,6 +1372,45 @@ def define_storage_unit_constraints(n: Network, sns: pd.Index) -> None:
...
+            if cp_overrides_c.any():
+                affected = c.static.index[cp_overrides_c.values].tolist()
+                logger.warning(
+                    "StorageUnits %s: Per-period cyclic (cyclic_state_of_charge_per_period=True) "
+                    "overrides global cyclic (cyclic_state_of_charge=True). "
+                    "Storage will cycle within each investment period, not across the entire horizon.",
+                    affected,
+                )
+    except Exception as e:
+        logger.debug("Could not check for ignored initial states: %s", e)
```
lkstrp: > logger.debug("Could not check for ignored initial states: %s", e)

Why would that be the case?

Irieo: you are right, whatever the case is it should be captured by our type checks not try/except. I remove it.

PR #1386, opened 2025-10-09
fix(statistics): allow at_port to be specified manually

coroa:

For retrieving statistics in respect to a specific port, we'd want the
`at_port` statistics argument be able to accept a single port or a list of
ports (as the type arguments already said they would).

This means that for links with an efficiency of 0.9, they could be accounted for at the output capacity, for instance:

```python
In [29]: n.optimize.expressions.capacity("Link", at_port="1")
Out[29]:
LinearExpression [carrier: 1]:
------------------------------
[DC]: +0.9 Link-p_nom[Norwich Converter] + 0.9 Link-p_nom[Norway Converter] + 0.9 Link-p_nom[Bremen Converter] + 0.9 Link-p_nom[DC link]
```

@FabianHofmann Does that make sense?

It's still marked as a draft, as we'll have to add documentation first.

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,7 +13,7 @@ SPDX-License-Identifier: CC-BY-4.0
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.-->
 
-- Statistics functions correctly interpret `at_port` as a port number. (<!-- md:pr 1386 -->)
+- Fix inconsistent `at_port` handling in statistics functions. The parameter now accepts integer port numbers and `"all"` with proper port resolution via component methods. Passing `True`/`False` is deprecated. (<!-- md:pr 1386 -->)
 
 ## [**v1.1.0**](https://github.com/PyPSA/PyPSA/releases/tag/v1.1.0) <small>17th February 2026</small> { id="v1.1.0" }
 


coroa:
The match statement is consistent with the minimal python version 3.10. https://peps.python.org/pep-0636/

FabianHofmann:
yes, I like that feature very much!

gianvicolux:
I added a `raise ValueError` when `at_port` is defined by means of the prefix "bus". I also updated the docstrings in [pypsa/statistics/expressions.py](https://github.com/open-energy-transition/PyPSA/blob/5c7231e1295399c80b8e1d572561a7ada44554bc/pypsa/statistics/expressions.py#L515-L519).

Now, the complete documentation must be updated, too.

@coroa @FabianHofmann

pypsa/statistics/expressions.py
```diff
@@ -633,6 +634,7 @@ def capex(  # noqa: D417
             - True: All ports of components
             - False: Exclude first port ("bus"/"bus0")
```
coroa: Should be
```
            - False: Consider only first port ("bus"/"bus0")
```

coroa:
@gianvicolux I think if you add a small release note this is ready for review.

pypsa/statistics/abstract.py
```diff
@@ -156,8 +156,23 @@ def _aggregate_components(
...
+
+            match at_port:
+                case str():
+                    if at_port in ports:
+                        ports = [at_port] if at_port in ports else []
+                    else:
+                        ports = []
+                        if at_port.startswith("bus"):
+                            msg = f"{at_port} with 'bus' prefix is not valid. Use the port number only."
+                            raise ValueError(msg)
```
coroa:
```suggestion
                    if at_port in ports:
                        ports = [at_port]
                    elif at_port.startswith("bus"):
                        msg = f"{at_port} with 'bus' prefix is not valid. Use the port number only."
                        raise ValueError(msg)
                    else:
                        ports = []    
```

coroa (commented):
Ok, with this it's good from my side.

lkstrp (changes requested):
I didn't notices that `at_port` is just this:
```
if not at_port:
    ports = [ports[0]]
```
which makes this PR a bug fix and not a feature.

Since we are touching it already, I'd make more changes and turn it into a bug fix and a new feature:

```
Which ports to consider:
- True: All ports of components
- False: Exclude first port ("bus"/"bus0")
- False: Consider only first port ("bus"/"bus0")
- str or list of str: Specific ports to include
    Note: use port number only, without 'bus' prefix (e.g., '1' instead of 'bus1')
```
is mixing types and is not intuitive.
What about removing the boolean and just allowing a list of ports? The default could be set to `None`, which would map to c.ports. `'bus'` and `'bus0'` can be treated equally. 

We are also now in 1.0 territory, which means:
- that we want to still allow `True` and `False` and map them to all ports or just the first one, raising a deprecation warning.
- don't wanna just raise a `ValueError` for `at_port="bus0"` which very buggy worked before, but would fail now.

Also why kicking the `'bus'` from `'bus1'`, but only allowing `'1'` but not `1` as int? Because that's what is returned from `c.ports`?

doc/references/release-notes.rst
```diff
@@ -15,6 +15,8 @@ Upcoming Release
 Features
 --------
 
+* New feature for retrieving statistics in respect to a specific port: ``at_port`` statistics argument is able to accept a single port or a list of ports (as the type arguments already said they would). Note: use port number only, without 'bus' prefix (e.g., '1' instead of 'bus1') (https://github.com/PyPSA/PyPSA/pull/1386).
```
lkstrp: Remove this file. This needs to go in `docs/release-notes.md` now and with markdown formatting so (`-` for bullet points, `(<!-- md:pr 1386 -->)` to link the pr and single '`' for code formatting.

test/test_optimization_expressions.py
```diff
@@ -176,3 +176,16 @@ def test_expressions_operation(prepared_network):
     expr = n.optimize.expressions.operation()
     assert isinstance(expr, LinearExpression)
     assert expr.size > 0
+
+
+def test_concrete_at_port(prepared_network):
```
lkstrp, line 266: Add test also for statistics expressions, not just optimization

coroa:
> I didn't notices that `at_port` is just this:
> 
> ```
> if not at_port:
>     ports = [ports[0]]
> ```
> 
> which makes this PR a bug fix and not a feature.

when i touched it first, i considered it as a feature, i only noticed the docstrings late in the process. It's true that the docstring is not implemented.

> Since we are touching it already, I'd make more changes and turn it into a bug fix and a new feature:
> 
> ```
> Which ports to consider:
> - True: All ports of components
> - False: Exclude first port ("bus"/"bus0")
> - False: Consider only first port ("bus"/"bus0")
> - str or list of str: Specific ports to include
>     Note: use port number only, without 'bus' prefix (e.g., '1' instead of 'bus1')
> ```
> 
> is mixing types and is not intuitive. What about removing the boolean and just allowing a list of ports? The default could be set to `None`, which would map to c.ports. `'bus'` and `'bus0'` can be treated equally.

I think, you need an option to specify all ports (across multiple component types) and this should not be the default, because it easily leads to double-counting and it was not the default up to now.

What about

    at_port: Literal["all"] | int | Sequence[int] = [0]

as preferred type and thus deviating from `c.ports` (which i don't like).


> We are also now in 1.0 territory, which means:
> 
> * that we want to still allow `True` and `False` and map them to all ports or just the first one, raising a deprecation warning.

Sure.

> * don't wanna just raise a `ValueError` for `at_port="bus0"` which very buggy worked before, but would fail now.

This did not work before, it returned effectively the result of `at_port=c.ports` including double-counting potential and all. I think people should know they made a major mistake.

> Also why kicking the `'bus'` from `'bus1'`, but only allowing `'1'` but not `1` as int? Because that's what is returned from `c.ports`?

Indeed, I like consistent naming and `c.ports` is the closest official definition i know. It's not a definition i like, but ambiguity is worse imho.

lkstrp:
I am happy with:
> at_port: Literal["all"] | int | Sequence[int] = [0]

And I agree, `c.ports` is not great. It returns `""` for single ports right now. Which is good for primitive column name building, but not as a argument set. What would you change there? Just switching it to return `0, 1, 2 ..` ?

coroa:
> I am happy with:
> 
> > at_port: Literal["all"] | int | Sequence[int] = [0]
> 
> And I agree, `c.ports` is not great. It returns `""` for single ports right now. Which is good for primitive column name building, but not as a argument set. What would you change there? Just switching it to return `0, 1, 2 ..` ?

I think either numbers 0, 1, 2 would be good or actual attribute names: `["bus"]` or `["bus0", "bus1"]`. What about having a second `c.port_attrs` or `c.port_cols` which returns the latter in additon?

lkstrp:
`c.port_cols` would clutter the API a bit, but since we won't align `"bus0"` with `"bus"` any time soon, I am sold

coroa:
Hmm .. i thought some more. There will be recurringly the need to accept both types of inputs, `port_cols` and `port_nums`. What about a new dataclass:
```python
type PortsLike = Literal["all"] | int | Sequence[int] | str | Sequence[str]

@dataclass
class Ports:
    nums: list[int]
    c: Component

    @classmethod
    def from_like(cls, ports: PortsLike, c: Component):
         match ports:
              ...
  
          return cls(nums, ct)

    def __iter__(self):
        return iter(self.nums)

    def cols(self):
        ...
```

and `c.ports` returns that instead. Not sure about transition path. What do you think?

lkstrp:
I think we should keep it simple. Returning a data class could cause confusion for a use case as simple as this. It also adds overhead. Accepting both types can be done everywhere with a simple `ports = _as_ports(ports)` or an internal data class if you think that's better. There are not even that many use cases for it right now, just for statistics I think? Do you have any other potential stuff in mind where we need ports?

coroa:
Not sure what's the difference between internal, external. here. I'd expect the overhead to be negligible.

You'll need at least two methods `_as_ports(:PortsLike) -> Sequence[int]` and `_as_port_cols(:PortsLike, :Component) -> Sequence[str]`.

My thinking was that the dataclass would make that more discoverable, but the use is mainly internal anyway.

The problem with ports also comes up frequently when formulating custom constraints, so making that easier would be a good thing.

coroa:
I still need something like this, is there a particular reason that you did close the PR?

coroa:
@lkstrp Ok, i updated that with a mostly working proposal. We need some sort of method proxying on network collections for the new `_as_ports(...)` method on the Component attribute which converts `PortsLike` to a list of int. Could you help me with that?

lkstrp:
Maybe I should more often ping close PRs 😅

Sure, I can do. Could you clean up the diff first?

coroa:
Diff was against wrong branch. Should be fine now.

coroa:
@lkstrp Happy if you can have a look. The diff was just against the wrong branch.

FabianHofmann:
Nice refactoring, there is one thing we have to be aware of though. 

If I see correctly, there is now an inconsistency how `bus_carrier` and `at_port` interact. When `bus_carrier` is specified the expression should loop over all ports and filter out the ones matching the bus carrier. For this case, `at_port` should automatically resolve to `"all"`.. Now, most functions have `at_port` "hardcoded" to `"bus0"`, meaning `bus_carrier` filtering silently misses connections on other ports. 

We could instead keep `None` as the `at_port` default  and centralize the implicit default in `_aggregate_components`.

pypsa/components/components.py
```diff
@@ -734,6 +737,69 @@ def ports(self) -> list:
             match.group(1) for col in self.static if (match := RE_PORTS.search(col))
         ]
 
+    def _as_port(self, port: int | str) -> int:
```
lkstrp, line 748: Can we merge `_as_port` with `_as_ports`?

pypsa/statistics/abstract.py
```diff
@@ -154,27 +154,39 @@ def _aggregate_components(
...
-            ports = [
-                match.group(1)
-                for col in n.c[c].static
-                if (match := RE_PORTS.search(str(col)))
-            ]
-            if not at_port:
-                ports = [ports[0]]
+            if at_port is False or at_port is None:
+                warnings.warn(
+                    f"Passing `at_port={at_port}` is deprecated. Use `at_port=0` instead.",
```
lkstrp: Add version 1.2 and removed in 2.0, just check format of other deprecation warnings

pypsa/statistics/abstract.py
```diff
@@ -154,27 +154,39 @@ def _aggregate_components(
...
+                    'Passing `at_port=True` is deprecated. Use `at_port="all"` instead.',
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                at_port = "all"
 
             values = []
-            for port in ports:
-                vals = func(n, c, port)
+            for port in c._as_ports(at_port):
```
lkstrp, line 195: Why do we not just return the correct col names in `_as_ports`? Why the integer? Is it of any use anywhere? We always need to cast it again anyway. Then we could also expose that function

coroa: The main problem is that you also do need the associated efficiency column `efficiencyN` (or `rateN` for processes) and dispatch attribute `pN` for a bus column `busN`. That's why i was suggesting a minimal port dataclass (or maybe enum) so that you can iterate and get all column names.

The integer seemed the cleanest way to encode this port number and it was easy from that one to get to the dispatch and the efficiency.

pypsa/components/components.py
coroa: I had it merged but then disentangled it again, because `port_efficiency` which is sort of official API also had to support this flexibility:
https://github.com/open-energy-transition/PyPSA/blob/1cb1500a6a9cd3dc18fd1f430b440622202d8520/pypsa/statistics/expressions.py#L43-L66

pypsa/statistics/abstract.py
lkstrp: Very true, sorry my bad, I should have remind me what is going on here

coroa:
> Nice refactoring, there is one thing we have to be aware of though.
> 
> If I see correctly, there is now an inconsistency how `bus_carrier` and `at_port` interact. When `bus_carrier` is specified the expression should loop over all ports and filter out the ones matching the bus carrier. For this case, `at_port` should automatically resolve to `"all"`.. Now, most functions have `at_port` "hardcoded" to `"bus0"`, meaning `bus_carrier` filtering silently misses connections on other ports.
> 
> We could instead keep `None` as the `at_port` default and centralize the implicit default in `_aggregate_components`.

I think, i only chose `at_port="bus0"` where there was `at_port=False` before. An example for the previous `at_port=None` is `optimal_capacity`:

There i have `at_port=None` as default,

https://github.com/PyPSA/PyPSA/pull/1386/changes#diff-ce7e31f541948b44c5a77e2e6d35c8ba65265254f2d1e97174cd234b2a548cdaR1112

and then a:
```python
if at_port is None:
    at_port = "all" if bus_carrier else "bus0"
```
https://github.com/PyPSA/PyPSA/pull/1386/changes#diff-ce7e31f541948b44c5a77e2e6d35c8ba65265254f2d1e97174cd234b2a548cdaR1191-R1192

Where do you think there is an inconsistency?

lkstrp:
Now I am confused. This PR is also on an old master where docstrings did not match the signature.

You indeed don't introduce any inconsistencies, they have been in there before. In my understanding we should just default to `None` in any expression/ statistic which also takes `bus_carrier`. This gets then always correctly resolved via

``` python
if at_port is None:
    at_port = "all" if bus_carrier else "bus0"
```

`capex` for example was defaulting to `False`, you changed it now to `"bus0"` which is fine. But it should have been `None` in the first place. So this is an additional bug fix.

coroa:
> Now I am confused. This PR is also on an old master where docstrings did not match the signature.

Is it? `capex` on the latest master is:

https://github.com/PyPSA/PyPSA/blob/e37b04f8f3c14f182e88b49cb3baa236f9204422/pypsa/statistics/expressions.py#L592-L632

and on this PR:

https://github.com/open-energy-transition/PyPSA/blob/1cb1500a6a9cd3dc18fd1f430b440622202d8520/pypsa/statistics/expressions.py#L601-L641


Where do you think is the inconsistency?

If you mean the difference in treatment between capex, optimal_capacity and energy_balance in regards to `at_port` and `bus_carrier` is convoluted, i wholeheartedly agree. I did not understand why one would want to choose the defaults as they are, but i assumed there would be a good reason.

Sounds like there might not be one. Maybe @FabianHofmann, you can clarify if there was intent. Or if you think your rule above:

> We could instead keep None as the at_port default and centralize the implicit default in _aggregate_components

should actually be general.

FabianHofmann:
@coroa sorry for being late here. the problem is that when specifying `at_port=bus0` and not to `False` as before the loop over the ports will not go beyond bus0. that is when the user wants to extract the information for a bus_carrier wo explicitly passing `at_port=False` or `None`, the code will only check if bus0 has that bus carrier. does that make sense? 

EDIT: Just wrote a test that checks the behavior, it confirms that this would be a breaking change. Fine if I push it?

coroa:
@FabianHofmann Feel free to push the test.

pypsa/statistics/expressions.py
```diff
@@ -950,8 +963,12 @@ def overnight_cost(  # noqa: D417
             Whether to aggregate across components.
         groupby : str | Sequence[str] | Callable, default="carrier"
             How to group components.
-        at_port : bool | str | Sequence[str], default=False
-            Which ports to consider.
+        at_port : PortsLike | None, default=None
+            Which ports to consider:
+            - None: Automatically set to "all" if bus_carrier is specified, otherwise "bus0"
+            - "all": All ports of components
+            - "bus0": Consider only first port
```
FabianHofmann, line 970: this line (coming from me) does not add a lot. happy to remove it again

PR #1396, opened 2025-10-14
fix: Near-opt space example notebook cleanup

koen-vg:
Since the near-optimal spaces example is kind of my baby, I thought I'd quickly submit this, which cleans up the output significantly.

There are also a bunch of these warnings:
```
/home/docs/checkouts/readthedocs.org/user_builds/pypsa/envs/latest/lib/python3.13/site-packages/linopy/expressions.py:2006: FutureWarning:

In a future version of xarray the default value for join will change from join='outer' to join='exact'. This change will result in the following ValueError: cannot be aligned with join='exact' because index/labels/sizes are not equal along these coordinates (dimensions): 'name' ('name',) The recommendation is to set join explicitly for this case.
```

I can't quite replicate this locally so I'm not sure how they might be cleaned up from the documentation.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,21 +99,24 @@ jobs:
       run: |
         import pypsa
         # ac-dc-meshed
-        n = pypsa.examples.ac_dc_meshed("examples/networks/ac-dc-meshed/ac-dc-meshed")
+        n = pypsa.Network("examples/networks/ac-dc-meshed/ac-dc-meshed")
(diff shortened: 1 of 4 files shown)


lkstrp:
> In a future version of xarray the default value for join will change from join='outer' to join='exact'. This change will result in the following ValueError: cannot be aligned with join='exact' because index/labels/sizes are not equal along these coordinates (dimensions): 'name' ('name',) The recommendation is to set join explicitly for this case.

That's a linopy issue that we should address. It's the last deprecation warning that hasn't been cleaned up

PR #1406, opened 2025-10-16
Change max_relative_growth unit to nan

lisazeyen:
Closes # (if applicable).

Correct unit for `max_relativ_growth` 

## Checklist


coroa:
Why nan? Will this not break existing uses of only adjusting `max_growth`?

lisazeyen:
> Why nan? Will this not break existing uses of only adjusting `max_growth`?

The `nan` is for the unit and not the default value. It is a relative factor without any unit which is getting multiplied with e.g. the installed capacity of the previous investment period

PR #1423, opened 2025-10-24
minimum cycle basis support

### LINKED ISSUE #1422: Minimum cycle basis for KVL
https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.minimum_cycle_basis.html

https://arxiv.org/abs/2510.03484

fneum:
closes #1422

## Background

CANOPI paper suggests using minimum cycle basis: https://arxiv.org/abs/2510.03484

There is a drop-in replacement for the `nx.cycle_basis()` we currently use in `networkx`, `n.minimum_cycle_basis()` which had bugs fixed in NetworkX 3.2.

https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.minimum_cycle_basis.html

However, it is pretty slow and there is some (untested) trade-off between time spent on finding the most sparse formulation and solving the model:

Image

Right now, this is what's implemented as an option for `find_cycles()`, but not called by default.

In this PR, the LP algorithm (equation 26, Algorithm 3) that the authors of the CANOPI paper propose should also be implemented here (or directly in NetworkX if some optimisation dependencies are already there).


FabianHofmann:
@fneum I think this is a sensible thing to do. I added some tests to check that it produces plausible results. the question is now whether we want to enable the minimum cycle bases in the `pf` and `constraints` module (with `x_pu` as weights)

coroa:
If i understood correctly the problem is that this is also more time intensive than the "naive" cycle basis construction in networkx, so that you would ideally want to do it once and persist it on the network.

@lkstrp Maybe you can account for such persistence in your io refactoring?

pz-max:
- this PR continues after the data store work is addressed
- because cycle mapping need to be stored first to reduce compute which requires a new sensible place to store the data

Then @coroa might address this PR

fneum:
Closing in favour of #1898

PR #1428, opened 2025-10-31
fix: fix ramping-/rolling horizon logic

### LINKED ISSUE #1424: Ramp limit links in 1.0 fails

### Issue Description

In 0.35, ramp_limit_up for comittable links gives an expression which limits ramping unless the link status changes, works.
```
[hour_1, electrolyzer]: +1 Link-p[hour_1, electrolyzer] - 1 Link-p[hour_0, electrolyzer] + 80 Link-status[hour_0, electrolyzer] - 100 Link-status[hour_1, electrolyzer]      ≤ -0.0
```
In 1.0, that fails it seems. I think I tracked it down to:
```
    # ---------------- Check if ramping is at start of n.snapshots --------------- #

    attr = {"p", "p0"}.intersection(c.dynamic.keys()).pop()
    start_i = n.snapshots.get_loc(sns[0]) - 1
    p_start = c.dynamic[attr].iloc[start_i]

    # Get the dispatch value from previous snapshot if not at beginning
    is_rolling_horizon = sns[0] != n.snapshots[0] and not p_start.empty
    p = m[f"{c.name}-{attr}"]
```
The attr yields p0 here, and then the function tries to formulate something with "Link-p0" which doesn't exist (should be Link-p).

### Reproducible Example

```python
see above
```

### Expected Behavior

see above

### Installed Versions

<details>
Replace this line.
</details>

Irieo:
Closes #1424

This PR fixes several issues we have in ramping constraints. Citing @lkstrp, it is another case of our very annoying `p == p0 but != p0`

the original point was whether `p0` here is redundant: https://github.com/PyPSA/PyPSA/blob/6f70f53f44bc27108219856e5a197b9ad1f46cae/pypsa/optimization/constraints.py#L608

A: No. Although during optimization it indeed `-p` (dynamic attribute `p0` does not exist at point of optimization and is created after solve in `assign_solution`). However, there is a case when `define_ramp_limit_constraints` is applied in rolling horizon mode. In this case we have to refer to previous period's dispatch (stored in `p0` for a Link specifically). 

- One problem here that our lovely test suite never had Link & ramp_limits together.
- If I add such test, the constraint breaks because of a buggy logic: c.dynamic.keys() assumes that at least one of {"p", "p0"} there, but on the first optimization run, nothing exists yet in the dynamic data for Links. 
- Furthermore, the constraint breaks also in a standard n.optimize() call (w/o RH) because function tries to access p0 before checking whether we are at all in the rolling horizon mode.
- And finally, docstring is wrong. Ramping limits are applied to attributes that have ramping attributes, which are only Generator and Links.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "pandas>=0.24",
     "xarray<=2025.9.0",
     "netcdf4",
-    "linopy>=0.5.5",
+    "linopy>=0.5.5,<=0.5.7",
     "matplotlib",
     "plotly",
     "pydeck",


Irieo:
🟢

fneum (approved):
LGTM

FabianHofmann:
Just stumbled over this bug,  and there comes the fix 😇 Thanks @Irieo

PR #1437, opened 2025-11-10
fix: snapshot selection for operational in multi period optimization

FabianHofmann:
Fix the wrong selection of last snapshot per period in case of investment period selection on operational limit constraint

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -5,21 +5,21 @@ SPDX-License-Identifier: CC-BY-4.0
 -->
 
 # Release Notes
-<!--## Upcoming Release
+## Upcoming Release
 
 !!! info "Upcoming Release"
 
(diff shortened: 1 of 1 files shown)


Irieo:
@FabianHofmann I made a small problem instance trying to understand/recreate the bug and realised that (i) bug was indeed there and (ii) the fix was incomplete because period weightings had the same problem (mismatch between global and investment period index). So I append the fix and add unit test for that.

Irieo (approved):
See my comment in PR discussion.

FabianHofmann:
> @FabianHofmann I made a small problem instance trying to understand/recreate the bug and realised that (i) bug was indeed there and (ii) the fix was incomplete because period weightings had the same problem (mismatch between global and investment period index). So I append the fix and add unit test for that.

awesome! thanks @Irieo !

PR #1438, opened 2025-11-12
fix: masking of nan constraints in kvl

FabianHofmann:

Apply KVL masking and bump linopy version. 

@lkstrp @Irieo the "problem" with the new linopy version (and yes it should have been version 0.6.0) is that it checks more rigorously on the constraints sanity. PyPSA's KVL constraint had nan's in the right hand side for a subset of constraints. These result from different sets of active branches across the periods. Having nan rhs values is formally incorrect, so linopy raising an error is what we want here. I have added a mask which filters out the unwanted constraints on the linopy side. 

So https://github.com/PyPSA/linopy/pull/520 is not needed and I still need to think about whether we want to have it in.  

## Checklist

### DIFF
--- a/docs/user-guide/optimization/stochastic.md
+++ b/docs/user-guide/optimization/stochastic.md
@@ -109,7 +109,7 @@ dtype: float64
 >>>
 >>> obj_deterministic = n.objective / 1e9
 >>> obj_deterministic  # doctest: +ELLIPSIS
-5.3574849734203...
+5.357...
 ```
 
 ### Scenario Definition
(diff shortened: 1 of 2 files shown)


lkstrp (approved):
Ah, it can be simple as that! thanks a lot @FabianHofmann !

I wasn't sure if we had more nans set, which linopy had previously handled for PyPSA. But you're right, it's better to be strict and fix it here

FabianHofmann:
@lkstrp sure thing! should I also fix the unrelated doc test here which seems to be related to  a missing carrier in one example?

lkstrp:
All good, I will fix it. Wanna check if we can unpin `xarray` with `linopy==0.5.8`

fneum (approved):

PR #1443, opened 2025-11-17
ci: use pixi for model tests

lkstrp:


lkstrp:
@brynpickering In here, we wanna run the same PyPSA-Eur test workflow (integration and unit tests), but install instead the current ref of PyPSA. I did a couple of unsuccessful tries and errors, and before it gets hacky, could you point me to the right way to do this with `pixi`?

I was thinking that the below does the job, but the lock file always blocks it or is created and then blocks it:
```
pixi add --pypi --git https://github.com/PyPSA/PyPSA.git --rev ecff0543638b26a217059d2b5215685b194c95d5 PyPSA
```

lkstrp:
Ref #1446

.github/workflows/test-models.yml
```diff
@@ -70,45 +77,21 @@ jobs:
...
     - name: Install package from ref
       if: env.pinned == 'false'
       run: |
-        python -m pip install git+https://github.com/${{ github.repository }}@${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
-        conda list
-
-    - name: Check pypsa version
-      run: |
-        python -c "import pypsa; print(pypsa.__version__)"
+        rm -f pixi.lock
```
brynpickering:
```suggestion
        pixi remove pypsa
```

You need to get rid of pypsa entirely as a conda dependency otherwise adding it from pypi in the next line will create the error you're seeing. No need to delete the lock file; it will be regenerated when doing this removal and the following addition.

lkstrp: Ahhh! I tried this but of course we are case sensitive here and I can't just take the repo name which is `PyPSA`. Thanks for catching!

.github/workflows/test-models.yml
```diff
@@ -80,7 +80,7 @@ jobs:
     - name: Install package from ref
       if: env.pinned == 'false'
       run: |
-        pixi remove pypsa
+        pixi remove ${{ lower(github.event.repository.name) }}
```
brynpickering: This command is removing the _package_ so it should stay as `pypsa`, no?

lkstrp: I just want it to be generalised so that we can copy and paste the workflow to e.g. linopy. But I've added a package name environment variable now

PR #1444, opened 2025-11-18
Unpin shapely upper bound

brynpickering:
The `shapely` upper pin is causing issues when using `PyPSA` in an environment when a newer `shapely` version's feature is required. 

Looking at the file blame, @FabianHofmann introduced the pin, possibly to handle deprecation warnings? Tests all pass and no shapely warning are raised in a version with this unpinned.

## Checklist


FabianHofmann:
@brynpickering yes I think I was even a "bug" or incompatibility on the shapely side. great to see it is working now

PR #1447, opened 2025-11-24
ci: free up disk space

lkstrp:
Hacky and I don't like it, but this resolves the disk space issues and gives us an extra ~20GB. The model runs within `PyPSA` are making problems again. Those in `PyPSA-Eur` not yet, but I guess it's just a matter of time.

This is just following 
https://carlosbecker.com/posts/github-actions-disk-space/

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -118,6 +118,10 @@ jobs:
           results
         retention-days: 3
 
+    - name: Show remaining disk space
+      if: always()
+      run: df -h
+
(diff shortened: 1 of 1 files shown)


coroa:
I have also repeatedly caused  out-of-disk-space issues in PyPSA-Eur, so this would also be worthwhile to add there.

Definitely a hack, but with rather limited shoot-yourself-in-the-foot surface, so i actually do like it.

I also think the article's suggestion to add a `df -h` as final job is actually a good way to keep in mind how close the wall is.

PR #1451, opened 2025-11-26
fix: expressions returning zero for non-extendable generators

### LINKED ISSUE #1316: Curtailment expression returns zero for non-extendable generators

### Issue Description

The `curtailment()` function in PyPSA's expressions module returns zero (`LinearExpression +0`) when applied to networks with only non-extendable generators, even when actual curtailment is occurring. 

### Reproducible Example

```python
import pandas as pd
import pypsa

n = pypsa.Network()
n.set_snapshots(pd.date_range("2024", periods=4, freq="h"))

n.add("Carrier", ["AC", "solar", "gas"])
n.add("Bus", "bus0", carrier="AC")
n.add(
    "Generator",
    "solar0",
    bus="bus0",
    p_nom=100,
    p_nom_extendable=False,
    marginal_cost=0,
    carrier="solar",
    p_max_pu=[1.0, 0.8, 0.6, 0.4],
)
n.add("Generator", "gas0", bus="bus0", p_nom=100, marginal_cost=50, carrier="gas")
n.add("Load", "load0", bus="bus0", p_set=[50, 50, 50, 50])

n.optimize.create_model()

# Test curtailment expression
curtailment_expr = n.optimize.expressions.curtailment(
    comps=["Generator"], aggregate_time=False
)
print(f"Expression result: {curtailment_expr}")
```

### Expected Behavior

The `curtailment()` expression should return a LinearExpression representing the curtailment (available generation minus actual generation) for all generators, including non-extendable ones. 

### Installed Versions

<details>
Replace this line.
</details>

FabianHofmann:
## Summary

Fix `curtailment()`, `capacity()`, and `capex()` expressions returning zero for networks with only non-extendable generators.

Closes #1316

## Root Cause

The methods accessed optimization model variables directly (`n.model.variables[f"{c}-{attr}"]`). For non-extendable generators, this variable doesn't exist since there's nothing to optimize. The `@pass_none_if_keyerror` decorator caught the KeyError and returned None, resulting in zero expressions.

## Solution

- Check if the optimization variable exists before accessing it
- Create a constant `LinearExpression` for non-extendable components when the variable doesn't exist
- Use `comp.fixed` property and `comp._operational_attrs["nom"]` for cleaner, more idiomatic code
- Removed unused `nominal_attrs` import

## Test Plan

- [x] Added unit tests
- [x] ruff check and format pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,6 +119,7 @@ filterwarnings = [
 
     # Ignore linopy warnings
     "default:Coordinates across variables not equal. Perform outer join.:UserWarning",
+    "ignore:In a future version of xarray the default value for join will change:FutureWarning",
 ]
 markers = [
     "mpl_image_compare",


pypsa/optimization/expressions.py
```diff
@@ -256,11 +270,25 @@ def capacity(
...
+            # Get non-extendable capacity using component's fixed property
+            non_ext_capacity = (
+                comp.static.loc[comp.fixed, nom_attr]
+                if include_non_extendable
+                else pd.Series(dtype=float)
+            )
+
+            # Build capacity expression handling both extendable and non-extendable
+            if var_name in m.variables:
+                capacity = m.variables[var_name] + non_ext_capacity
```
Irieo, line 286: I think you're aware of it, just flagging to be sure: when testing these expressions, I get an xarray warning in this place where we add a linopy Variable to a pandas series (here: non-extendable capacity), and linopy internally handles mixes var/par expressions in a `merge()` function.

> In a future version of xarray the default value for join will change from join='outer' to join='exact'. This change will result in the following ValueError: cannot be aligned with join='exact' because index/labels/sizes are not equal along these coordinates (dimensions): 'name' ('name',) The recommendation is to set join explicitly for this case.

lkstrp: As discussed, with was fixed with the latest `linopy` release. You need to update your environment

pypsa/optimization/expressions.py
```diff
@@ -198,11 +197,26 @@ def capex(
         @pass_none_if_keyerror
         def func(n: Network, c: str, port: str) -> pd.Series | None:
             m = n.model
-            capacity = m.variables[f"{c}-{nominal_attrs[c]}"]
-            if include_non_extendable:
-                query = f"~{nominal_attrs[c]}_extendable"
-                capacity = capacity + n.c[c].static.query(query)["p_nom"]
-            costs = n.c[c].static[cost_attribute][capacity.indexes["name"]]
+            comp = n.c[c]
```
lkstrp: The internal convention we started is `c` for `pypsa.Component` and `component` as parameter name to pass any string or `pypsa.Component`. I think we can just also rename `c: str` here since this is internal. I will just adjust

pypsa/optimization/expressions.py
FabianHofmann: yes, this is what I had in mind as well ( but wanted to verify first).

pypsa/optimization/expressions.py
FabianHofmann: sure thing, I'll do that

PR #1318, opened 2025-08-10
Add SALib global sensitivity analysis example

fneum:

PR #1492, opened 2025-12-13
add water values example

fneum:


docs/examples/examples.md
```diff
@@ -61,6 +61,12 @@ The examples below demonstrate PyPSA's capabilities for energy system modeling.
 
     [:octicons-arrow-right-24: Go to example](rolling-horizon.ipynb)
 
+-   :material-notebook:{ .lg .middle } **Water Values**
+
+    Explores how water values, the marginal values of stored energy, can improve seasonal storage operation in rolling-horizon optimization.
```
brynpickering, line 66: Add an a.k.a. for `future cost function` here, since it is often used for this concept as well

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "pypsa.options.params.optimize.solver_name = \"gurobi\"\n",
+    "logging.getLogger(\"gurobipy\").setLevel(logging.ERROR)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ec3b7948",
+   "metadata": {},
+   "source": [
+    "For this example, we consider a simple energy system with a single bus and 3-hourly time steps over one year (8760 hours). We have time-varying load, wind and solar generation at zero marginal cost, a gas generator with marginal cost function $C'(g) = 80 + 0.01 g$ €/MWh, and hydrogen storage consisting of electrolysis, turbine, and a non-cyclic storage tank with a set initial energy. All components have fixed capacities; investments are not optimised here. "
```
brynpickering: 3-hourly means 2920 snapshots, so maybe say that in the brackets, instead of  `(8760)`, to avoid confusion

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,541 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "26742d60",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "n.buses_t.marginal_price[\"electricity\"].resample(\"D\").mean().plot(\n",
+    "    figsize=(6, 2), ylabel=\"€/MWh\", xlabel=\"\"\n",
+    ");"
```
brynpickering, line 147: unintended trailing semi-colon?

brynpickering: Or is there a value to including them in example notebooks?

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,541 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "44a14c8f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "n.model.solver_model = None\n",
+    "n2 = n.copy()\n",
+    "\n",
+    "n2.optimize.optimize_with_rolling_horizon(\n",
```
brynpickering, line 223: Can we set the log level to avoid having the per-horizon logging bloating the output?

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "    log_to_console=False,\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "14dfc7d5",
+   "metadata": {},
+   "source": [
+    "This leads to higher fuel costs of 1337 M€/a, as the storage is not optimally utilised for seasonal shifting and the gas generator must step in."
```
brynpickering: I wouldn't put the actual number into the markdown comment (same in an earlier comment) as it can easily be missed if something changes in the example network and updates are made to the examples

docs/examples/water-value.ipynb
fneum: intended, suppresses the return of the matplotlib axis and looks a bit cleaner

docs/examples/water-value.ipynb
fneum: right!

docs/examples/water-value.ipynb
fneum: could not quite eliminate everything, but reduced linopy logging.

lkstrp (changes requested):
Similar to #1318, the model created by the example is too large

>  GurobiError: Model too large for size-limited license; visit https://gurobi.com/unrestricted for more information 

We can either rewrite the example to get a smaller model, or add functionality to mark notebooks to not be run in the CI. But there is no guarantee that those stay up to date then

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,559 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e56dc0a4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "n.optimize(assign_all_duals=True, log_to_console=False)"
```
lisazeyen, line 108: I guess in your example one does not need to assign all duals since the marginal prices of hydrogen are saved in the network after the optimisation in any case. But maybe it is nice to keep it for the storage units
```suggestion
    "n.optimize(log_to_console=False)"
```

docs/examples/water-value.ipynb
```diff
@@ -0,0 +1,559 @@
...
+    "n.stores.e_cyclic = False\n",
+    "\n",
+    "n.add(\"Carrier\", \"gas\", color=\"darkorange\")\n",
+    "n.add(\n",
+    "    \"Generator\",\n",
+    "    \"gas\",\n",
+    "    bus=\"electricity\",\n",
+    "    p_nom=40_000,\n",
+    "    marginal_cost=80,\n",
+    "    marginal_cost_quadratic=0.01,\n",
```
lisazeyen, line 86: are the `marginal_cost_quadratic` necessary for the example? Other solvers like `highs` cannot handle it...

lisazeyen (commented):
Really nice example @fneum! We are currently implementing the rolling horizon in the open TYNDP, so that's helpful. Maybe it makes sense to rename the jupyter-notebook? Since it does not really show the implementation of hydro storage to something like `rolling-horizon-seasonal-storage`?

For storage units the right shadow prices from which one would assign the marginal costs is `n.storage_units_t.mu_energy_balance`, is this correct? If so I can add this information in a text box to your example.

docs/examples/water-value.ipynb
fneum: We'll switch to 'gurobi' in docs build (it's much easier with quadratic, otherwise need many generators to construct reasonable example)

docs/examples/water-value.ipynb
fneum: Yes, only required if you need the `mu_energy_balance` of the storage units (which I started with).

fneum:
Sorry, I didn't see your comments earlier @lisazeyen!

> Maybe it makes sense to rename the jupyter-notebook?

Hmm, I think the "water values" are a canonical concept in the literature that is now just transferred to long-duration storage. I would keep it.

> For storage units the right shadow prices from which one would assign the marginal costs is n.storage_units_t.mu_energy_balance, is this correct? 

Yes, exactly! But there's a complication for storage units in that their marginal cost is only assigned to dispatch (not charging). One has to split it up into two links and a store to model the charging bids as well.


The elastic demand is not required, just makes the example easier to construct.

PR #1495, opened 2025-12-16
Line losses with error control

### LINKED ISSUE #1320: Changing `s_nom_max` affects line losses

lindnemi:
Closes #1320.
Closes #1470.

Improved version of #1470.
Related to #1409.

- Make the approximation of the loss parabola independent of `s_nom_max`. To be able to do this a bound for the approximation error is needed, which has the added benefit of giving more fine-grained control of the approximation quality. Here an error bound based on absolute and relative error tolerances is used. This fixes #1320.

- Use secants instead of tangents for the approximation. That way, the approximation error is zero at the "corners" of the stepwise linear approximation. Since the optimum of the linear program tends to be located at a corner, this will hopefully reduce the effective error of the loss approximation

- Introduce lines losses close to 0. At the moment, flows smaller than `s_nom_max/(2*n_tangents)` are lossless.

## Example

Here is a plot of how the approximation with secants for an exemplary line looks like with `atol = 1 #MW` and `rtol = 0.1`. I chose the parameters such that the numbers of loss constraints is relatively small.

secants_example

The approximation could be improved by setting the endpoint of the last secant to `s_nom_max`. However, that would re-introduce a dependency on `s_nom_max` and should ideally only be done when it can be guaranteed that s_nom_max does not change between consecutive runs of the model. 

## Questions

- [ ] Are we happy with the overestimation of losses that may occur at `s_nom_max`? Of course a similar approximation with tangents could be constructed. However, that would underestimate the error at `s_nom_max`.
- [x] ~~Should we use equality constraints instead of inequality constraints? The plot in #1409 suggests that there are many snapshots when the losses are bigger than they should be, i.e. the model uses them to destroy energy.~~

## Derivation of formulas for relative and absolute error

Find a piecewise linear approximation $l(x)$ of the loss parabola $L(x) := rx^2$ on $[0, X]$ with:
1) $x_0 = 0$
2) $l(x_i) = L(x_i) = r x_i^2$, where $x_i$ are the endpoints of the linear segments

We want the endpoints to be on the parabola, since the optimal solution tends to be on the corners of a simplex, and hence that is where the approximation should be best.

(2) implies that our approximation consists of secants. Denote the endpoints of a secant as $a,b$ then

$l_{ab}(x) = \frac{L(b) - L(a)}{b - a }(x-a) + L(a) = r\frac{b^2 - a^2}{b - a }(x-a) + ra^2 = r(b+a)(x-a) + ra^2 = r((a+b)x - ab)$

The absolute error $AE(x)$ between $l_{ab}(x)$ and $L(x)$ is

$AE(x) = r((a+b)x - ab) - rx^2$  

This is positive for $a < x < b$, since there the secant is above the parabola. To compute it's maximum, remember that the derivative has to be 0.

$0 = \frac{d}{dx} AE(x) = r(a+b)x - 2rx \Rightarrow x_{max} = \frac{a+b}{2}$

Hence the error is maximal at the midpoint. Assume we specify the absolute error tolerance as $\varepsilon$. From this we can compute the interval length $b-a$

$\varepsilon = AE(\frac{a+b}{2}) = r((a+b)\frac{a+b}{2} - ab) - r(\frac{a+b}{2})^2 = r( \frac{(a+b)^2}{2} - \frac{(a+b)^2}{4} - ab) = r(\frac{(a+b)^2}{4} - \frac{4ab}{4}) = \frac{r(a-b)^2}{4}$ 

which implies $b(a, \varepsilon) = 2\sqrt{\frac{\varepsilon}{r}} + a$. 

Hence, to construct an approximation with absolute error at most $\varepsilon$, we choose the endpoints of the secants at $x_i = \dots, -4\sqrt{\frac{\varepsilon}{r}}, -2\sqrt{\frac{\varepsilon}{r}}, 0, 2\sqrt{\frac{\varepsilon}{r}}, 4\sqrt{\frac{\varepsilon}{r}}, \dots$

For small error tolerances and large line flows we might need many secants, which would translate to many additional constraint and slow down solving. So, in this case we will use an approximation based on the relative error instead.

The relative error is

$RE(x) = \frac{r((a+b)x - ab) - rx^2}{rx^2} = \frac{a+b}{x} - \frac{ab}{x^2} -1$

It's derivative is $\frac{2ab - (a+b)x}{x^3}$, which is zero at $x_{max}=\frac{2ab}{a+b}$.
Assuming the relative error tolerance is $\delta$ we get:

$\delta = RE(\frac{2ab}{a+b})= \dots = \frac{(b-a)^2}{4ab}$

We want to build a step rule $b(a)$ to construct a new end point $b$ from a given starting point $a$. Hence we rewrite the equation above as a polynomial in $b$ with constants $a$ and $\delta$:

$\delta = \frac{(b-a)^2}{4ab} \Leftrightarrow b^2 - a(2+4\delta)b + a^2 = 0$

This polynomial has the roots $b = a(1 + 2\delta \pm 2\sqrt{\delta(1+\delta)})$. It can be shown that the root with `-` term is smaller than $a$. Since we want $b>a$ we take the other root and get 

$b(a,\delta) = a(1 + 2\delta + 2\sqrt{\delta(1+\delta)})$

For small $a$ this gets very impractically small. Therefore we use the step based on absolute error for small $a$, and the step based on relative error for large $a$. Combined we have

$x_{i+1}(x_{i},\varepsilon,\delta) := b(a,\varepsilon,\delta) = \max(2\sqrt{\frac{\varepsilon}{r}}+ a, 2(\delta + \sqrt{\delta +\delta^2})a + a)$

We compute new endpoints until $x_I > X$. Then $l(x) = \sum_{i=0}^I l_{x_{i}x_{i+1}}(x)$.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ requires-python = ">=3.11"
 dependencies = [
     "numpy",
     "scipy",
-    "pandas>=0.24,<3.0.0",
+    "pandas>=2.0",
     "xarray",
     "netcdf4!=1.7.4",  # See https://github.com/PyPSA/PyPSA/issues/1510
     "linopy>=0.6.1",
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -47,6 +47,8 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - In version 2.0, capital costs of existing capacity on extendable assets will no longer be included in the objective by default (`n.objective_constant` will be set to zero), which improves LP numerical conditioning. A new `include_objective_constant` parameter was added to [`n.optimize()`][pypsa.optimization.OptimizationAccessor.__call__] and [`n.optimize.create_model()`][pypsa.optimization.OptimizationAccessor.create_model] to allow controlling this behavior and opt-in to the new default. It can also be configured via `pypsa.options.params.optimize.include_objective_constant` (see <!-- md:guide options.md -->). (<!-- md:pr 1509 -->)
 
+- Add support for pandas 3.0, while maintaining compatibility with pandas 2.x. (<!-- md:pr 1556 -->)
+
 ### Bug Fixes
 
 - Fix ramp limit constraints failing with mismatched index for multi-investment-period models with extendable or committable components. (<!-- md:pr 1537 -->)
--- a/pypsa/components/array.py
+++ b/pypsa/components/array.py
@@ -286,7 +286,7 @@ class ComponentsArrayMixin(_ComponentsABC):
 
             # Concatenate only if there is existing dynamic data
             if len(dynamic) > 0:
-                res = pd.concat([dynamic, static_to_dynamic], axis=1, copy=False)
+                res = pd.concat([dynamic, static_to_dynamic], axis=1)
                 res = res[index]
             else:
                 res = static_to_dynamic
--- a/pypsa/network/descriptors.py
+++ b/pypsa/network/descriptors.py
@@ -195,7 +195,7 @@ class NetworkDescriptorsMixin(_NetworkABC):
         def is_same_indices(i1: pd.Index, i2: pd.Index) -> bool:
             return len(i1) == len(i2) and (i1 == i2).all()
 
-        if is_same_indices(fixed_i.append(varying_i), index):
+        if is_same_indices(fixed_i.union(varying_i, sort=False), index):
 
             def reindex_maybe(s: pd.Series | pd.DataFrame) -> pd.Series | pd.DataFrame:
                 return s
--- a/pypsa/network/io.py
+++ b/pypsa/network/io.py
@@ -1089,7 +1089,7 @@ def _sort_attrs(
         return axis_labels
 
     remaining = axis_labels.difference(attrs_index, sort=False)
-    target = existing.append(remaining)
+    target = existing.union(remaining, sort=False)
 
     if axis_labels.equals(target):
         return axis_labels
--- a/pypsa/network/power_flow.py
+++ b/pypsa/network/power_flow.py
@@ -1316,7 +1316,7 @@ class SubNetworkPowerFlowMixin:
         self.pvs = buses_control.index[buses_control == "PV"]
         self.pqs = buses_control.index[buses_control == "PQ"]
 
-        self.pvpqs = self.pvs.append(self.pqs)
+        self.pvpqs = self.pvs.union(self.pqs, sort=False)
 
         # order buses
         self.buses_o = self.pvpqs.insert(0, self.slack_bus)
--- a/pypsa/statistics/expressions.py
+++ b/pypsa/statistics/expressions.py
@@ -398,7 +398,7 @@ class StatisticsAccessor(AbstractStatisticsAccessor):
         self, values: list[pd.DataFrame], agg: Callable | str
     ) -> pd.DataFrame:
         """Concatenate a list of DataFrames."""
-        df = pd.concat(values, copy=False) if len(values) > 1 else values[0]
+        df = pd.concat(values) if len(values) > 1 else values[0]
         if not df.index.is_unique:
             df = df.groupby(level=df.index.names).agg(agg)
         return df


fneum:
> Should we use equality constraints instead of inequality constraints? The plot in https://github.com/PyPSA/PyPSA/pull/1409 suggests that there are many snapshots when the losses are bigger than they should be, i.e. the model uses them to destroy energy.

Equality constraints would require binary variables to identify the segments wouldn't it?

pypsa/optimization/constraints.py
```diff
@@ -1688,25 +1694,79 @@ def define_loss_constraints(
...
-                lhs >= offset_k, name=f"{c.name}-loss_tangents-{k}-{sign}", mask=active
+                lhs >= offset,
+                name=f"{c.name}-loss_secants-{s}",
+                mask=active,
             )
+    else:
+        # Add upper limit constraint
+        n.model.add_constraints(
+            loss <= upper_limit, name=f"{c.name}-loss_upper", mask=active
+        )
```
fneum: I think this one should apply to the "secant" mode as well. It just bounds the losses (in case of negative prices). Can just remove out of if-else block.

pypsa/optimization/constraints.py
```diff
@@ -1688,25 +1694,79 @@ def define_loss_constraints(
     loss = n.model[f"{c.name}-loss"]
     flow = n.model[f"{c.name}-s"]
 
-    # Add upper limit constraint
-    n.model.add_constraints(
-        loss <= upper_limit, name=f"{c.name}-loss_upper", mask=active
-    )
+    if mode == "secants":
+        atol = 1
+        rtol = 0.1
```
fneum: Should become function arguments (and can than be described in more detail in the docstring).

pypsa/optimization/constraints.py
```diff
@@ -1688,25 +1694,79 @@ def define_loss_constraints(
...
+        target = (s_nom_max * s_max_pu).where(lossy, 0)
+
+        factor_rtol = 2 * (rtol + sqrt(rtol + rtol**2))  # + 1
+        step_atol = where(lossy, 2 * sqrt(atol / r_pu_eff), 0)
+
+        # start at 0 everywhere
+        p_k = zeros_like(r_pu_eff)
+        p_k.name = "p_k"
+        breakpoints = [p_k.copy()]
+        max_iter = 50
```
fneum: Should become function argument.

pypsa/optimization/constraints.py
```diff
@@ -1688,25 +1694,79 @@ def define_loss_constraints(
...
-            lhs = n.model.linexpr((1, loss), (sign * slope_k, flow))
+        for sign, s in [(-1, "neg"), (1, "pos")]:
+            lhs = n.model.linexpr((1, loss), (sign * slope, flow))
             n.model.add_constraints(
-                lhs >= offset_k, name=f"{c.name}-loss_tangents-{k}-{sign}", mask=active
+                lhs >= offset,
+                name=f"{c.name}-loss_secants-{s}",
+                mask=active,
             )
+    else:
```
fneum: Should probably check explicitly for `mode == "tangents"`.

pypsa/optimization/constraints.py
```diff
@@ -1616,7 +1616,11 @@ def define_store_constraints(n: Network, sns: pd.Index) -> None:
 
 
 def define_loss_constraints(
-    n: Network, sns: pd.Index, component: str, transmission_losses: int
+    n: Network,
+    sns: pd.Index,
+    component: str,
+    transmission_losses: int,
+    mode: str = "secants",
```
fneum:
```suggestion
    transmission_losses: int, # deprecate, use max_segments instead
    atol: float = 1,
    rtol: float = 0.1,
    max_segments = 30,
    mode: str = "secants",
```

I would give a few more new keyword arguments here. The argument `transmission_losses` was never well designed and I would suggest to deprecate it here.

What is still needed is that the arguments can be passed through `n.optimize()` and `n.optimize.create_model()` (everywhere where `transmission_losses` shows up and `define_loss_constraints` is called).

Maybe here, the arguments could be bundled in a dictionary for `transmission_losses`?

```py
n.optimize(
    transmission_losses: int | dict = 0
)

transmission_losses = dict(
    max_segments = 10,
    atol = 10,
    rtol = 0.2,
    mode = "secants",
)

transmission_losses = dict(
    max_segments = 3,
    mode = "tangents",
)
```

Maybe @lkstrp can help with making the transition as smooth as possible and has an idea for a good keyword argument design.

The default should probably stay as `tangents` for now, and change with a future version.

lindnemi:
Thank you @fneum for the quick review and the valuable suggestions.

I implemented the changes and refactored the code for computing the secants one last time. By algebraically rearranging the computation, it was possible to improve the numerical stability, s.t. the offsets should all be the same now (before that there were some very small rounding errors, which should not impact results, but potentially solver speed??)

I also sketched a potential interface for tranmissions losses based on dicts, and a few unit tests. @lkstrp maybe you can take over from here? Of course i am very happy to support if any questions arise.

lindnemi:
btw, i noticed that for passive branch components like `Transformers` the loss variables get defined, but there are no constraints on them, s.t. they end up NA. This becomes apparent when solvend the `ac_dc_meshed` example with losses and inspecting the Line loss variables -  a lot of nans turn up:

image
@fneum

In the screenshot there is also a UserWarning visbile: `Coordinates across variables not equal. Perform outer join.
`

@lkstrp do you if this is because i introduced the new `secant` coordinate?

lkstrp:
> @lkstrp do you if this is because i introduced the new secant coordinate?

What you are showing there is just the solution, which is a single xarray dataset and shares a single `name` coord, which has all component names. So those `nan`s are probably the generators of `ac_dc_meshed`? It shouldn't have a Transformer

And the user warning should be gone if you update to `linopy>=0.5.8`.


I need to properly understand this and have a detailed look. But sure all the API stuff, tests etc I can take over. What about docs? Can you write up something about the additional formulation? https://docs.pypsa.org/latest/user-guide/optimization/power-flow/?h=line+loss#loss-approximation

lindnemi:
>  So those nans are probably the generators of ac_dc_meshed? 

Yes, that was a misunderstanding on my side.

> And the user warning should be gone if you update to linopy>=0.5.8.

I am actually using linopy 0.5.8.

> What about docs? Can you write up something about the additional formulation?

I didn't know we had docs on line losses, but that's great, I will write something up.

> But sure all the API stuff, tests etc I can take over.

Great, thanks! The new secant formulation should become the default with the next PyPSA version. So maybe we need some deprecation warnings?

pypsa/optimization/constraints.py
```diff
@@ -1700,10 +1636,170 @@ def define_loss_constraints(
...
+
+    # Call the intersection points of a secant with the loss curve a and b, then we have:
+    a_factors = breakpoint_factors.isel(secant=slice(None, -1))  # k segments: 0..K-1
+    b_factors = breakpoint_factors.isel(secant=slice(1, None))  # k segments: 1..K
+    b_factors["secant"] = b_factors["secant"] - 1  # align indices
+
+    # The final breakpoint can be made smaller by setting it to the target
+    # This improves the approximation, however may introduce a dependency on s_nom_max:
+    # Therefore we do it only for lines that are non-extendable:
+    b_factors = where(
```
lindnemi: This line of code introduces a slight dependency of the line losses on the variable `s_nom_extendable`.

If `s_nom_extendable=False` for a given line, the losses for large flows on that line can be made  more accurate and hence will be slightly lower.  This may imply that between two identical lines, one with `s_nom_extendable=False` and the other one with True, the non-extendable line is prefarable (for large flows).

It also means that switching the extendability of a line may affect the flows on that line, even if everything else stay equal. 

I think in most cases it will be preferable to be slightly wrong but consistent accross lines and extendability. Hence, i tend to remove L 1778-1780.

However, removing it means, that we will always overestimate the losses of the maximum possible flow on that line and that we sacrificy a bit of accuracy. Any thoughts on that, @fneum ?

fneum: Ok, I'm fine with that change and convinced by the consistency argument.

docs/user-guide/optimization/power-flow.md
```diff
@@ -103,36 +90,82 @@ splitting losses equally between both connection points.
...
-enabled by providing a number of tangents in [`n.optimize()`][pypsa.optimization.OptimizationAccessor.__call__].
+Two linearization methods are available in PyPSA: the tangent-based approximation ([`define_tangent_loss_constraints()`][pypsa.optimization.constraints.define_tangent_loss_constraints]) and the secant-based approximation with error control ([`define_secant_loss_constraints()`][pypsa.optimization.constraints.define_secant_loss_constraints]).
+
+The transmission loss approximation is not activated by default, but must be enabled in [`n.optimize()`][pypsa.optimization.OptimizationAccessor.__call__].
 
-``` py
-n.optimize(transmission_losses=3)
+```py
+# Secant-based approximation with error control
+n.optimize(transmission_losses=True)
```
fneum, line 100: Just to confirm that `n.optimize(transmission_losses=3)` is still possible to be non-breaking? Happy if it's deprecated.

docs/user-guide/optimization/power-flow.md
```diff
@@ -103,36 +90,82 @@ splitting losses equally between both connection points.
...
+m_{k, l} = r_l (\rho_{k-1} + \rho_{k})
+$$
+
+$$
+a_{k, l} = - r_l \rho_{k-1} \rho_{k}
+$$
+
+where $\rho_k$ are suitably chosen points on the loss parabola. For details on how these points are derived see the [original PR](https://github.com/PyPSA/PyPSA/pull/1495).
+
+### Legacy Tangent-Based Linearization
```
fneum: Remove "Legacy" since we for now intend to keep it.

docs/user-guide/optimization/power-flow.md
lindnemi: @lkstrp

lkstrp: Yes, deprecated and announced as FutureWarning for `2.0` but until then works as right now. Default is now `False` instead of `0`, which does the same and passing a int will then not work anymore

pypsa/optimization/optimize.py
```diff
@@ -434,11 +431,19 @@ def __call__(
...
             investment periods. Then, snapshots should be a `pd.MultiIndex`.
-        transmission_losses : int, default 0
-            Whether an approximation of transmission losses should be included
-            in the linearised power flow formulation. A passed number will denote
-            the number of tangents used for the piecewise linear approximation.
-            Defaults to 0, which ignores losses.
+        transmission_losses : bool | int | dict, default False
+            Include piecewise linear approximation of transmission losses for
+            passive branches:
+
```
lkstrp, line 437:
```suggestion
```

pypsa/optimization/optimize.py
```diff
@@ -434,11 +431,19 @@ def __call__(
...
+            - ``True``: secant-based approximation with default tolerances
+            - ``int``: *(deprecated)* tangent-based with that many segments
+            - ``dict``: explicit config with key ``mode`` ("secants" or
+              "tangents") and mode-specific options. Secant options:
+              ``atol`` (default 1), ``rtol`` (default 0.1),
+              ``max_segments`` (default 20). Tangent options: ``segments``
+              (required).
+
```
lkstrp, line 445:
```suggestion
```

pypsa/optimization/optimize.py
```diff
@@ -564,9 +569,18 @@ def create_model(
...
+            Include piecewise linear approximation of transmission losses for
+            passive branches:
+            - ``True``: secant-based approximation with default tolerances
+            - ``int``: *(deprecated)* tangent-based with that many segments
+            - ``dict``: explicit config with key ``mode`` ("secants" or
+              "tangents") and mode-specific options. Secant options:
+              ``atol`` (default 1), ``rtol`` (default 0.1),
+              ``max_segments`` (default 20). Tangent options: ``segments``
+              (required).
+
```
lkstrp, line 582:
```suggestion
```

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "pandas>=0.24,<3.0.0",
     "xarray",
     "netcdf4!=1.7.4",  # See https://github.com/PyPSA/PyPSA/issues/1510
-    "linopy>=0.6.1",
+    "linopy==0.6.1",
```
lkstrp:
```suggestion
    "linopy>=0.6.1",
```

PR #1506, opened 2026-01-02
docs: add security policy

lkstrp:
With the email setup, we can also have a proper security policy now. Good to have as a private contact channel instead of going via public issues/ a random maintainers email


fneum:
Should be integrated somewhere in docs, right?

PR #822, opened 2024-01-26
Fix of capex description

Alex-Neve:
Closes PyPSA Issue #817.

## Changes proposed in this Pull Request
Correction of the capex descriptions of the attributes CSV files.

## Checklist

- [ N/A ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ N/A ] Unit tests for new features were added (if applicable).
- [ N/A ] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [ x ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [ x ] I consent to the release of this PR's code under the MIT license.

Issue #817, opened 2024-01-22
CAPEX documentation issue


Alex-Neve:
## Checklist

- [ X] I am using the current [`master`](https://github.com/pypsa/pypsa/tree/master) branch or the latest [release](https://github.com/pypsa/pypsa/releases). Please indicate:

## Describe the Bug

* The CSV files of generators, lines, links, storage_units, stores and transformers stored under the "_component_attrs_" folder incorrectly state in the "_description_" column that the "_capital_cost_" attribute is the "_Capital cost of extending p_nom by 1 MW_."
* However, as pointed out in the PyPSA article "[Power System Optimization](https://pypsa.readthedocs.io/en/latest/optimal_power_flow.html)", "_the capital_cost of the assets is now the fixed annual costs, including annuity and FOM._"
* I suggest replacing the "_capital_cost_" attribute description with "_Fixed annual costs, including investment costs annuity and Fixed O&M costs._" (This is done in the attached CSV files) 
[transformers.csv](https://github.com/PyPSA/PyPSA/files/14015212/transformers.csv)
[generators.csv](https://github.com/PyPSA/PyPSA/files/14015213/generators.csv)
[lines.csv](https://github.com/PyPSA/PyPSA/files/14015214/lines.csv)
[links.csv](https://github.com/PyPSA/PyPSA/files/14015216/links.csv)
[storage_units.csv](https://github.com/PyPSA/PyPSA/files/14015217/storage_units.csv)
[stores.csv](https://github.com/PyPSA/PyPSA/files/14015218/stores.csv)

FabianHofmann:
Hey @Alex-Neve, thank you for your suggestion. I am not sure whether the proposed description is general enough. Assume you have a time horizon of less or more than one year. Then the term `annualized costs` is not correct any more, right?

Alex-Neve:
Hi @FabianHofmann ,

Would you prefer "periodized costs"? This would leave it to the user to pick his/her preferred time-scale.

In my sector of the industry we use annualized costs because the investment periods are so long that monthly discounting of the costs would not make sense, but other users may work at different time-scales.

If you prefer "periodized costs", I might leave an indication in the CSV files that if periods are defined as years, these periodized costs are identical to annualized costs.

Kr,
Alex

Alex-Neve:
Here are the CSV files that I have reworded to avoid the terms "annual" and "annualized".
FYI, I cannot create a branch to propose this change myself in a pull request.

[...]

PR #1507, opened 2026-01-03
feat: split capital_cost into investment and FOM costs

### LINKED ISSUE #371: Split capital cost into FOM and investment costs

FabianHofmann:
Closes #371.

Add support for separating `capital_cost` into overnight investment costs and fixed O&M:

**EDIT (updated summary)**

Add support for separating overnight investment costs from fixed O&M costs:

  - New component attributes:
    - overnight_cost - Overnight (upfront) investment cost per MW (default NaN)
    - discount_rate - Discount rate for annuity calculation (default NaN)
    - fom_cost - Fixed operation and maintenance costs per year (default 0)
  - New pypsa.costs module with:
    - annuity(discount_rate, lifetime) - Calculate annuity factor
    - annuity_factor(discount_rate, lifetime) - Alias for annuity
    - periodized_cost(overnight_cost, discount_rate, lifetime, fom_cost, nyears) - Full periodized cost calculation
  - New statistics methods:
    - n.statistics.overnight_cost() - Report overnight investment costs
    - n.statistics.fom() - Report fixed O&M costs
  - Backward compatible: When overnight_cost is not set (NaN, default), capital_cost is used directly as annualized cost, preserving existing behavior.

## Checklist

### DIFF
--- a/docs/examples/capacity-expansion-planning-single-node.ipynb
+++ b/docs/examples/capacity-expansion-planning-single-node.ipynb
@@ -119,7 +119,7 @@
    },
    "outputs": [],
    "source": [
-    "from pypsa.common import annuity"
+    "from pypsa.costs import annuity"
    ]
   },
   {
--- a/docs/examples/example-3.ipynb
+++ b/docs/examples/example-3.ipynb
@@ -65,7 +65,7 @@
     "import pandas as pd\n",
     "\n",
     "import pypsa\n",
-    "from pypsa.common import annuity\n",
+    "from pypsa.costs import annuity\n",
     "\n",
     "n = pypsa.Network()\n",
     "\n",
--- a/docs/examples/islanded-methanol-production.ipynb
+++ b/docs/examples/islanded-methanol-production.ipynb
@@ -50,7 +50,7 @@
     "import pandas as pd\n",
     "\n",
     "import pypsa\n",
-    "from pypsa.common import annuity"
+    "from pypsa.costs import annuity"
    ]
   },
   {
--- a/docs/examples/modular-expansion.ipynb
+++ b/docs/examples/modular-expansion.ipynb
@@ -38,7 +38,7 @@
    "outputs": [],
    "source": [
     "import pypsa\n",
-    "from pypsa.common import annuity\n",
+    "from pypsa.costs import annuity\n",
     "\n",
     "n = pypsa.examples.model_energy()\n",
     "\n",
--- a/docs/examples/myopic-pathway.ipynb
+++ b/docs/examples/myopic-pathway.ipynb
@@ -39,7 +39,7 @@
     "import pandas as pd\n",
     "\n",
     "import pypsa\n",
-    "from pypsa.common import annuity\n",
+    "from pypsa.costs import annuity\n",
     "\n",
     "temp = pypsa.examples.model_energy()\n",
     "temp.remove(\"Generator\", \"load shedding\")\n",
--- a/docs/examples/stochastic-optimization.ipynb
+++ b/docs/examples/stochastic-optimization.ipynb
@@ -55,7 +55,7 @@
     "import pandas as pd\n",
     "\n",
     "import pypsa\n",
-    "from pypsa.common import annuity"
+    "from pypsa.costs import annuity"
    ]
   },
   {


lindnemi:
Very elegant solution given the current `capital_cost` attribute, and covers what i had in mind. However...

> However, I would prefer instead to have investment costs and capital cost side by side and only allow one of them to be defined by the user. 

I totally agree.

pypsa/components/array.py
```diff
@@ -349,3 +350,28 @@ def _as_xarray(self, attr: str) -> xarray.DataArray:
             _assert_xarray_integrity(self, res)
 
         return res
+
+    def get_effective_annual_cost(self, idx: pd.Index) -> xr.DataArray:
```
fneum:
```suggestion
    def periodized_cost(self, idx: pd.Index) -> xr.DataArray:
```
shorter, also consistent with #817 and #822

pypsa/data/component_attrs/generators.csv
```diff
@@ -22,7 +22,9 @@ marginal_cost_quadratic,static or series,currency/MWh,0.,Quadratic marginal cost
 active,boolean,n/a,True,"Whether to consider the component in optimization or not",Input (optional)
 build_year,int,year,0,Build year of the generator.,Input (optional)
 lifetime,float,years,inf,Lifetime of the generator.,Input (optional)
-capital_cost,float,currency/MW,0,"Fixed period costs of extending `p_nom` by 1 MW, including periodized investment costs and periodic fixed O&M costs (e.g. annuitized investment costs).",Input (optional)
+capital_cost,float,currency/MW,0,"Fixed period costs of extending `p_nom` by 1 MW. If `discount_rate=0`, this is the annuitized cost (current behavior). If `discount_rate>0`, this is the overnight investment cost and PyPSA calculates the annuity using `lifetime`.",Input (optional)
```
fneum: Unfortunately, the distinction by `discount_rate` (zero or non-zero) whether `capital_costs` denotes annualized or overnight investment costs does not work because `discount_rate=0` is a valid input for the annuity function.

If I gave as input overnight cost of 100, lifetime 20, discount rate 0, I would expect annual cost to be 5. The current implementation would assume 100 as already annuitized cost.

(The handling of this case was accidentally removed at some point https://github.com/PyPSA/PyPSA/pull/1512)

pypsa/optimization/optimize.py
```diff
@@ -136,8 +136,8 @@ def define_objective(n: Network, sns: pd.Index) -> None:
         if ext_i.empty:
             continue
 
-        capital_cost = c.da.capital_cost.sel(name=ext_i)
-        if capital_cost.size == 0:
+        annual_cost = c.get_effective_annual_cost(ext_i)
```
fneum: Try to minimize use of "annual"; prefer "periodic" as per discussion in #817 and #822 (more general to also cover networks with multiple years).

Example: A network expanding over two years should have as `capital_cost` two times the annuity.

pypsa/costs.py
```diff
@@ -0,0 +1,180 @@
...
+    --------
+    >>> import pypsa
+    >>> pypsa.costs.annuity(0.07, 25)  # doctest: +ELLIPSIS
+    0.0858...
+
+    """
+    # Input validation for scalars
+    if isinstance(discount_rate, int | float):
+        if discount_rate < 0:
+            msg = f"discount_rate must be non-negative, got {discount_rate}"
```
fneum: In principle, real discount rates can also be negative! That case should not be excluded (penalizes the present).

pypsa/costs.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError(msg)
+        if discount_rate == 0:
+            msg = "discount_rate=0 is undefined for annuity. Use annuity_factor() instead."
```
fneum: Discount rate of zero **is** defined for annuity function.

pypsa/costs.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError(msg)
+    if isinstance(lifetime, int | float) and lifetime <= 0:
+        msg = f"lifetime must be positive, got {lifetime}"
+        raise ValueError(msg)
+
+    # Handle infinite lifetime: limit of annuity as n -> inf is r
+    if isinstance(lifetime, int | float) and np.isinf(lifetime):
+        return discount_rate
```
fneum: Nice edge case that is not covered by the current annuity function! The limit is only $r$ if $r>0$, if $r<0$ I think the limit is 0.

pypsa/costs.py
```diff
@@ -0,0 +1,180 @@
...
+    elif isinstance(lifetime, xr.DataArray):
+        result = xr.where(np.isinf(lifetime), discount_rate, result)
+
+    return result
+
+
+def annuity_factor(
+    discount_rate: T,
+    lifetime: T,
+) -> T:
```
fneum: As soon as discount rate of zero is allowed, this extra function is not necessary. I would avoid this extra function; it's a bit confusing.

fneum (changes requested):
Thanks @FabianHofmann for tackling this. It's good that the issue of separating FOM from `capital_cost` and a way to compute both periodized **and** overnight investment cost is moving forward! 

First, I'll give the reasons why I think this solution unfortunately does not work (sorry!), then I'll think of a solution in a separate comment.

1. A discount rate of zero is a valid input for annualising investment costs and should therefore not be used as a switch. (Example: overnight cost=100, discount rate=0, lifetime=20 give periodized cost of 100/20=5). That's the strongest reason.

2. I don't think overloading `capital_cost` is good style. The meaning of values in `capital_cost` thereby depend on another attribute and are not directly clear. If you want to do anything with the `capital_cost` attribute, you will always have to make a case distinction, which is inconvenient. That's a good reason, but less strong than 1.

I would prefer a solution where `capital_cost` still denotes the annuitized investment costs, and build a solution around that. I like how the FOM are separated from `capital_cost` and the additional `discount_rate` attribute.

pypsa/statistics/expressions.py
```diff
@@ -913,6 +914,182 @@ def expanded_capex(  # noqa: D417
...
 
+    @MethodHandlerWrapper(handler_class=StatisticHandler, inject_attrs={"n": "_n"})
+    @deprecated_kwargs(
+        deprecated_in="1.0",
+        removed_in="2.0",
+        comps="components",
+        aggregate_groups="groupby_method",
+        aggregate_time="groupby_time",
+    )
+    def investment(  # noqa: D417
```
fneum: Isn't the CAPEX function wrong now, mixing annuitized and overnight costs (because it simply uses `capital_cost` attribute)? That's a high risk.

pypsa/costs.py
```diff
@@ -0,0 +1,180 @@
...
+        Annual cost = capital_cost * annuity_factor(discount_rate, lifetime) + fom_cost.
+
+    Examples
+    --------
+    >>> import pypsa
+    >>> pypsa.costs.effective_annual_cost(1000, 0.07, 25, 20)  # doctest: +ELLIPSIS
+    105.8...
+
+    """
+    annuitized = capital_cost * annuity_factor(discount_rate, lifetime)
```
fneum: This only works for the special case of a single year of snapshots. A more general function would multiply overnight investment cost with annuity factor times number of full years.

pypsa/data/component_attrs/generators.csv
```diff
@@ -22,7 +22,9 @@ marginal_cost_quadratic,static or series,currency/MWh,0.,Quadratic marginal cost
...
+discount_rate,float,per unit,0,"Discount rate for annuity calculation. If >0, `capital_cost` is treated as overnight investment cost.",Input (optional)
+fom_cost,float,currency/MW/a,0,"Fixed operation and maintenance costs per year, added to annuitized investment cost.",Input (optional)
```
fneum:
```suggestion
fom_cost,float,currency/MW,0,"Periodic fixed operation and maintenance costs, added to annuitized investment cost.",Input (optional)
```
Same unit as `capital_cost`. These are per period (i.e. all snapshots), not necessarily one year. Consider the cases where more or less then one year is in network.

fneum:
The way I see it, this PR tries to achieve three things:

1. A way to disentangle FOM from `capital_cost` attribute.
2. A way to access both annualized **and** overnight investment costs.
3. More convenience for calculating annualized investment costs (common task) from known
    - overnight cost,
    - lifetime,
    - discount rate, and
    - number of years in snapshots.

(1.) and (2.) are relatively straightforward to implement with backwards compatibility and minimal invasiveness to the current use of the `capital_cost` attribute; (3.) is a challenge but also an add-on and could be left for later. 

A solution for (1.) and (2.)  could look like this (recycled and adapted from https://github.com/PyPSA/PyPSA/issues/371#issuecomment-3051888434):

1. Split `capital_cost` into `fom_cost` and `capital_cost`. The **new optional input attribute** `fom_cost` would default to 0. In the objective, both attributes are simply summed up and associated with the investment variable. This is compatible with old networks where FOM is included in `capital_cost`.
2. Add a **new optional input attribute** `discount_rate` (defaults to 0). Together with the **existing optional input attribute** `lifetime`, the annuity factor can be calculated with `pypsa.common.annuity`.
3. Add a **new output attribute** `overnight_cost` that is a dependent attribute (similar to `x_pu_eff`) and a function of `capital_cost`, `lifetime`, and `discount_rate` (excluding `fom_cost`. Use `annuity_factor()` function to calculate it in `n.calculate_dependent_values()`. The status quo remains that the annualized investment costs are entered by the user, the overnight cost are derived.
4. Add a new statistics function for calculating overnight costs (upfront investments) based on `overnight_cost`, called `n.statistics.overnight_cost()`.
5. The statistics function `n.statistics.total_cost()` must be amended by `fom_cost` since they are no longer included in `n.statistics.capex()`.

Overall, I think this offers a good compromise between backward compatibility, invasiveness to the current use of `capital_cost`, and the ability to distinguish investment costs (as annualized **and** overnight value) and FOM. 

The suggestion does not offer more convenience for calculating annualized `capital_costs` (3.), but that's a fairly simple calculation with the utility function `pypsa.common.annuity` at hand. The dilemma here is the clear distinction of inputs and outputs in PyPSA by design. Ideally, we'd want the user to specify either annualized **or** overnight costs, and the respective other attribute is filled based on derived values. But that option requires handling if both attributes are filled with values. Maybe a simple consistency check and routine in calculating dependent values will already do the trick?

lindnemi:
> The dilemma here is the clear distinction of inputs and outputs in PyPSA by design. Ideally, we'd want the user to specify either annualized or overnight costs, and the respective other attribute is filled based on derived values. 

Isn't that very simple to achieve with a wrapper, that throws an error when both are there?

Basically `n.add` would start by checking its arguments for consistency, and derive either capital_cost or overnight_cost, and explain what's wrong otherwise. Or maybe I am missing something?

fneum:
> Basically n.add would start by checking its arguments for consistency, and derive either capital_cost or overnight_cost, and explain what's wrong otherwise. Or maybe I am missing something?

Doesn't sound too bad, actually. Just needs to be extended to all other cases of IO and network building / manipulation with a strong consistency check.

lkstrp:
> Basically n.add would start by checking its arguments for consistency, and derive either capital_cost or overnight_cost, and explain what's wrong otherwise. Or maybe I am missing something?

That just starts mixing input and output variables, which we don't currently do by design. But I am not opposed to the idea in general. There is just more to it. In `n.add`, this can also be easily checked with good user feedback. But `n.add` is not our only interface, we also need to check any dataframe manipulation and IO. And then the UX get's more tricky.

> Doesn't sound too bad, actually. Just needs to be extended to all other cases of IO and network building / manipulation with a strong consistency check.

We can't extend the same check, since in `n.add` I only pass one of both. Via IO or in the dataframe, I have both values calculated already, without knowing which one was passed. The consistency check would basically need to recalculate all values to check if they still match, and not just check "are both given?". And if they don't match you say "add one of them again, to overwrite the other"? If you changed `discount_rate`/ `life_time`, do you take capital or overnight cost to calculate the other? Also `n.add` would only solve this with `overwrite=True`, which is not default. It is also currently not deriving any attributes. Compute wise that would probably work, but the additional compute as well as UX isn't very clean. 

I am mainly wondering how many other use cases we would have, where an attribute could be both output and input. If there are not many, I would rather keep the existing input vs output structure, but assign them dynamically which we could do here with a hidden helper attribute, which stores the info which one was the input. And keep to (or start to properly) enforce only input can be changed, which would be best for any interface (`n.add`, DataFrame, Excel, App etc.). But as @FabianHofmann said, it's not much fun to implement without proper validation. I should probably just start resurrecting #1128

fneum:
> That just starts mixing input and output variables, which we don't currently do by design. 

Yes, ideally that does not change.

> The consistency check would basically need to recalculate all values to check if they still match, and not just check "are both given?"

Yes, that's what I thought. Not sure if that's elegant. Probably not.

> And if they don't match you say "add one of them again, to overwrite the other"?

Hmm, difficult. The user has to fix it by aligning either `capital_cost` or `overnight_cost` with the rest? Not great.

lindnemi:
Thanks Lukas for elaborating a bit on the complexity of the I/O. Regarding the consistency_check:

> > And if they don't match you say "add one of them again, to overwrite the other"?
> 
> Hmm, difficult. The user has to fix it by aligning either `capital_cost` or `overnight_cost` with the rest? Not great.

I would still suggest to do exactly that and I don't think it's too bad. If the user specifies both values for some odd reason, that's most likely due to a mistake in their model building pipeline. So in the best case throwing a hard error in the consistency check would only improve their model by forcing the elimination of ambiguities.

FabianHofmann:
thanks all for the comments. I am wondering if there is good way to make the stretch between @lindnemi requirements for convenience which are valid and touch quite some use cases and @fneum's request to keep a clear data model and avoid ambiguity which I also strongly second, why I came up with the overloading approach. 

**Proposal 1:**

A boolean attribute `capital_as_overnight_cost` (default false) decides the kind of cost input and how to process. (It's essentially what is now hidden in the discount rate which is bad API I have to admit).

Like this we:
* allow user to set all costs with 
   * overnight cost, lifetime, discount rate → new use case
   * capital cost (old use case)
* don't change inputs on the fly and set ambiguous attributes potentially seen as inputs
* can derive all important cost parameters internally and expose them via component attributes:
   * `n.c.Generator.overnight_cost`
   * `n.c.Generator.capital_cost`
   * `n.c.Generator.total_cost`
   * `n.c.Generator.annuity`
* new component attributes feed into statistics functions

Note: FOM is independent and simply adds to period capital costs in optimization.

```python
# Old behavior (default)
n.add("Generator", "gas", capital_cost=105.8)

# New behavior with overnight costs
n.add("Generator", "wind",
    capital_cost=1000,  # interpreted as overnight
    capital_as_overnight_cost=True,
    discount_rate=0.07,
    lifetime=25,
    fom_cost=20,
)
# → n.c.generators.capital_cost["wind"] ≈ 85.8 (derived)
# → n.c.generators.total_cost["wind"] ≈ 105.8
```

---

**Proposal 2:**

Introduce an `overnight_cost` attribute (default NaN) which takes precedence over `capital_cost` when set. Don't calculate one from the other—full information is exposed through component attributes (`n.c.Generator....`). Raise `consistency_check` warning if both are given, skip checks in `n.add` and other IO until we have proper data validation.

```python
# Old behavior
n.add("Generator", "gas", capital_cost=105.8)

# New behavior
n.add("Generator", "wind",
    overnight_cost=1000,
    discount_rate=0.07,  # 0% is valid here!
    lifetime=25,
    fom_cost=20,
)
# → optimization uses: 1000 * annuity(0.07, 25) + 20 ≈ 105.8
```

---

Proposal 2 might be cleaner—no boolean flag, separate attributes with clear semantics, and `discount_rate=0` works correctly. Proposal 1 reuses `capital_cost` for two meanings, same flexibility, mitigates confusion on the user side.

fneum:
Out of proposals 1 and 2, proposal 2 is cleaner.

@lindnemi needs to chime in, since he already uses `overnight_cost` in PyPSA-DE networks where both are stored.

The attribute `total_cost` should be called `periodic_cost`, since (e.g. in statistics), total cost includes also OPEX.

Not sure I understood the `n.c.Generator.capital_cost`. Wouldn't it be an issue if `n.generators.capital_cost` would not return simply the values of the column in `n.generators` but a (potentially) computed value? That I would find confusing.

lindnemi:
> @lindnemi needs to chime in, since he already uses `overnight_cost` in PyPSA-DE networks where both are stored.

For me both is fine. I like Proposal 2 better, even though it will mean a bit more refactoring work on the PyPSA-DE side because there every asset already has `capital_cost` as well as `overnight_cost` defined, which will then lead to warnings.

What's most important for me, is that capital_cost, overnight_cost, FOM and lifetime are separately accesible in the solved networks  (and ideally via n.statistics)

coroa:
I like proposal 2, too.

> Not sure I understood the `n.c.Generator.capital_cost`. Wouldn't it be an issue if `n.generators.capital_cost` would not return simply the values of the column in `n.generators` but a (potentially) computed value? That I would find confusing

`n.c.Generator.capital_cost` (does not exist yet and can be chosen) is not the same level as `n.c.Generator.static.capital_cost` (column in df)

docs/api/other/costs.md
```diff
@@ -0,0 +1,12 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+::: pypsa.costs
```
lkstrp, line 7: Needs to be added to `mkdocs.yml`

pypsa/components/array.py
```diff
@@ -347,3 +348,109 @@ def _as_xarray(self, attr: str) -> xarray.DataArray:
             _assert_xarray_integrity(self, res)
 
         return res
+
+    @property
+    def periodized_cost(self) -> xr.DataArray:
```
lkstrp: Why should we put any of this under `c.da`? Currently `c.da` is only transforming df data, not giving you any custom properties

pypsa/common.py
```diff
@@ -928,9 +928,17 @@ def generate_colors(n_colors: int, palette: str = "tab10") -> list[str]:
     return colors
 
 
+@deprecated(
+    deprecated_in="0.35.0",
```
lkstrp: Claude

pypsa/consistency.py
```diff
@@ -825,6 +878,8 @@ def consistency_check(
             check_assets(self, c, "assets" in strict)
             # Checks generators
             check_generators(c, "generators" in strict)
+            # Checks cost attributes consistency
+            check_cost_consistency(c, "cost_consistency" in strict)
```
lkstrp: Why not make it strict by default

pypsa/costs.py
```diff
@@ -0,0 +1,245 @@
...
+
+from __future__ import annotations
+
+from typing import TypeVar
+
+import numpy as np
+import pandas as pd
+
+# Type variable for numeric types that support array operations
+T = TypeVar("T", float, pd.Series)
```
lkstrp: I don't like this for something that simple

pypsa/costs.py
```diff
@@ -0,0 +1,245 @@
...
+        result = result.where(~np.isinf(lifetime), inf_val)
+
+    return result
+
+
+def annuity_factor(
+    discount_rate: T,
+    lifetime: T,
+) -> T:
+    """Calculate annuity factor, handling NaN discount_rate.
```
lkstrp: Agree with @fneum 

`md` instead of `rst` and a claude clean up is needed here

>   This is a wrapper around :func:`annuity` that returns 1.0 where
    ``discount_rate`` is NaN. This is used internally when computing
    periodized costs - if discount_rate is NaN, it means the user
    provided capital_cost (already annuitized) instead of overnight_cost.

This is not a docstring, it's a justification for your prompt

pypsa/costs.py
```diff
@@ -0,0 +1,245 @@
...
+    1.0
+    >>> pypsa.costs.annuity_factor(0.0, 20)  # 0% rate = simple depreciation
+    0.05
+
+    """
+    if isinstance(discount_rate, pd.Series):
+        is_nan = discount_rate.isna()
+        safe_rate = discount_rate.where(~is_nan, 0.07)
+        if isinstance(lifetime, pd.Series):
+            safe_lifetime = lifetime.where(~lifetime.isna(), 25.0)
```
lkstrp: Where is the 25 coming from? Also 0.07? This just defines defaults out of nowhere?

pypsa/costs.py
```diff
@@ -0,0 +1,245 @@
...
+    discount_rate: T,
+    lifetime: T,
+    fom_cost: T | None = None,
+    nyears: float | pd.Series = 1.0,
+) -> T:
+    """Calculate fixed costs for the modeled horizon from capital or overnight cost.
+
+    This function calculates the total fixed cost for the modeled horizon by:
+
+    1. If ``overnight_cost`` is provided (not NaN): annuitize it using
```
lkstrp: rst

lkstrp (changes requested):
First batch of comments and the main question, why you wanna stuff properties on `c.da`. I think that isn't a wise design.

We are also very close to riding claude into lands of verbosity without editing

docs/user-guide/optimization/objective.md
```diff
@@ -26,6 +26,65 @@ To minimise **long-run annual system costs** (currency/a), capital costs for com
...
+          p_nom_extendable=True,
+          overnight_cost=1200000,  # Upfront cost: €/MW
+          discount_rate=0.07,      # 7% discount rate
+          lifetime=25,             # 25 years
+          fom_cost=12000)          # Fixed O&M: €/MW/a
+    ```
+
+    The effective capital cost used in optimization is calculated as:
+
+    $$c = c_{\text{overnight}} \cdot \text{annuity}(r, n) \cdot N_{\text{years}} + c_{\text{fom}}$$
```
fneum, line 60: `c_{\text{fom}}` should also be multiplied by number of years (since it has unit €/MW/a)

See comment below: Could also move FOM cost completely out of equation and treat as entirely separate objective term.

docs/user-guide/optimization/objective.md
```diff
@@ -26,6 +26,65 @@ To minimise **long-run annual system costs** (currency/a), capital costs for com
...
+
+    where the annuity factor converts overnight cost to annual payments:
+
+    $$\text{annuity}(r, n) = \frac{r}{1 - (1 + r)^{-n}}$$
```
fneum, line 64: or 1/n if r=0

docs/user-guide/optimization/objective.md
```diff
@@ -26,6 +26,65 @@ To minimise **long-run annual system costs** (currency/a), capital costs for com
...
+
+!!! tip "When to use overnight costs"
+
+    The overnight cost approach is useful when:
+
+    - Working with technology cost databases (e.g., [technology-data](https://github.com/pypsa/technology-data)) that provide overnight costs
+    - Different technologies have different discount rates or lifetimes
+    - You want to separately report overnight investment costs and fixed O&M in results
+
+    When `overnight_cost` is not set (default NaN), `capital_cost` is used directly, preserving backward compatibility.
```
fneum: I think this does not add much. It's clear that the overnight cost approach is useful if you have the data in disaggregated way.

docs/user-guide/optimization/objective.md
```diff
@@ -26,6 +26,65 @@ To minimise **long-run annual system costs** (currency/a), capital costs for com
...
+
+### Fixed Operation & Maintenance Costs
+
+Fixed O&M costs (`fom_cost`) represent annual costs that are incurred regardless of dispatch, such as maintenance, insurance, and land lease. They are added to the annualized overnight cost when using the overnight cost approach:
```
fneum: Actually, the FOM cost can just be added to the `capital_cost`, regardless whether they are given or calculated from `overnight_cost`, `discount_rate`, `lifetime`. 

I think it might be slightly cleaner if:

- `(capital_cost + fom_cost) * p_nom_opt`
- `(capital_cost(overnight_cost, discount_rate, lifetime) + fom_cost) * p_nom_opt)`

This way, you can have separated FOM cost, even if you are not using overnight calculation; and it's more uniform across the two methods (i.e. fom_cost is not included in capital_cost calcuation).

In this case, you'd have to edit objection function (`c` + `fom`)

docs/release-notes.md
```diff
@@ -34,6 +34,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - Add support for Python 3.14. Note that not all optional dependencies and solvers
   are available for Python 3.14 yet. (<!-- md:pr 1511 -->)
+- Add support for splitting `capital_cost` into overnight investment cost and fixed O&M. New component attributes `overnight_cost`, `discount_rate`, and `fom_cost` allow specifying overnight costs that are automatically annuitized during optimization. When `overnight_cost` is provided, PyPSA calculates the annuity using `discount_rate` and `lifetime` (supports 0% rate for simple depreciation). When `overnight_cost` is not set (default NaN), `capital_cost` is used. New `pypsa.costs` module with `annuity()`, `annuity_factor()`, and `periodized_cost()` functions. New statistics methods `n.statistics.overnight_cost()` and `n.statistics.fom()` for reporting. (<!-- md:pr 1507 -->)
```
fneum: I would split into two new features:

1. Separate FOM from `capital_cost` via new `fom` attribute. This is unrelated to annuitization support.
2. Annuitization support via new `overnight_cost` attribute.

pypsa/data/component_attrs/generators.csv
```diff
@@ -22,7 +22,10 @@ marginal_cost_quadratic,static or series,currency/MWh,0.,Quadratic marginal cost
 active,boolean,n/a,True,"Whether to consider the component in optimization or not",Input (optional)
 build_year,int,year,0,Build year of the generator.,Input (optional)
 lifetime,float,years,inf,Lifetime of the generator.,Input (optional)
-capital_cost,float,currency/MW,0,"Fixed period costs of extending `p_nom` by 1 MW, including periodized investment costs and periodic fixed O&M costs (e.g. annuitized investment costs).",Input (optional)
+capital_cost,float,currency/MW,0,"Annuitized investment cost per MW per year. Used directly in optimization unless `overnight_cost` is specified.",Input (optional)
```
fneum:
```suggestion
capital_cost,float,currency/MW,0,"Periodized investment cost per MW per year. Used directly in optimization unless `overnight_cost` is specified.",Input (optional)
```

pypsa/data/component_attrs/generators.csv
```diff
@@ -22,7 +22,10 @@ marginal_cost_quadratic,static or series,currency/MWh,0.,Quadratic marginal cost
...
+overnight_cost,float,currency/MW,NaN,"Overnight (upfront) investment cost per MW. If specified, PyPSA calculates annuity using `discount_rate` and `lifetime`. Takes precedence over `capital_cost`.",Input (optional)
+discount_rate,float,per unit,NaN,"Discount rate for annuity calculation when using `overnight_cost`. Supports 0% rate (simple depreciation).",Input (optional)
+fom_cost,float,currency/MW/a,0,"Fixed operation and maintenance costs per year, added to annuitized investment cost.",Input (optional)
```
fneum: Concretely, added to `capital_cost` whether calculated or not.

pypsa/network/index.py
```diff
@@ -622,6 +623,20 @@ def snapshot_weightings(self, df: pd.DataFrame) -> None:
...
+
+        Returns a Series indexed by investment period when the network has
+        investment periods.
+        """
+        if self.has_investment_periods:
+            return (
+                self.snapshot_weightings["objective"].groupby(level="period").sum()
+                / HOURS_PER_YEAR
+            )
+        return self.snapshot_weightings["objective"].sum() / HOURS_PER_YEAR
```
fneum, line 655: This is fine as long as `nyears` is only used for financial aspects (i.e. capital cost in the objective). Otherwise, it needs to be distinguished by "stores", "objective", "generators".

pypsa/statistics/expressions.py
```diff
@@ -913,6 +918,178 @@ def expanded_capex(  # noqa: D417
...
+        carrier: str | Sequence[str] | None = None,
+        bus_carrier: str | Sequence[str] | None = None,
+        nice_names: bool | None = None,
+        drop_zero: bool | None = None,
+        round: int | None = None,
+    ) -> pd.DataFrame:
+        """Calculate the **overnight investment costs** (excluding fom).
+
+        If ``discount_rate > 0``, returns ``capital_cost * annuity(discount_rate, lifetime) * capacity``.
+        If ``discount_rate == 0``, returns ``capital_cost * capacity`` (pre-annuitized).
```
fneum: This is a leftover from a previous implementation.

Should return `overnight_cost` * `p_nom_opt` if given, else `capital_cost / annuity_factor(lifetime, discount_rate)` (PVF calculation).

pypsa/statistics/expressions.py
```diff
@@ -913,6 +918,178 @@ def expanded_capex(  # noqa: D417
...
+
+        Returns
+        -------
+        pd.DataFrame
+            Annuitized overnight investment costs with components as rows and a single column
+            of aggregated values.
+
+        See Also
+        --------
+        capex : Returns total fixed costs (overnight_cost + fom_cost).
```
fneum: Not correct. Leftoever?
```suggestion
        capex : Returns periodized investment costs (without `fom_cost`).
```

pypsa/statistics/expressions.py
```diff
@@ -913,6 +918,178 @@ def expanded_capex(  # noqa: D417
...
+        fom_cost : Returns fixed operation and maintenance costs.
+
+        """
+
+        @pass_empty_series_if_keyerror
+        def func(n: Network, c: str, port: str) -> pd.Series:
+            comp = n.c[c]
+            capacity = comp.static[f"{nominal_attrs[c]}_opt"]
+            return capacity * comp.overnight_cost
```
fneum, line 983: The PVF calculation from `capital_cost`, `lifetime`, `discount_rate` should be done here as well, I would say.

pypsa/consistency.py
```diff
@@ -458,6 +458,58 @@ def check_assets(n: NetworkType, component: Components, strict: bool = False) ->
             )
 
 
+def check_cost_consistency(component: Components, strict: bool = False) -> None:
+    """Check if both overnight_cost and capital_cost are set for the same asset.
+
+    When both are specified, overnight_cost takes precedence and capital_cost is
+    ignored. This may indicate user confusion.
```
fneum:
```suggestion
    When both are specified, overnight_cost takes precedence and capital_cost is
    ignored.
```

pypsa/consistency.py
```diff
@@ -458,6 +458,58 @@ def check_assets(n: NetworkType, component: Components, strict: bool = False) ->
...
+        missing_discount_rate = has_overnight & static["discount_rate"].isna()
+        if missing_discount_rate.any():
+            assets = static.index[missing_discount_rate].tolist()
+            _log_or_raise(
+                True,
+                "Component %s has assets with 'overnight_cost' set but missing "
+                "'discount_rate': %s. Provide discount_rate for annuitization.",
+                component.name,
+                ", ".join(assets[:5]) + ("..." if len(assets) > 5 else ""),
+            )
```
fneum, line 510: There should also be a similar `lifetime` check.

docs/release-notes.md
FabianHofmann: but this is what is happening right? fom is absolute terms per MW (expected to be already periodized). only `overnight_cost` is periodized internally

docs/user-guide/optimization/objective.md
FabianHofmann: yes, this is perhaps a misunderstanding. it is a separate cost term and is not annuitized in the code. it is simply added to the objective. perhaps this paragraph should rather say. "The effective periodized cost per MW installation used in optimization is calculated as:"

docs/user-guide/optimization/objective.md
FabianHofmann: yes, this is what is happening. the text is misleading though, as above. updating it.

pypsa/components/array.py
FabianHofmann: It is not under `.da` ; it was in one of my previous attempts. but that did not feel good. it is a stand-alone components feature now. 

but I guess what you mean is that this is not the best module where to define it? can move it to `pypsa/components/components.py`

pypsa/common.py
FabianHofmann: doing

pypsa/consistency.py
FabianHofmann: done

pypsa/consistency.py
FabianHofmann: yes, probably better

pypsa/costs.py
FabianHofmann: legacy and laziness. it was int, float, xarray.DataArray, pd.Series before

pypsa/costs.py
FabianHofmann: thanks, good point!

pypsa/costs.py
FabianHofmann: rst => md clean up done

pypsa/costs.py
FabianHofmann: function is removed anyway

pypsa/statistics/expressions.py
FabianHofmann: yes, this is already done by using the `comp.overnight_cost` component-attribute

pypsa/statistics/expressions.py
FabianHofmann: done

pypsa/costs.py
FabianHofmann: fixed

pypsa/data/component_attrs/generators.csv
FabianHofmann: have adjusted the docstring

pypsa/data/component_attrs/generators.csv
FabianHofmann: done

docs/user-guide/optimization/objective.md
FabianHofmann: done

pypsa/common.py
FabianHofmann: @lkstrp deprecation now aligning with the convention?

FabianHofmann:
@lkstrp @fneum thanks for the reviews. I addressed all of them. And yes, I still have to learn to better handle iterative changes with CC

FabianHofmann:
@fneum I guess two things were not clear: 

1. `fom_cost` is expected to be periodized by the user, just as capital_costs. they are added wo periodization. I have updated the docs, they were a bit vague
2. back-calculation in `n.c.<component>.overnight_cost` is automatically done internally. this will raise an error now if `lifetime` and `discount_rate` are not given.

docs/user-guide/optimization/objective.md
```diff
@@ -26,6 +26,57 @@ To minimise **long-run annual system costs** (currency/a), capital costs for com
 
 If no extendable components are present, only the dispatch of the components is optimised as in a **short-run market model**.
 
+### Specifying Capital Costs
+
+PyPSA supports two approaches for specifying investment costs:
+
+=== "Direct capital_cost (default)"
```
fneum:
```suggestion
=== "Direct periodized capital cost (default)"
```

PR #1508, opened 2026-01-03
feat: Add temporal clustering functionality

FabianHofmann:

This PR implements temporal clustering functionality for PyPSA, enabling users to reduce the temporal resolution of networks while preserving total modeled hours through snapshot weighting adjustments. This is an upstream implementation based on patterns from PyPSA-EUR.

### New Functions in `pypsa.clustering.temporal`

- **`resample(n, offset)`** - Aggregate snapshots at regular intervals (e.g., "3h", "6h", "24h")
- **`downsample(n, stride)`** - Select every Nth snapshot as representative with scaled weights
- **`segment(n, num_segments)`** - Use TSAM k-means clustering for variable-duration segments
- **`from_snapshot_map(n, snapshot_map)`** - Apply pre-computed temporal aggregation mapping

### Accessor Pattern

All methods are accessible via `n.cluster.temporal.*`:

```python
# Resample to 3-hour resolution
m = n.cluster.temporal.resample("3h")

# Select every 4th snapshot  
m = n.cluster.temporal.downsample(4)

# TSAM segmentation (requires tsam package)
m = n.cluster.temporal.segment(100)

# Get full result with mapping for disaggregation
result = n.cluster.temporal.get_resample_result("3h")
print(result.snapshot_map)  # Original -> aggregated mapping
```

### Additional Features

- **`n.nyears` property** - Returns total modeled years based on snapshot weightings
- **`TemporalClustering` dataclass** - Returns both clustered network and snapshot mapping
- **Configurable aggregation rules** - Default mean, with min for `e_max_pu` and max for `e_min_pu`
- **Multi-period support** - `resample()` handles networks with investment periods
- **Leap day handling** - Transfer weights to March 1 rather than dropping hours
- **`tsam` optional dependency** - For `segment()` function

### Weight Preservation Invariant

All methods preserve the invariant: `sum(weights) == total_modeled_hours`

```python
original_hours = n.snapshot_weightings["objective"].sum()
m = n.cluster.temporal.resample("3h")
assert np.isclose(m.snapshot_weightings["objective"].sum(), original_hours)
```

## Checklist


pypsa/clustering/__init__.py
```diff
@@ -15,6 +15,81 @@
 if TYPE_CHECKING:
     from pypsa import Network
     from pypsa.clustering.spatial import Clustering
```
fneum, line 17: Long-term: Should be renamed to `SpatialClustering`.

test/test_temporal_clustering.py
```diff
@@ -0,0 +1,466 @@
...
+    n.add("Line", "line01", bus0="bus0", bus1="bus1", s_nom=100, x=0.1, r=0.01)
+
+    return n
+
+
+@pytest.fixture
+def yearly_network():
+    """Create a network with full year hourly resolution."""
+    n = pypsa.Network()
+    n.set_snapshots(pd.date_range("2020-01-01", periods=8760, freq="h"))
```
fneum: 2020 was a leap year with 8784 hours. Probably not a good example to use with 8760 hours.

pypsa/network/index.py
```diff
@@ -622,6 +622,28 @@ def snapshot_weightings(self, df: pd.DataFrame) -> None:
...
+        >>> n.nyears  # doctest: +SKIP
+        1.0
+        >>> n.cluster.temporal.resample("3h")  # doctest: +SKIP
+        >>> n.nyears  # Still 1.0, weights are summed  # doctest: +SKIP
+        1.0
+
+        """
+        from pypsa.clustering.temporal import HOURS_PER_YEAR  # noqa: PLC0415
+
+        return self.snapshot_weightings["objective"].sum() / HOURS_PER_YEAR
```
fneum, line 645: Is there a specific reason for using `objective` weightings here and not considering `generators` or `stores`?

test/test_temporal_clustering.py
```diff
@@ -0,0 +1,466 @@
...
+        result = resample(n, "24h")
+
+        assert len(result.n.snapshots) == 7
+
+    def test_resample_time_series_aggregated(self, simple_network):
+        n = simple_network
+        result = resample(n, "3h")
+
+        assert "p_max_pu" in result.n.c.generators.dynamic
+        assert len(result.n.c.generators.dynamic["p_max_pu"]) == len(result.n.snapshots)
```
fneum, line 117: Could also check that means of `p_max_pu` are correctly calculated in aggregated network.

test/test_temporal_clustering.py
```diff
@@ -0,0 +1,466 @@
...
+        assert len(m.snapshots) == 168 // 3
+
+
+class TestDownsample:
+    def test_downsample_stride_4(self, simple_network):
+        n = simple_network
+        result = downsample(n, 4)
+
+        assert isinstance(result, TemporalClustering)
+        assert len(result.n.snapshots) == 168 // 4
```
fneum, line 144: Could test what happens if number of snapshots is not divisible by downsample frequency.

test/test_temporal_clustering.py
```diff
@@ -0,0 +1,466 @@
...
+        result = downsample(n, 1)
+
+        assert len(result.n.snapshots) == len(n.snapshots)
+
+    def test_large_stride(self, simple_network):
+        n = simple_network
+        result = downsample(n, 168)
+
+        assert len(result.n.snapshots) == 1
+        assert np.isclose(result.n.snapshot_weightings["objective"].iloc[0], 168)
```
fneum, line 337: Could test case where frequency is larger then length of snapshots.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
 # SPDX-License-Identifier: MIT
 
-"""Functions for temporal clustering of networks."""
+"""Functions for temporal clustering of networks.
+
+This module provides methods to reduce the temporal resolution of PyPSA networks
+while preserving the total modeled hours through snapshot weighting adjustments.
+
+The core abstraction is **snapshot weighting** - each snapshot has a weight
+representing the number of hours it represents.
```
fneum: This doesn't add much.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
 #
 # SPDX-License-Identifier: MIT
 
-"""Functions for temporal clustering of networks."""
+"""Functions for temporal clustering of networks.
+
+This module provides methods to reduce the temporal resolution of PyPSA networks
+while preserving the total modeled hours through snapshot weighting adjustments.
```
fneum:
```suggestion
while preserving the total modeled hours through snapshot weighting adjustments, so that the total number of hours modelled is kept invariant.
```

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+from typing import TYPE_CHECKING
+
+import numpy as np
+import pandas as pd
+
+if TYPE_CHECKING:
+    from pypsa import Network
+
+
+HOURS_PER_YEAR = 8760
```
fneum, line 51: Simplification given that there are leap years. Possibly, a more accurate number of years could be inferred from `pd.DateTimeIndex`.

pypsa/clustering/temporal.py
fneum, line 1: What would you think about warning about applying temporal clustering on solved networks? Storage state of charge and dispatch values could easily become inconsistent.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+        Pandas offset string (e.g., "3h", "6h", "24h").
+    drop_leap_day : bool, default False
+        Transfer Feb 29 weights to March 1.
+    aggregation_rules : dict, optional
+        Override default aggregation per attribute.
+
+    Returns
+    -------
+    TemporalClustering
+        Result with clustered network and snapshot mapping.
```
fneum, line 198: Should we state that this is not an inplace operation? I guess that's implied by returning a `TemporalClustering` object, but in the accessor, only the network is returned where it could be unclear.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    """Resample network to coarser temporal resolution.
+
+    Parameters
+    ----------
+    n : Network
+        Network with original resolution.
+    offset : str
+        Pandas offset string (e.g., "3h", "6h", "24h").
+    drop_leap_day : bool, default False
+        Transfer Feb 29 weights to March 1.
```
fneum: I thought the point of the `drop_leap_day` option is to ensure all networks with 1 year have 8760 as sum of snapshot weightings (regardless of whether they are leap years or not). Then it does not make sense to transfer the weights, and it would be fine if the sum of snapshot weightings were reduced by 24 hours.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+        Mapping from original to resampled snapshots.
+
+    """
+    if isinstance(original_snapshots, pd.DatetimeIndex):
+        resampler = pd.Series(index=original_snapshots, dtype=object)
+        for ts in resampled_snapshots:
+            mask = (original_snapshots >= ts) & (
+                original_snapshots < ts + pd.Timedelta(offset)
+            )
+            resampler.loc[mask] = ts
```
fneum: Should be vectorised (think about the 70 years of hourly data we solved in one model https://www.sciencedirect.com/science/article/pii/S014098832500307X).

If we can assume that indices of `resampled_snapshots` are subset of `original_snapshots` (not sure), could do something like:

```py
pd.Series(resampled_snapshots, index=resampled_snapshots).reindex(original_snapshots).ffill()
```

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+
+    """
+    if n.has_periods:
+        return _resample_with_periods(
+            n, offset, drop_leap_day=drop_leap_day, aggregation_rules=aggregation_rules
+        )
+
+    m = n.copy()
+
+    original_snapshots = n.snapshots
```
fneum, line 225: The `resample` only works for `pd.DatetimeIndex`. Should there be an assert here?

docs/release-notes.md
```diff
@@ -5,13 +5,30 @@ SPDX-License-Identifier: CC-BY-4.0
...
+## Upcoming Release
 
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
-    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.-->
+    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
+
+### Enhancements
```
fneum, line 16: Before merging, the new functionality will have to be integrated into the documentation:

- [ ] Here: https://docs.pypsa.org/latest/api/networks/cluster/
- [ ] User guide after https://docs.pypsa.org/latest/user-guide/clustering/

docs/release-notes.md
```diff
@@ -5,13 +5,30 @@ SPDX-License-Identifier: CC-BY-4.0
...
+
+- Add temporal clustering functionality via `n.cluster.temporal.*` accessor. (<!-- md:pr 1508 -->)
+
+    - `resample(offset)` - Aggregate snapshots at regular intervals (e.g., "3h", "24h")
+    - `downsample(stride)` - Select every Nth snapshot as representative
+    - `segment(num_segments)` - TSAM k-means clustering for variable-duration segments
+    - `from_snapshot_map(snapshot_map)` - Apply pre-computed temporal aggregation
+
+    All methods preserve the snapshot weighting invariant and return the clustered network.
+    Use `get_*_result()` methods to also get the snapshot mapping for disaggregation.
```
fneum: Add example, otherwise it's unclear what has to be filled for `*`.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    years = pd.DatetimeIndex(n.snapshots).year.unique()
+    snapshot_weightings_list = []
+    for year in years:
+        year_mask = n.snapshots.year == year
+        sws_year = n.snapshot_weightings[year_mask]
+        sws_year = sws_year.resample(offset).sum()
+        snapshot_weightings_list.append(sws_year)
+    snapshot_weightings = pd.concat(snapshot_weightings_list)
+
+    snapshot_weightings = snapshot_weightings.query("objective != 0")
```
fneum, line 240: I didn't remember what this filter was for. Maybe carry over some of the comments from PyPSA-Eur version:

```py
# The resampling produces a contiguous date range. In case the original
# index was not contiguous, all rows with zero weight must be dropped
# (corresponding to time steps not included in the original snapshots).
```

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    if n.has_periods:
+        return _resample_with_periods(
+            n, offset, drop_leap_day=drop_leap_day, aggregation_rules=aggregation_rules
+        )
+
+    m = n.copy()
+
+    original_snapshots = n.snapshots
+
+    years = pd.DatetimeIndex(n.snapshots).year.unique()
```
fneum, line 228: Maybe add comment here, that year-wise resampling handles non-contiguous years.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+
+
+def downsample(
+    n: Network,
+    stride: int,
+) -> TemporalClustering:
+    """Select every Nth snapshot as representative.
+
+    Weightings are scaled by stride to preserve total modeled hours.
+    Time series are not aggregated - values at selected timestamps are used.
```
fneum:
```suggestion
    Time series are not aggregated - values at selected snapshots are used.
```

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    >>> assert len(clustered.snapshots) == len(n.snapshots) // 4  # doctest: +SKIP
+
+    """
+    if stride < 1:
+        msg = f"stride must be >= 1, got {stride}"
+        raise ValueError(msg)
+
+    if n.has_periods:
+        msg = "downsample() does not yet support networks with investment periods"
+        raise NotImplementedError(msg)
```
fneum: Is there a challenge here? I would imagine you just do the downsampling separately by investment period.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+
+    m = n.copy()
+    original_snapshots = n.snapshots
+
+    selected = n.snapshots[::stride]
+    m.set_snapshots(selected)
+
+    new_weightings = n.snapshot_weightings.loc[selected].copy()
+    new_weightings *= stride
```
fneum, line 463: I think a (frequent) edge case is not handled here, namely when the number of snapshots is not divisible by `stride`.

Options:
1. the last snapshot might have a different weighting
2. the function throws an error in this case.

I prefer Option 1.

Otherwise, the sum of snapshot weightings is not invariant to clustering.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    with optimal duration allocation.
+
+    Parameters
+    ----------
+    n : Network
+        Network with original resolution.
+    num_segments : int
+        Target number of segments.
+    solver : str, default "highs"
+        MIP solver for duration optimization.
```
fneum:
```suggestion
        MIP solver for time clustering.
```

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    num_segments: int,
+    *,
+    solver: str = "highs",
+    exclude_attrs: list[str] | None = None,
+    aggregation_rules: dict[str, str] | None = None,
+) -> TemporalClustering:
+    """Cluster time series into variable-duration segments using TSAM.
+
+    Uses k-means clustering to identify representative time segments
+    with optimal duration allocation.
```
fneum: I don't think `tsam` uses k-means for segmentation clustering.

> Agglomerative clustering of adjacent time steps

Source: https://tsam.readthedocs.io/en/latest/segmentationDoc.html

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+        Result with segmented network.
+
+    Note
+    ----
+    Requires `tsam` package: pip install tsam
+
+    """
+    if num_segments < 1:
+        msg = f"num_segments must be >= 1, got {num_segments}"
+        raise ValueError(msg)
```
fneum, line 526: Should also check for maximum `num_segements`, right?

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+                    col_attrs.append((c.name, attr, col))
+
+    if not dfs:
+        msg = "No time-varying data found for segmentation"
+        raise ValueError(msg)
+
+    combined = pd.concat(dfs, axis=1)
+    combined.columns = range(len(combined.columns))
+
+    annual_max = combined.max().replace(0, 1)
```
fneum: Rename `annual_max` since clustering does not need to be applied on single-year network.

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+            m._import_series_from_df(data_col, comp, attr, overwrite=True)
+
+    snapshot_map = pd.Series(index=original_snapshots, dtype=object)
+    for i, new_ts in enumerate(new_snapshots):
+        start = offsets[i]
+        if i + 1 < len(offsets):
+            end = offsets[i + 1]
+        else:
+            end = len(original_snapshots)
+        snapshot_map.iloc[start:end] = new_ts
```
fneum: Ideally, this should be vectorised as well (considering there could be `70*8760` snapshot networks).

pypsa/clustering/temporal.py
```diff
@@ -2,4 +2,578 @@
...
+    df_normalized = df_normalized.fillna(0)
+
+    agg = tsam_module.TimeSeriesAggregation(
+        df_normalized,
+        hoursPerPeriod=len(df_normalized),
+        noTypicalPeriods=1,
+        noSegments=num_segments,
+        segmentation=True,
+        solver=solver,
+    )
```
fneum, line 577: I don't think it's necessary for the first version, but I will just note that unlike the `resample` function, the `segment` function does not handle non-contiguous years gracefully.

pypsa/network/index.py
```diff
@@ -622,6 +622,28 @@ def snapshot_weightings(self, df: pd.DataFrame) -> None:
         if options.debug.runtime_verification:
             _assert_data_integrity(self)
 
+    @property
+    def nyears(self) -> float:
```
fneum, line 626: This property is not used anywhere, is it?

fneum:
Took the liberty to update `time-series-aggregation.ipynb` example. Reproduces same results with new functions :+1:

pypsa/clustering/__init__.py
FabianHofmann: yes

pypsa/network/index.py
FabianHofmann: no reason, just that objective is the most common one and I would argue that the sum of weights should be the same across `objective`, `generators` or `stores` columns.

test/test_temporal_clustering.py
FabianHofmann: sure, I think it does not make a difference here, but will harmonize that

FabianHofmann: done

test/test_temporal_clustering.py
FabianHofmann: done

pypsa/clustering/temporal.py
FabianHofmann: I added a warning if the users applies the clustering to a solved network

docs/release-notes.md
FabianHofmann: done

pypsa/network/index.py
FabianHofmann: it is used in the tests only for simplicity. however we can keep it and merge with the implementation in #1507

pypsa/clustering/temporal.py
FabianHofmann: yes. done

FabianHofmann:
@fneum thanks for the review, I (hope I) addressed all of your points. the resample is now much saver (datetime index required, multiperiod support, stoch networks raising error, zero/undivisble arguments tested). have a look again if you want

pypsa/network/index.py
fneum: Well, there are valid cases for non-uniform weightings, e.g. when using non-contiguous representative periods. But fine for now; we can address that when supporting clustering algorithms for representative days/weeks.

PR #1509, opened 2026-01-04
Add include_objective_constant parameter to optimize

### LINKED ISSUE #503: Remove objective constant from objective function assignment
We are currently adding the cost of pre-installed capacity to the objective function. This leads to large ranges for variables and the objective in the LP. I'd suggest to remove the assignment of the constant to the objective function and leave the `objectiv_constant` as a network property that comprises these cost. For back-wards compatibility we can still add this constant  to `n.objective` after solving but in the long-term I'd suggest to remove that as well.

FabianHofmann:
## Summary

Implements the feature request from #503:

- Add `include_objective_constant` parameter to `n.optimize()` and `n.optimize.create_model()` to control whether the objective constant (capital costs of existing infrastructure) is included as a variable in the objective function
- When `include_objective_constant=None` (default), raises a `FutureWarning` indicating the default will change from `True` to `False` in a future version
- Setting to `False` improves LP numerical conditioning by not including the constant as a variable
- The `objective_constant` is still calculated and stored on the network regardless of this parameter

## Checklist

Closes #503


pypsa/optimization/optimize.py
```diff
@@ -170,7 +175,7 @@ def define_objective(n: Network, sns: pd.Index) -> None:
     else:
         n._objective_constant = float(constant)
         has_const = constant != 0
-    if has_const:
+    if has_const and include_objective_constant:
```
fneum: Works, but couldn't the check for `include_objective_constant` be done earlier, also skipping the constant calculation if it is not required?

FabianHofmann: yes, we can to that. that would mean `n.objective_constant` is nan/zero with `include_objective_constant` is false (now it is still pointing to the same value but not included in the objective function). I guess that is even clearer, so we don't rely on `n.objective_constant` anymore

docs/release-notes.md
```diff
@@ -5,13 +5,18 @@ SPDX-License-Identifier: CC-BY-4.0
...
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
-    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.-->
+    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
+
+### Enhancements
+
+- Add `include_objective_constant` parameter to `n.optimize()` and `n.optimize.create_model()` to control whether the objective constant (capital costs of existing infrastructure) is included as a variable in the objective function. Setting to `False` improves LP numerical conditioning. The default will change from `True` to `False` in a future version. (<!-- md:pr 503 -->)
```
lkstrp:
```suggestion
- Add `include_objective_constant` parameter to `n.optimize()` and `n.optimize.create_model()` to control whether the objective constant (capital costs of existing infrastructure) is included as a variable in the objective function. Setting to `False` improves LP numerical conditioning. The default will change from `True` to `False` in a future version. (<!-- md:pr 1509 -->)
```

pypsa/optimization/optimize.py
```diff
@@ -434,6 +448,12 @@ def __call__(
         compute_infeasibilities : bool, default False
             Whether to compute and print Irreducible Inconsistent Subsystem (IIS) in case
             of an infeasible solution. Requires Gurobi.
+        include_objective_constant : bool | None, default None
+            Whether to include the objective constant (capital costs of existing
+            infrastructure) as a variable in the objective function. If None (default),
+            a FutureWarning is raised indicating that the default behavior will change
+            to False in a future version. Set to True to keep the current behavior
+            (include constant), or False to exclude it (improves LP numerical conditioning).
```
lkstrp:
```suggestion
            Whether to include the objective constant (capital costs of existing
            infrastructure) as a variable in the objective function. Setting to False
            improves LP numerical conditioning. Defaults to True.
```

pypsa/optimization/optimize.py
```diff
@@ -457,6 +477,18 @@ def __call__(
         if solver_options is None:
             solver_options = options.params.optimize.solver_options.copy()
 
+        # Handle include_objective_constant deprecation
+        if include_objective_constant is None:
+            warnings.warn(
+                "The default value of `include_objective_constant` will change from "
+                "True to False in a future version. Set `include_objective_constant` "
```
lkstrp:
```suggestion
                "True to False in version 2.0. Set `include_objective_constant` "
```

pypsa/optimization/optimize.py
lkstrp: Would that mean we can get rid of it entirely mid-term? Or would you want to use `include_objective_constant=true` if it's not just for keeping existing code intact

FabianHofmann: not sure. I don't know any person really making sense of the objective constant. but if there is any, we could keep the `include_objective_constant` as an argument and only change the default to false in v2.

test/conftest.py
```diff
@@ -15,6 +15,7 @@
 from pypsa.constants import DEFAULT_EPSG
 
 pypsa.options.debug.runtime_verification = True
+pypsa.options.params.optimize.include_objective_constant = True
```
lkstrp, line 18: Why did you added this?

PR #1515, opened 2026-01-06
Skip the subset check in `as_index()` (in some cases)

chriselion:
Closes # (if applicable): n/a

I was doing some basic profiling, and it looked like a decent portion of the setup time was being spent in this part of the code: 
https://github.com/PyPSA/PyPSA/blob/59a7d5d0177c1beb26e6e436a7b497637ec11104/pypsa/common.py#L275-L280

In the case the input `values` is `None`, since we set `values_ = n_attr`, the subset check is redundant and we can skip that block.

(more timing information below)

## Checklist

- [x] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `docs`.  _(n/a it's a one-line change with a comment)_
- [x] Unit tests for new features were added (if applicable). _(n/a)_

---

I used this to measure the total time:
```python
from pypsa.networks import Network
import time

PATH = 'examples/networks/model-energy/model-energy.nc'

def main():

    start_time = time.monotonic()
    n = Network(PATH)
    n.optimize(solver_name="highs", log_to_console=False)
    
    end_time = time.monotonic()
    print(f"elapsed: {end_time - start_time}")

if __name__ == "__main__":
    main()
```

Running before and after the change 5 times each, the best times were:
before: `elapsed: 7.80225883400999`
after: `elapsed: 7.512157290999312`

So it's not a huge difference overall, but since most of that time is spent in the solver, I'd say it's still worthwhile.

---

To profile where the time was going, I used the technique described here: https://www.youtube.com/watch?v=ey_P64E34g0
In order to exclude the solver time from the profile, I used this to save a `pstats` file:
```python
pr = cProfile.Profile()
pr.enable()
def stop_profile(n, snapshots):
    pr.disable()
n.optimize(solver_name="highs", log_to_console=False, extra_functionality=stop_profile)
pr.dump_stats("model-energy.optimize.pstats")
```
and converted to an svg with
```
gprof2dot model-energy.optimize.pstats -n 5 | dot -Tsvg -o model-energy.optimize.svg
```
That produced a graph that looked like this:
image
So roughly half of the non-solver time was spent in `as_index()`, in particular the `all` block (although take the actual times with a grain of salt, because the profiler introduces some overhead)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
 # Run ruff to lint and format
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.13.3
+  rev: v0.14.10
   hooks:
     # Run the linter.
   - id: ruff
@@ -53,6 +53,6 @@ repos:
 
 # Check for FSFE REUSE compliance (licensing)
 - repo: https://github.com/fsfe/reuse-tool
-  rev: v6.0.0
+  rev: v6.2.0
   hooks:
   - id: reuse-lint-file


docs/release-notes.md
```diff
@@ -24,6 +24,10 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - Fix `pypsa.common.annuity` function to correctly handle discount rate of 0. (<!-- md:pr 1512 -->)
 
+### Minor improvements
+
+- Speed up OptimizationAccessor.create_model() by avoiding some redundant checks. (<!-- md:pr 1515 -->)
```
lkstrp:
```suggestion
- Speed up creating the model by avoiding some redundant checks. (<!-- md:pr 1515 -->)
```

chriselion: Thanks, wasn't sure how detailed to get there.

chriselion:
@lkstrp Thanks, I'll have a look at that other issue later. I'm still learning my way around the code (and don't know pandas very well) but hopefully I can find something to improve.

Are there any larger examples that would be worth examining? `examples/networks/model-energy` looked like it was the slowest.

PR #1519, opened 2026-01-07
docs: add partial elastic demand and log-log approximation

flxlchr:
Addition of partial demand elasticity and a piecewise-linear demand curve as approximation for a log-log demand curve as proposed in Brown, Neumann, Riepin (2025)

Closes # (if applicable).

## Checklist

### DIFF
--- a/docs/examples/demand-elasticity.ipynb
+++ b/docs/examples/demand-elasticity.ipynb
@@ -465,7 +465,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "n.optimize(solver_name=\"gurobi\")",
+    "n.optimize(solver_name=\"gurobi\")"
    ]
   },
   {
@@ -559,7 +559,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "n.optimize(solver_name=\"gurobi\")",
+    "n.optimize(solver_name=\"gurobi\")"
    ]
   },
   {


flxlchr:
I originally used Gurobi locally because due to the quadratic nature of the added examples, they did not converge with HiGHS on my machine. Since Gurobi is not usable on Read the Docs, I switched back to HiGHS in a follow-up commit. With HiGHS the example now runs into a timeout during the RTD docs build...

flxlchr:
Thanks for your comment @lkstrp 

I only updated the existing notebook and did not change its scope or placement, so I assumed no mkdocs.yml change was necessary. Happy to adjust navigation if you think it should be referenced differently.

I now adjusted the number of days, that are included. Now one day per month instead of every 5th. Works with HiGHS now, will update the file the next days

lkstrp:
> I only updated the existing notebook and did not change its scope or placement, so I assumed no mkdocs.yml change was necessary. Happy to adjust navigation if you think it should be referenced differently.

Ah, I missed that, thanks!

> I now adjusted the number of days, that are included. Now one day per month instead of every 5th. Works with HiGHS now, will update the file the next days

Reducing the complexity would be the best. We might also have fixed this issue with #1526 hopefully soon

flxlchr:
I’ve now committed and pushed the simplified version that works with HiGHS.

docs/examples/demand-elasticity.ipynb
```diff
@@ -82,27 +92,13 @@
...
-   ]
-  },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To save some computation time, we will sample just every fifth day of the year. Each day is considered at 3-hourly resolution, so we will have 8 snapshots per representative day."
+    "To save computation time, we will only sample the first day of each month. Each day will be considered at a 3-hourly resolution, so there will be eight snapshots per representative day.\n",
+    "\n",
+    "For computational reasons, the correct snapshot weighting is not applied here. Using smaller weights keeps the objective coefficients smaller and improves numerical stability, which is necessary for reliable convergence with the default HiGHS solver. As a consequence, the resulting marginal costs should be interpreted in relative terms. When using a more robust solver such as Gurobi, the weighting can be re-enabled."
```
fneum: I would like to keep a consistent weighting.One option is to use Gurobi for this example and revert to every 5th day.

fneum (approved):
Excellent work @flxlchr! Thanks for completing this example!

PR #1530, opened 2026-01-19
fix: pathlib paths in collections

### LINKED ISSUE #1525: NetworkCollection(...) should support same path-like types as Network(...)
`pypsa.NetworkCollection` does currently not support `Path` objects and requests `str`-like paths for network files. It should support the same datatypes as `pypsa.Network(...)` does for loading from file.

lkstrp:
Closes #1525


lkstrp:
> Thanks.
> 
> Would also be nice to be able to pass `dict` objects and have `nc.items()`, such that on can iterate like
> 
> ```python
> for name, network in nc.items():
> ...
> ```

I agree on passing `dicts`. But `nc.items` will not work because the index can be multi dimensional, which will also have an effect on statistics and plots. You can always loop on `nc.networks`.

PR #1531, opened 2026-01-19
feat: set pypsa options via `.env` file

lkstrp:
Addition to #1513

### DIFF
--- a/test/test_options.py
+++ b/test/test_options.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 import importlib
+import importlib.util
+import os
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/options.md
```diff
@@ -42,16 +42,17 @@ For example, to set the default solver to Gurobi:
...
 ### Priority
 
 When the same option is set in multiple ways, the following priority applies (lowest to highest):
 
 1. **Default values**: Defined in PyPSA
-2. **Environment variables**: `PYPSA_*` (loaded once at import time)
-3. **Runtime setting**: `pypsa.set_option()` or `pypsa.options.x = ...`
-4. **Function arguments** (where applicable): e.g., `n.optimize(solver_name="gurobi")`
+2. **`.env` file**: Via `python-dotenv` (if installed)
+3. **Environment variables**: `PYPSA_*` (loaded once at import time)
```
euronion, line 53: Shouldn't the priority be switched? `.env` with higher prio than `env vars`? Or am I confused?

euronion:
Sweet <3

docs/user-guide/options.md
lkstrp: Convention is that env vars have higher priority, which is also default of `dotenv`

Issue #592, opened 2023-03-31
Links ramp_start_up_limit not behaving as expected


pmcwhannel:
## Checklist

- [ ] I am using the current version of PyPSA with the recent pull request #582.

## Describe the Bug

Seems the `ramp_limit_start_up` for link components is not working as intended. I have tried rerunning the the contents of `test_link_unit_commitment(api)` with the addition of the attribute `ramp_limit_start_up=0.01` and the output corresponding to OCGT does not change. Code is below to reproduce.

Thanks for any help in advance.

## Code
```
n = pypsa.Network()

snapshots = range(4)

n.set_snapshots(snapshots)

n.madd("Bus", ["gas", "electricity"])

n.add("Generator", "gas", bus="gas", marginal_cost=10, p_nom=20000)

n.add(
    "Link",
    "OCGT",
    bus0="gas",
    bus1="electricity",
    committable=True,
    p_min_pu=0.1,
    efficiency=0.5,
    up_time_before=0,
    ramp_limit_start_up=0.01, # CHANGE IS HERE
    min_up_time=3,
    start_up_cost=3333,
    p_nom=12000,
)

n.add(
    "Generator",
    "wind",
    bus="electricity",
    p_nom=800,
)

n.add("Load", "load", bus="electricity", p_set=[4000, 6000, 800, 5000])

n.optimize()

```

## Error Message

N/A.

fneum:
Confirmed bug. (Even when setting `down_time_before=1` as well).

PR #1540, opened 2026-01-28
Fix: ramp_limit_start_up attribute is ignored in binary UC

Irieo:
Closes # https://github.com/PyPSA/PyPSA/issues/592 (considering the implications of the bug, this should have rather been high priority)

## Problem

The `ramp_limit_start_up` and `ramp_limit_shut_down` attributes were completely ignored in binary unit commitment, even when explicitly set. 

This happened because:

a) Constraint creation check only evaluated `ramp_limit_up`/`ramp_limit_down`, ignoring values of start_up/shut_down
b) The ramping constraint `p[t] - p[t-1] <= ramp_up * p_nom * status[t-1] + ramp_start_up * p_nom * (status[t] - status[t-1])` was not applied for the first snapshot due to an overlap with the rolling horizon logic.

For more detail on binary UC formulation see [these docs](https://docs.pypsa.org/latest/user-guide/optimization/unit-commitment/#rampin) 

## Fix

1. (optional) The fact that `ramp_limit_up` and `ramp_limit_start_up` have defaults `NaN` and `1` accordingly adds a touch of madness here. In this implementation, both defaults are aligned to NaN. Leave it to @lkstrp's moral compass. 
2. Constraint is now created if either `ramp_limit_start_up`/down or `ramp_limit_up`/down are explicitly set.
3. New special constraint for the first snapshot for the non-rolling horizon case to set start-up limits

## Other

- [x]  Few RH tests fail, likely due to the new defaults. Can fix those or do revert defaults.
- [x]  Treatment of ramp_limit_shut_down for the first snapshot

## Reproduce

```
import pypsa
n = pypsa.Network()
snapshots = range(4)
n.set_snapshots(snapshots)

n.add("Bus", ["gas", "electricity"])
n.add("Generator", "gas", bus="gas", marginal_cost=10, p_nom=20000)

n.add(
    "Link",
    "OCGT",
    bus0="gas",
    bus1="electricity",
    committable=True,
    p_min_pu=0.1,
    efficiency=0.5,
    ramp_limit_start_up=0.4,  # This must have an effect
    # ramp_limit_up=0.2,
    start_up_cost=3333,
    p_nom=10000,
    up_time_before=0,
)

n.add(
    "Generator", "expensive_backstop", bus="electricity", p_nom=5000, marginal_cost=1e5
)
n.add("Load", "load", bus="electricity", p_set=[4000, 5000, 2000, 5000])

n.optimize()

print(n.c.links.dynamic.status)
print(n.c.links.dynamic.start_up)
print(-n.c.links.dynamic.p1)
print(n.c.generators.dynamic.p)

```

NB for testing `ramp_limit_shut_down` functionality that prevents immediate shutdown when the unit starts ON (eventually forcing production even with zero demand) you might want to add slack variable to trace correct behavior in otherwise infeasible problem:

```
n.add(
    "Generator",
    "slack_dump",
    bus="electricity",
    p_nom=20000,
    marginal_cost=1e6,
    sign=-1,
)
```

## Checklist

### DIFF
--- a/pypsa/optimization/constraints.py
+++ b/pypsa/optimization/constraints.py
@@ -598,9 +598,10 @@ def define_ramp_limit_constraints(
     ramp_limit_down = c.da.ramp_limit_down.sel(snapshot=sns)
     all_null = (ramp_limit_up.isnull() & ramp_limit_down.isnull()).all()
 
-    if not c.committables.empty:
-        ramp_limit_start_up = c.da.ramp_limit_start_up
-        ramp_limit_shut_down = c.da.ramp_limit_shut_down
+    com_i = c.committables.difference(c.inactive_assets)
+    if not com_i.empty:
+        ramp_limit_start_up = c.da.ramp_limit_start_up.sel(name=com_i)
+        ramp_limit_shut_down = c.da.ramp_limit_shut_down.sel(name=com_i)
         all_null = (
             all_null
             and (ramp_limit_start_up.isnull() & ramp_limit_shut_down.isnull()).all()
@@ -629,7 +630,6 @@ def define_ramp_limit_constraints(
     p = m[f"{c.name}-{var_attr}"]
 
     # Get different component groups for constraint application
-    com_i = c.committables.difference(c.inactive_assets)
     fix_i = c.fixed.difference(c.inactive_assets)
     fix_i = fix_i.difference(com_i).rename(fix_i.name)
     ext_i = c.extendables.difference(c.inactive_assets)


Irieo:
The implementation `ramp_limit_shut_down` for the snapshot 0 appeared to be not an easy thing. I suggest implementation as here, so that unit test with such parametrisation as here https://github.com/PyPSA/PyPSA/pull/1540/changes/7424cd3113eed527be111763fa682f4dd269eb52 yields for print(n.c.generators.dynamic.p)

```
| snapshot | gas    | expensive_backstop | slack_dump |
|----------|--------|-------------------|------------|
| 0        | 9000.0 | 0.0               | 4500.0     |
| 1        | 10000.0| 0.0               | 0.0        |
| 2        | 1000.0 | 1500.0            | 0.0        |
| 3        | 0.0    | 0.0               | 0.0        |
```
This is not unambiguous, so @FabianHofmann check carefully if it follows your understanding.

fneum (approved):
LGTM! Fine with the change in default. Thanks @Irieo for digging into it.

pypsa/optimization/constraints.py
```diff
@@ -596,11 +596,16 @@ def define_ramp_limit_constraints(
...
-
-    # Skip if there are no ramp limits defined or if all are set to 1 (no limit)
-    if (ramp_limit_up.isnull() & ramp_limit_down.isnull()).all():
-        return
-    if (ramp_limit_up == 1).all() and (ramp_limit_down == 1).all():
+    all_null = (ramp_limit_up.isnull() & ramp_limit_down.isnull()).all()
+
+    if not c.committables.empty:
+        ramp_limit_start_up = c.da.ramp_limit_start_up
+        ramp_limit_shut_down = c.da.ramp_limit_shut_down
```
FabianHofmann: we should select by `com_i` here too, right? otherwise all_null and ramp_limit_{start_up/shut_down} have different shapes, leading to broadcast nans

pypsa/optimization/constraints.py
```diff
@@ -778,11 +783,62 @@ def p_previous(idx: pd.Index) -> DataArray:
                 lhs, "<=", rhs, name=f"{c.name}-com-{attr}-ramp_limit_up", mask=mask
             )
 
+        # Special constraints for first snapshot when NOT rolling horizon
+
+        if not is_rolling_horizon and ~ramp_limit_start_up_com.isnull().any():
```
FabianHofmann:
```suggestion
        if not is_rolling_horizon and ramp_limit_start_up_com.notnull().any():
```
for readability

pypsa/optimization/constraints.py
```diff
@@ -778,11 +783,62 @@ def p_previous(idx: pd.Index) -> DataArray:
...
+                    name=f"{c.name}-com-{attr}-ramp_limit_start_up_first",
+                    mask=mask_first,
+                )
+
+        if not is_rolling_horizon and ~ramp_limit_shut_down_com.isnull().any():
+            # Check which units start from ON state (up_time_before > 0)
+            up_time_before = c.da.up_time_before.sel(name=com_i)
+            starts_from_on = up_time_before > 0
+            if starts_from_on.any():
+                # For first snapshot: p(0) >= (p_nom - ramp_limit_shut_down * p_nom)
```
FabianHofmann, line 817: this assumes the component was dispatched at full capacity in the pre-optimization snapshot right? I am not sure if that is a good assumption. it is a bit arbitrary (the model forces you to be). I could imagine it is a likewise assumption that the dispatch was at `p_min_pu` and therefore it is feasible to shut down at the very first snapshot. So I am not sure if we should add this in the first place.

pypsa/data/component_attrs/generators.csv
```diff
@@ -37,8 +37,8 @@ up_time_before,int,snapshots,1,Number of snapshots that the generator was online
 down_time_before,int,snapshots,0,Number of snapshots that the generator was offline before `n.snapshots` start. Only used if `committable=True` and `min_down_time>0`. Does not consider snapshot weightings.,Input (optional)
 ramp_limit_up,static or series,per unit,NaN,"Maximum active power increase from one snapshot to the next, per unit of the nominal power. Ignored if NaN. Does not consider snapshot weightings.",Input (optional)
 ramp_limit_down,static or series,per unit,NaN,"Maximum active power decrease from one snapshot to the next, per unit of the nominal power. Ignored if NaN. Does not consider snapshot weightings.",Input (optional)
-ramp_limit_start_up,float,per unit,1,"Maximum active power increase at start up, per unit of the nominal power. Only used if `committable=True`.",Input (optional)
-ramp_limit_shut_down,float,per unit,1,"Maximum active power decrease at shut down, per unit of the nominal power. Only used if `committable=True`.",Input (optional)
+ramp_limit_start_up,float,per unit,NaN,"Maximum active power increase at start up, per unit of the nominal power. Only used if `committable=True`. Ignored if NaN.",Input (optional)
```
FabianHofmann, line 40: this is a breaking change (not in pypsa functions, but possibly downstream). let's write that out in the release notes

Irieo: Right, it is mentioned in the release notes (I happen to push that simultaneously with the comment) 
upd: I mean release notes mention change in default for the ramp_limit_start_up/down, w/o "breaking"

pypsa/optimization/constraints.py
Irieo: Exactly. This indeed assumes a full capacity in the pre-optimization. I saw arguments for both `p_nom` and `p_min_pu`. I think the latter is less intuitive assumption, although it allows switching off units in the first snapshot for any value exceeding p_min_pu (so any reasonable value)

pypsa/optimization/constraints.py
Irieo: agree!

Irieo:
> deal or no deal? : I would like to proposal how to integrate it into the rolling horizon constraints to have things more side by side. would you mind if I make a pr against yours? or create a separate one, to the see the effective diff? and if I don't do it this week, we pull this pr in

Full 🟢 for making it more compact. So Deal (w/ either of the two ways you suggest)

pypsa/optimization/constraints.py
Irieo: You're right this is logically imprecise (should lock only relevant subset of components). I think combination of things that NaN is now a default (so applies for all non-committable) and that later all_null includes .all() that reduces the array to a single boolean made the overall logic to run through w/o errors

pypsa/data/component_attrs/generators.csv
lkstrp: > This is a breaking change (not in pypsa functions, but possibly downstream). let's write that out in the release notes

@FabianHofmann 
Is it though? Result will be the same, just some none constraining constraints are not going to be written anymore. Which is only a problem if you do custom stuff on top of them? At least that is my understanding and then I am also fine with the rename. Specially since this is first of all fixing a quite big bug. Otherwise my moral compass would probably go on

PR #1552, opened 2026-02-05
Add typical periods agg method; add inter-typical-period storage math

### LINKED ISSUE #1288: Implement Kotzur method for representative days
### Theory

Kotzur et al. (2018)

https://arxiv.org/pdf/1710.07593

### Other Implementations

- Calliope: https://calliope.readthedocs.io/en/stable/user/advanced_features.html#time-resolution-adjustment

- eTraGo in older PyPSA version: https://github.com/openego/eTraGo/blob/bc3447c2c9b2df77bbf3c66168ed67846eca38c7/etrago/tools/constraints.py#L2420-L3473

- FINE: https://vsa-fine.readthedocs.io/en/master/mathematicalDocumentation/storageDoc.html

### Motivation

While hourly consecutive modelling has become the norm in many energy system models, several types of problems may make it necessary to reduce the time dimension. One key example is capacity expansion with multiple investment periods (with piecewise linearised endogenous technological learning). See Zeyen et al. (2023): https://www.nature.com/articles/s41467-023-39397-2

### Tips

Implement simplification in Appendix B. https://github.com/PyPSA/PyPSA/issues/36#issuecomment-704752151

brynpickering:
WIP

This feature touches a lot of places in the code and I am wary of continuing too far down this road before there's been some eyes on it. Namely, there are new attrs attached to the network object and the need to store results in `dynamic` attr stores when they don't reference `snapshot` (they are dynamic, but over different dimensions; more below).

Closes #1288.

## TODO
- Docs and changelog
- integration tests to check optimised results for constraint validity (with and without cyclic storage activated). ~From a brief review, I think there is one constraint incorrectly defined that is overconstraining the problem somehow.~ EDIT: I'm happy with the math now.
- Add typical day weighting to statistics
- Consider adding statistic for a reconstructed full timeseries (namely to visualise the storage reservoir flow after combining all the auxiliary variables used in the math).
- Decide on naming. 
  - I went with `typical_periods` as it is less verbose than `representative_periods` and I used `period` not `day` as you can feasibly define typical `weeks` as well (by clustering to 7 days). `clusters` might be neater but that could cause confusion with the `clustering` module.
  - This is slightly contradicted by my use of `day` as one of the new dimensions. Ideally it would be `period` in this naming convention, but that's already taken.
- ~Add consistency check to stop attempts at clustering when the snapshot weightings don't align for the chosen number of days (e.g. if choosing 7 days, the snapshot weighting profile must be identical across all 7-day blocks in the timeseries).~ TSAM catches this anyway, so I've just passed its error onto the user.
- Consider whether to have the option of a much simpler storage constraint instead of the inter-typical-period constraint. This constraint would set the start and end state of charge within each typical period to equal a fixed value so that you retain consistency across all time periods. As you can imagine, this might work for short-term storage, which tends to cycle on the order of days, but would heavily misrepresent seasonal storage.
- Handle cases of temporal clustering when the network has been optimised. This adds _loads_ of timeseries dfs to the clustering algorithm which effectively kills it. Anything that is a result needs to be excluded (haven't found a way yet to collect these df names).

- Add `typical_periods` method to the temporal clustering module which creates a non-contiguous timeseries of typical periods that represent the full timeseries. Typical periods are always blocks starting from the first day of the year. That is, a typical week will always be a week starting from the same day of the week as the first day of the year.
- Add two new attrs to the `Network` object:
  -  `typical_periods` gives the typical period to which each final snapshot belongs
  - `typical_period_map` gives the typical period to which each period in the original timeseries belongs (e.g. which typical week does each 7-day block belong).
- Add new dimensions (`day`, `typical_period`), decision variables, and constraints to the optimisation problem to track storage levels between periods of the original timeseries. This is a work-around for the non-contiguous nature of the snapshots with this time agg method, developed by [Kotzur et al](https://arxiv.org/pdf/1710.07593).

  
## Checklist

### DIFF
--- a/pypsa/network/io.py
+++ b/pypsa/network/io.py
@@ -163,6 +163,12 @@ class _Exporter(_ImpExper):
         """Save investment periods data."""
 
     @abstractmethod
+    def save_typical_periods(
+        self, typical_periods: pd.Series, typical_period_map: pd.Series
+    ) -> None:
+        """Save typical periods data."""
+
+    @abstractmethod
     def save_scenarios(self, scenarios: pd.DataFrame) -> None:
         """Save scenarios data."""
 
@@ -263,14 +269,21 @@ class _ImporterCSV(_Importer):
             fn, index_col=0, encoding=self.encoding, quotechar=self.quotechar
         )
 
-    def get_typical_periods(self) -> pd.Series:
+    def get_typical_periods(self) -> tuple[pd.Series, pd.Series]:
         """Get typical periods data."""
         fn = self.path.joinpath("typical_periods.csv")
         if not fn.is_file():
-            return None
-        return pd.read_csv(
+            return None, None
+        typical_periods = pd.read_csv(
             fn, index_col=0, encoding=self.encoding, quotechar=self.quotechar
         )
+        fn = self.path.joinpath("typical_period_map.csv")
+        if not fn.is_file():
+            return typical_periods, None
+        typical_period_map = pd.read_csv(
+            fn, index_col=0, encoding=self.encoding, quotechar=self.quotechar
+        )
+        return typical_periods.squeeze(), typical_period_map.squeeze()
 
     def get_static(self, list_name: str) -> pd.DataFrame:
         """Get static components data."""
@@ -360,6 +373,18 @@ class _ExporterCSV(_Exporter):
         fn = self.path.joinpath("investment_periods.csv")
         with fn.open("w"):
             investment_periods.to_csv(
+                fn, encoding=self.encoding, quotechar=self.quotechar
+            )
+
+    def save_typical_periods(
+        self, typical_periods: pd.Series, typical_period_map: pd.Series
+    ) -> None:
+        fn = self.path.joinpath("typical_periods.csv")
+        with fn.open("w"):
+            typical_periods.to_csv(fn, encoding=self.encoding, quotechar=self.quotechar)
+        fn = self.path.joinpath("typical_period_map.csv")
+        with fn.open("w"):
+            typical_period_map.to_csv(
                 fn, encoding=self.encoding, quotechar=self.quotechar
             )
 
@@ -524,16 +549,20 @@ class _ImporterExcel(_Importer):
         else:
             return df
 
-    def get_typical_periods(self) -> pd.Series:
+    def get_typical_periods(self) -> tuple[pd.Series, pd.Series]:
         """Get typical periods data."""
         try:
-            df = self.sheets["typical_periods"]
-            df = df.set_index(df.columns[0])
-            df.index = df.index.astype(int)
+            typical_periods = self.sheets["typical_periods"]
+            typical_periods = typical_periods.set_index(typical_periods.columns[0])
+
+            typical_period_map = self.sheets["typical_period_map"]
+            typical_period_map = typical_period_map.set_index(
+                typical_period_map.columns[0]
+            )
         except (ValueError, KeyError):
-            return None
+            return None, None
         else:
-            return df
+            return typical_periods, typical_period_map
 
     def get_static(self, list_name: str) -> pd.DataFrame:
         """Get static components data."""
@@ -648,6 +677,12 @@ class _ExporterExcel(_Exporter):
         """Save investment periods data."""
         investment_periods.to_excel(self.writer, sheet_name="investment_periods")
 
+    def save_typical_periods(
+        self, typical_periods: pd.Series, typical_period_map: pd.Series
+    ) -> None:
+        typical_periods.to_excel(self.writer, sheet_name="typical_periods")
+        typical_period_map.to_excel(self.writer, sheet_name="typical_period_map")
+
     def save_scenarios(self, scenarios: pd.DataFrame) -> None:
         """Save scenarios data."""
         msg = "Stochastic networks are not supported in the Excel exporter. Use netcdf instead."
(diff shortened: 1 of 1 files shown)


brynpickering:
An interesting feature of pypsa compared to the original implementation is that `Store` components can have a time-dependent `e_min_pu` / `e_max_pu`. It took me a while to work out the most reasonable way to implement this change. It involves accounting for the reduced reservoir availability in `Store_e_intra_period_<lower/upper>`. For the upper bound I believe it should be `- (1 - e_max_pu) * e_nom`. For the lower bound, `+ e_min_pu * e_nom`.

What this depends on is that `max_hours` for `StorageUnit` components will always be static. From what I can see from the source code, there is never really any expectation that it would be time-dependent. But that doesn't stop a user setting it as a dynamic attr... I think it will become quite complex if it is time-dependent since the equivalent of `e_nom` is `max_hours * p_nom`. `max_hours * p_nom * p_max_pu` isn't equivalent to `e_max_pu * e_nom`. You'd need something like `max_hours_<min/max>_pu`.

pypsa/optimization/common.py
```diff
@@ -48,13 +48,16 @@ def _set_dynamic_data(n: Network, component: str, attr: str, df: pd.DataFrame) -
...
     if (attr not in c.dynamic) or (c.dynamic[attr].empty):
-        c.dynamic[attr] = df.reindex(n.snapshots)
+        c.dynamic[attr] = df
 
     else:
         c.dynamic[attr].loc[df.index, df.columns] = df
 
-    # Reindex to match network snapshots and component names
-    result = c.dynamic[attr].reindex(n.snapshots, level="snapshot", axis=0)
+    if df.index.name == "snapshot":
```
FabianHofmann: introduce a `c.has_typical_periods` for these cases?

brynpickering:
I've made some tests to see how this implementation performs, before going to the effort of cleaning up the IO/statistics/attr definitions etc.

TL;DR: we get no benefit from this feature compared to e.g. time segmentation (variable frequency resampling), so I recommend it isn't implemented in PyPSA given how intrusive it will be.

I have tested this in two cases:

##  PyPSA `model.energy` example

This is the same example as used in the temporal aggregation example notebook in the docs), with the following size:

```
Components:
 - Bus: 2
 - Carrier: 9
 - Generator: 3
 - Link: 2
 - Load: 1
 - StorageUnit: 1
 - Store: 1
Snapshots: 8760
```

I aggregated here using `k-means`, choosing the representative day in each cluster by the median.

I will show a few different results. The main conclusion is that using typical days is significantly worse than time segmentation for the same level of aggregation / solve time. There are many more plots I could show. The story is the same across them all.

### Seasonal storage

In this example, we have a seasonal storage device (hydrogen) whose storage level we begin to emulate from about 60/90 typical days (75-85% snapshot dimension reduction), at least in terms of general trend. However, Even at 270 typical days (26% snapshot dimension reduction), we aren't really closely following the original storage level.

image

We cannot emulate it well until we reach 36 typical _weeks_ (i.e., only 30% snapshot dimension reduction):

image

With time segmentation, we do a much better job, even a high level of snapshot aggregation:

image

### Time to solve

Due to the inter-typical storage constraints and auxiliary variables, the time to solve is slightly longer for the same level of aggregation with typical periods compared to time segmentation:

image

### Objective function value

The objective function value is also generally poorly reflected by typical days, although both methods seem to do OK over 4000 snapshots (50% snapshot dimension reduction).

image

### Renewables curtailment

Same story here.

image

## `gb-dispatch-model` network

The [`gb-dispatch-model` workflow](https://github.com/open-energy-transition/gb-dispatch-model) creates dispatch-only networks (all capacities are fixed) with the following size:

```
Components:
 - Bus: 576
 - Carrier: 44
 - Generator: 272
 - Line: 85
 - Link: 911
 - Load: 288
 - StorageUnit: 106
 - Store: 336
 - SubNetwork: 2
Snapshots: 8760
```

So, it's much larger than the example network. The conclusions remain the same. However, I've been unable to cluster with k-means as the problem size proved too much for `tsam` and it seems to hang. It may be that results would improve with k-means clustering, but I doubt it would be by much.

> [!NOTE]
>  I'm showing number of snapshots here, which hides the fact that some are typical weeks and some typical days. For transparency:
>
> - typical day N snapshots: 96, 192, 288, 576, 1752
> - typical week snapshots: 168, 336, 672, 864, 1344, 2016
>
> I have removed any under 365 snapshots when comparing results due to IO issues which lead to stored networks having lost some data below this number.

### Seasonal storage

There are few seasonal storage devices. Here, I'm showing the aggregated hydrogen storage level across all regions in the network.

As in the simple example, typical days do better than typical weeks. However, the error remains high even at a higher number of snapshots, with a major mid-summer peak that isn't in the original network. This would be because the weeks leading up to that peak are all net charging typical days.

Typical weeks:
image

Typical days:
image

Segmentation doesn't work as perfectly as it does in the simple model case, but it is significantly better than typical days for the same level of aggregation.

### Time to solve

image

Again, longer runtimes for the same level of aggregation (those extra constraints and auxiliary variables). We would tend to expect a >Nx improvement in solve time for an Nx snapshot aggregation (e.g., 4x agg -> >4x reduction in solve time). When we undertake a 4x agg here, we get a >4x solve time improvement for segmentation, but only about 3x improvement with typical periods. 

### Objective function value

Generally poor performance on objective function value for typical periods compared to the original and to using segmentation:

image

### Renewables curtailment

You probably expect this by now: typical periods poorly represent renewables curtailment as well. Segmentation begins to perform particularly poorly as well below about a 5x snapshot dimension reduction, but it is always _better_ than typical periods.

image

brynpickering:
Given everything I've stated above, I'm closing this PR.

PR #1553, opened 2026-02-06
ramp limit consolidation and fix

### LINKED ISSUE #592: Links ramp_start_up_limit not behaving as expected

FabianHofmann:
## Changes

requires https://github.com/PyPSA/linopy/pull/568
closes #592 
closes #1540 
supersedes #1540 

This refactor consolidates ramp limit constraint definitions into a unified formulation that handles all operational types (fixed, extendable, committable) with a single constraint. It also introduces `p_init` to enable ramp constraints at the first snapshot.

New Attribute: `p_init` added to generators and links (`pypsa/data/component_attrs/`):

- Specifies the initial active power for ramp limit constraints at the first snapshot
-  When set, ramp constraints are applied at `sns[0]` using `p_init` as the previous dispatch
- For committable**: Initial status is derived from `up_time_before > 0`

### Default Value Changes (Breaking)

`ramp_limit_start_up` and `ramp_limit_shut_down`:
- **Before**: Default `1` (full ramp allowed)
- **After**: Default `NaN` (no limit / ignored)
(like in #1540)

### Consolidated Constraint Formulation

The constraint code (`pypsa/optimization/constraints.py`) was simplified from ~230 lines with separate handling for fix/ext/com to ~75 lines with a unified formulation:

**Ramp Up:**
```
p[t] - p[t-1] <= ru * p_nom * u[t-1] + rusu * p_nom * (u[t] - u[t-1])
```

**Ramp Down:**
```
p[t] - p[t-1] >= -rd * p_nom * u[t] - rdsd * p_nom * (u[t-1] - u[t])
```

Key insight: For non-committable components, `status = 1` for all `t`, so start-up/shut-down terms vanish naturally. For extendable components, `p_nom` is replaced by the capacity variable.

### Constraint Naming (Breaking)

- **Before**: `{c}-fix-p-ramp_limit_*`, `{c}-ext-p-ramp_limit_*`, `{c}-com-p-ramp_limit_*`
- **After**: `{c}-p-ramp_limit_up`, `{c}-p-ramp_limit_down`

### First Snapshot Handling

| Scenario | Previous Dispatch | Previous Status |
|----------|-------------------|-----------------|
| Rolling horizon (`sns[0] != n.snapshots[0]`) | `c.dynamic[hist_attr][start_i]` | `c.dynamic.status[start_i]` |
| First snapshot with `p_init` set | `p_init * initially_up` | `up_time_before > 0` |
| First snapshot without `p_init` | No constraint at `sns[0]` | - |

# Merge Order

This should go in before #1007 because there the ramp limits not yet fully sorted out. 

## Checklist

### DIFF
--- a/docs/assets/javascripts/readthedocs.js
+++ b/docs/assets/javascripts/readthedocs.js
@@ -25,4 +25,5 @@ ${ config.versions.active.map(
 </div>`;
 
     document.querySelector(".md-header__topic").insertAdjacentHTML("beforeend", versioning);
-    });
+    },
+    { once: true });
--- a/docs/user-guide/warnings.md
+++ b/docs/user-guide/warnings.md
@@ -1,0 +1,55 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Warnings Reference
+
+PyPSA emits warnings to help catch common mistakes when building networks.
+This page explains what they mean and how to resolve them.
+
+!!! warning
+
+  This page is a work in progress and not all warnings are documented yet.
+
+## Misleading attribute name { #warning-attr-misleading }
+
+> *"The attribute 'X' is a standard attribute for other components but not for Y.
+> This could cause confusion and it should be renamed."*
+
+This warning appears when you add a custom attribute to a component that happens
+to be a standard attribute of a different component type. For example, adding
+`s_nom` (a standard Line/Transformer attribute) to a Generator:
+
+```py
+n.add("Generator", "gen1", bus="bus1", s_nom=100)  # triggers warning
+```
+
+This usually indicates that the attribute was expected to exist for this component
+but it does not. Additional attributes can be added to components, but they are
+only descriptive and not used by PyPSA internally. Consider renaming it to avoid
+confusion with the standard attribute.
+
+!!! note
+    This warning cannot be suppressed via options.
+
+## Potential attribute typo { #warning-attr-typo }
+
+> *"The attribute 'X' is not a standard attribute for Y. Did you mean 'Z'?"*
+
+This warning appears when a custom attribute name is very similar to a standard
+attribute (edit distance of 1), suggesting a typo. For example:
+
+```python
+n.add("Generator", "gen1", bus="bus1", p_no=100)  # did you mean 'p_nom'?
+```
+
+This warning can be suppressed if you intentionally use non-standard attribute
+names that happen to be close to standard ones:
+
+```python
+pypsa.options.warnings.attribute_typos = False
+```
+
+See also [Warnings options](./options.md#warnings-options).
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -85,6 +85,7 @@ nav:
     - Network Collection: user-guide/collection.md
     - Options: user-guide/options.md
     - Logging: user-guide/logging.ipynb
+    - Warnings: user-guide/warnings.md
 
 - Examples:
   - Overview: examples/examples.md
--- a/pypsa/optimization/mga.py
+++ b/pypsa/optimization/mga.py
@@ -306,7 +306,7 @@ class OptimizationAbstractMGAMixin:
         model_kwargs : dict, optional
             Keyword arguments used by `linopy.Model`, such as `solver_dir` or `chunk`.
             Defaults to module wide option (default: {}). See
-            `https://`go.pypsa.org/options-params` for more information.
+            `https://go.pypsa.org/options-params` for more information.
         **kwargs:
             Keyword argument used by `linopy.Model.solve`, such as `solver_name`,
 
@@ -456,7 +456,7 @@ class OptimizationAbstractMGAMixin:
         model_kwargs : dict, optional
             Keyword arguments used by `linopy.Model`, such as `solver_dir` or `chunk`.
             Defaults to module wide option (default: {}). See
-            `https://`go.pypsa.org/options-params` for more information.
+            `https://go.pypsa.org/options-params` for more information.
         **kwargs:
             Keyword argument used by `linopy.Model.solve`, such as `solver_name`,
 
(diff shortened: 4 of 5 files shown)


Irieo:
The plan is to close #1540 or to merge it before this PR?
The release notes here repeat information about new defaults for `ramp_limit_start_up` and `ramp_limit_shut_down` but do not mention the main reason #1540 was there is to fix the bug that these attributes were ignored in binary UC.

FabianHofmann:
> The plan is to close #1540 or to merge it before this PR? The release notes here repeat information about new defaults for `ramp_limit_start_up` and `ramp_limit_shut_down` but do not mention the main reason #1540 was there is to fix the bug that these attributes were ignored in binary UC.

if you don't mind, we could give this one prio and close #1540 ; I have to release a new linopy version first, though

Irieo:
> > The plan is to close #1540 or to merge it before this PR? The release notes here repeat information about new defaults for `ramp_limit_start_up` and `ramp_limit_shut_down` but do not mention the main reason #1540 was there is to fix the bug that these attributes were ignored in binary UC.
> 
> if you don't mind, we could give this one prio and close #1540 ; I have to release a new linopy version first, though

No worries. I thought we wanted merge 1553 after 1540 due to different logic (direct fix vs code refac + new att), but if you think it is better so you can just close #1540. Maybe just pull my unit tests.

docs/release-notes.md
```diff
@@ -47,10 +47,13 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - In version 2.0, capital costs of existing capacity on extendable assets will no longer be included in the objective by default (`n.objective_constant` will be set to zero), which improves LP numerical conditioning. A new `include_objective_constant` parameter was added to [`n.optimize()`][pypsa.optimization.OptimizationAccessor.__call__] and [`n.optimize.create_model()`][pypsa.optimization.OptimizationAccessor.create_model] to allow controlling this behavior and opt-in to the new default. It can also be configured via `pypsa.options.params.optimize.include_objective_constant` (see <!-- md:guide options.md -->). (<!-- md:pr 1509 -->)
 
-- Add support for pandas 3.0, while maintaining compatibility with pandas 2.x. (<!-- md:pr 1556 -->)
```
lkstrp, line 50: Overwrite of existing release note

pypsa/optimization/constraints.py
```diff
@@ -565,228 +565,84 @@ def define_ramp_limit_constraints(
...
-    com_i = c.committables.difference(c.inactive_assets)
-    fix_i = c.fixed.difference(c.inactive_assets)
-    fix_i = fix_i.difference(com_i).rename(fix_i.name)
-    ext_i = c.extendables.difference(c.inactive_assets)
-
-    # Auxiliary variables for constraint application
-    ext_dim = ext_i.name if ext_i.name else c.name
-    original_ext_i = ext_i.copy()
-    original_com_i = com_i.copy()
+    status = DataArray(1, coords=[sns, idx])
```
lkstrp, line 608: I think we should get rid of all the extra DataArray initialisations. Can we not do something like below?
```suggestion
    status = c.da.status.sel(snapshot=sns).where(~is_com, 0)
```
and remove where in the if statement

pypsa/optimization/constraints.py
```diff
@@ -565,228 +565,84 @@ def define_ramp_limit_constraints(
...
-        non_null_down = ~ramp_limit_down_fix.isnull()
-        if non_null_down.any():
-            lhs = p_actual(fix_i) - p_previous(fix_i)
-            rhs = (-ramp_limit_down_fix * p_nom) + rhs_start
-            mask = active & non_null_down
-            m.add_constraints(
-                lhs, ">=", rhs, name=f"{c.name}-fix-{attr}-ramp_limit_down", mask=mask
-            )
+        initially_up = c.da.up_time_before > 0
+        p_init = c.da.p_init.where(initially_up, 0)
```
lkstrp, line 620: This is replacing any `p_init` since passing it wouldn't make sense when `up_time_before` is 0?
Which makes sense, but can we not raise an error instead? We wanna move to strict user feedback

Or just mark it for now


```suggestion
        p_init = c.da.p_init.where(initially_up, 0) # TODO add strict validation check
```

pypsa/optimization/constraints.py
FabianHofmann: this does not really make sense, given that status is only a output of the optimization. with that we should use the status as a setter for the initial condition for non-rolling horizon optimizations. there however `up/down_time_before` are the important variables.

pypsa/optimization/constraints.py
FabianHofmann: good point, am an adding a consistency check that raises a warning when non nan p_init +  zero update_time_before is given.

PR #1558, opened 2026-02-11
Smspp optimization

AlessandroPampado99:
Closes # (if applicable).

Added optimization with SMS++ (using pypsa2smspp and pySMSpp) with n.smspp

- pypsa2smspp and pySMSpp are not mandatory
- test included
- n.smspp can be used without any argument (default pypsa2smspp config file)

## Checklist

- [X] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `docs`.


pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,88 @@
...
+        Returns
+        -------
+        Network
+            The network returned by the transformation pipeline.
+
+        """
+        _require_smspp_deps()
+
+        # Lazy import without violating Ruff PLC0415 (no import statement inside function)
+        module = import_module("pypsa2smspp")
```
coroa: I'd prefer to do, which is a lot more explicit, the main use case for import_module is when you have variable `pkg = pypsa2smspp` that you want to import.
```
from pypsa2smspp import Transformation   # noqa: PLC0415
```

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,88 @@
...
+        Transformation = module.Transformation
+
+        n_in = self._n if network is None else network
+        cfg_path = str(config)
+
+        logger.info("Running SMS++ pipeline with config: %s", cfg_path)
+        tr = Transformation(cfg_path)
+        return tr.run(n_in, verbose=verbose)
```
coroa: Does this also sync results back into the network? Sorry, i'm quite clueless about pypsa2smspp.

coroa (changes requested):

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,89 @@
...
+
+        """
+        _require_smspp_deps()
+
+        module = import_module("pypsa2smspp")
+        Transformation = module.Transformation
+
+        n_in = self._n if network is None else network
+
+        # Do not coerce to str: Transformation now supports None and possibly dict overrides
```
coroa: Historic argument -> Remove

pyproject.toml
```diff
@@ -56,6 +56,7 @@ Documentation = "https://docs.pypsa.org"
 
 [project.optional-dependencies]
 hdf5 = ["tables"]
+smspp = ["pypsa2smspp", "pySMSpp"]
```
davide-f:
```suggestion
smspp = ["pypsa2smspp", "pysmspp"]
```

davide-f: FYI smspp-project is now available in conda-forge for linux and macos, however it is not available as pip installation.

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,89 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from pypsa import Network
+
+logger = logging.getLogger(__name__)
+
+
+def _require_smspp_deps() -> None:
```
davide-f, line 26: let's add a check if sms++ umbrella is available.
It is possible to check it by ensuring that the system call "os.system("ucblock_solver --help")" doesn't return an error or we can rely on pysmspp to do so.
I'm adding an issue to pysmspp to add an utility on that.

I'm testing a pysmspp.is_smspp_installed() to facilitate that

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if missing:
+        raise ImportError(
+            "SMS++ backend requires optional dependencies. Missing: "
+            + ", ".join(missing)
+            + ". Install with: pip install 'pypsa[smspp]'"
+        )
+
+
+class SMSppAccessor:
```
davide-f, line 43: To facilitate the usability with existing workflows, I'd suggest to link the [OptimizationAccessor](https://github.com/PyPSA/PyPSA/blob/f036060e9b152f5c0f3bc6bd2a39af789c0b8a66/pypsa/optimization/optimize.py#L408) with this accessor with a simple if like:
```python
if solver_name=='smspp':
   return self.smspp(...)
```
This would allow a smoothless integration with PyPSA-X models, as it would imply to specify a different solver and (optionally) custom solver_options.

If we go for this, we need to ensure the __call__ of smspp returns the same output as optimize

That would allows to explicitly specify warnings of unsupported options that the user may have specified.
Looking forward to hear your perspectives @coroa , @Irieo, @lkstrp

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, n: Network) -> None:
+        """Initialize the accessor with a bound network."""
+        self._n = n
+
+    def __call__(
+        self,
+        config: str | Path | dict[str, Any] | None = None,
+        verbose: bool = False,
+        network: Network | None = None,
```
davide-f: this option is not needed as this accessor would be used only on the current instance

pypsa/optimization/smspp.py
davide-f: Yes it returns a Network with populated results; duals are not yet attached.
Some duals (e.g. those of the energy balance) are planned to be attached

davide-f (commented):
Many thanks Alessandro :)
Sharing comments as discussed

pypsa/optimization/smspp.py
davide-f: Update: this is supported

pypsa/optimization/smspp.py
fneum: I like the idea. Is `smspp` to be equated with a solver though? Ultimately, there should just be one user-facing way to run SMS++. Either `n.optimize.smspp()` or `n.optimize(solver_name='smspp')`, don't like `n.smspp()` as it's too unclear.

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,89 @@
...
+        Returns
+        -------
+        Network
+            The network returned by the transformation pipeline.
+
+        """
+        _require_smspp_deps()
+
+        module = import_module("pypsa2smspp")
+        Transformation = module.Transformation
```
fneum: why not just

```py
import pypsa2smspp
tr = pypsa2smspp.Transformation(config)
```

test/test_smspp.py
fneum, line 1: There should definitely be a couple of tests where `n.optimize()` is compared against `n.optimize(solver_name='smspp')`, demonstrating they produce same results.

It would also be nice to have an example notebook demonstrating the benefit of using SMS++.

fneum (changes requested):

davide-f:
As discussed with Alessandro, we'll be working on this PR together

pypsa/optimization/smspp.py
davide-f: Done

pypsa/optimization/smspp.py
davide-f: Agree, revised

pyproject.toml
davide-f: Done

pypsa/optimization/smspp.py
davide-f: Done

pypsa/optimization/smspp.py
davide-f: I agree. I'm in favor of `n.optimize(solver_name='smspp')` as it allows smoothless interaction with existing workflows, with appropriate design.

I'd like a design such that a user may do alternatively:

`n.optimize(solver_name='smspp')`

or something like

```
n.optimize.smspp.create_model(...)
n.optimize.smspp.optimize(...)
n.optimize.smspp.retrieve_solution(...)
```
that under the hood calls the corresponding function of pypsa2smspp.
See example here:
https://pypsa2smspp.readthedocs.io/en/latest/examples/conversion_step_by_step/

The three-step approach allows higher customization, namely, a user may edit the SMS++ object created in create_model before solving, or using it directly.

What do you think?

pypsa/optimization/smspp.py
davide-f: network is dropped; need to work on the description of the configs

pypsa/optimization/smspp.py
davide-f: Done

test/test_smspp.py
davide-f: Absolutely, the idea is to create something like this:
https://pypsa2smspp.readthedocs.io/en/latest/examples/capacity_expansion_2bus/

we have a prototype also for stochastic optimization, though it requires smspp-project >=0.6.0 that is not yet available [upcoming]
https://pypsa2smspp--38.org.readthedocs.build/en/38/examples/stochastic_demand_intermittent/

If you wish to have it solved in the documentation, we need to include also the conda package of smspp-project and python 3.12+.
Same for testing.
Shall we include a separate workflow for testing this or do shall I include it in the current test?

davide-f:
The smspp test is currently tested with a separate CI not to disturb the current CI approach.
See https://github.com/AlessandroPampado99/PyPSA/pull/1 the test

The current CI relies on the release in conda, alternatively, it can be recasted to install from upstream.
My recommendation is to use conda for stability purposes.

Would you prefer to have an unique CI or separate? Please advise on your preference
I believe having everything in an unique file would be preferred. Before proceeding, I'd like feedback.

Currently, no documentation is added yet: focus is on functionality.
For the documentation, I'd envision a couple of example taken from [here](https://pypsa2smspp.readthedocs.io/en/latest/examples/capacity_expansion_2bus/).

.github/workflows/test-smspp.yml
```diff
@@ -0,0 +1,68 @@
...
+      run:
+        shell: bash -el {0}
+    env:
+      MPLBACKEND: Agg  # https://github.com/orgs/community/discussions/26434
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0 # Needed for setuptools_scm
+
+    - name: Setup Miniforge
```
lkstrp: Can we use pixi here? We kind of want to migrate everything to pixi, see pypsa-eur and the model tests in here.

Also, since this can also be installed via pypi, I would rather stick to uv here and install all os dependencies within the CI. If pypi installs are not stable, we shouldn't emphasise them

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,167 @@
...
+        import pysmspp  # noqa: PLC0415
+    except ImportError as err:
+        raise ImportError(
+            "SMS++ backend requires optional dependencies (pypsa2smspp and pysmspp)."
+            + " Install with: pip install 'pypsa[smspp]'"
+        ) from err
+
+    if not pysmspp.is_smspp_installed():
+        logger.warning(
+            "pySMSpp detects that SMS++ is not installed. SMS++ pipeline will not work."
```
lkstrp, line 39:
```suggestion
            "pySMSpp detects that SMS++ is not installed correctly. SMS++ pipeline will not work."
```

This should not be a warning but a `RunTimeError`

pypsa/optimization/smspp.py
lkstrp: I would kick `n.smspp` completely. Then keep `n.optimize(solver_name='smspp')` as a little shortcut and alias. As you have it already.

And then what you propose but mirror what we already have:
```
n.optimize(...)                                      # simple access to create + solve
n.optimize.create_model(...)               # linopy only
n.optimize.solve_model(...)                 # linopy only
n.optimize.smspp.create_model(...)    # smspp only
n.optimize.smspp.solve_model(...)      # smspp only
```
We already have a `retrieve_solution` equivalent, which is called `assign_solution`. But it's internal and retrieve solution should be as well.

You could argue now that `n.optimize.smspp()` should also be possible and equivalent to `n.optimize(solver_name='smspp')`. But I think I would rather avoid that. Or just do `n.optimize.smspp()` and kick it from the shortcut `n.optimize(..)`. But not sure.

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,167 @@
...
+            solver_options,
+            kwargs,
+        )
+        self.transformation = pypsa2smspp.Transformation(**transformation_kwargs)
+        self.result = None
+
+        self.sms_network = self.transformation.create_model(self._n, verbose=verbose)
+        return self.sms_network
+
+    def optimize(self, verbose: bool = False) -> SMSPPSolverTool:
```
lkstrp: See comment above, rename to `solve_model`

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,167 @@
...
+        """Optimize the SMS++ model created by :meth:`create_model`."""
+        if self.transformation is None:
+            msg = "Call `create_model` before `optimize`."
```
lkstrp: I agree that this is the better pattern. We should align the linopy call with `v2.0`. But if we make this IO compatible, you could also resolve, without creating it again.
```suggestion
            msg = "No SMS++ transformation is available. Call `n.optimize.smspp.create_model()` first."
```

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,167 @@
...
+            raise ValueError(msg)
+
+        self.result = self.transformation.optimize(verbose=verbose)
+        return self.result
+
+    def retrieve_solution(self, verbose: bool = False) -> tuple[str, str]:
```
lkstrp: See comment above, integrate into `solve_model` and make internal.

pypsa/optimization/smspp.py
```diff
@@ -0,0 +1,167 @@
...
+        """Map a pypsa2smspp result object to PyPSA's status tuple."""
+        if isinstance(result, tuple) and len(result) == 2:
+            return str(result[0]), str(result[1])
+
+        status = getattr(result, "status", None)
+        if status is None:
+            return "ok", "optimal"
+
+        condition = str(status)
+        if "success" in condition.lower():
```
lkstrp: Isn't there a more stable way to pass smspp conditions?

pypsa/networks.py
```diff
@@ -186,6 +187,10 @@ def __init__(
         """
         Network [plotting functionality][pypsa.plot.PlotAccessor] accessor.
         """
+        self.smspp: SMSppAccessor = SMSppAccessor(self)
```
lkstrp: That is a bug, since you initialize the SMSppAccessor now twice. That one should be removed.

See the other comments, I think we only wanna have `n.optimize.smspp..`

lkstrp (changes requested):
Hey @davide-f,

I have a couple of comments and an answer to your question in the comments as well!

pypsa/optimization/smspp.py
davide-f: I thought about that and decided to have a warning here to let the user be free to create a sms network if desired; while an error would be through if the .optimize is executed.
I'll check and update here

pypsa/networks.py
davide-f: Absolutely, it was an old refuse. Removed

pypsa/optimization/smspp.py
davide-f: Done

pypsa/optimization/smspp.py
davide-f: In pysmspp we log the output log of SMS++ that include also the condition.
I'd like to add a better "Result" object in pysmspp, not there yet, though I believe we need a converter between the standard linopy/pypsa result and SMS++.

Do you have other suggestions? I can make it a non-static method too.

pypsa/optimization/smspp.py
davide-f: Implemented as suggested

pypsa/optimization/smspp.py
davide-f: Done

davide-f:
I answered the comments and included an example.
If you prefer to skip documentation for the moment I can revert the changes

PR #1567, opened 2026-02-13
clustering api change and better docs

### LINKED ISSUE #1393: Improve function signatures in API docs of cluster/plot modules
### Location

API docs

### Documentation problem

Image

Function signatures empty.

lkstrp:
Closes #1393

- Moves all spatial clustering functions from `n.clustering.*` to `n.clustering.spatial.*`, next to `n.clustering.temporal.*`.
- Uses same Mixin architecture as already done everywhere else for `v1`. This yields better docs and typing


fneum:
Good to go when docs build is green.

PR #1569, opened 2026-02-17
Add link delay for output ports

FabianHofmann:

Adds configurable time delays for Link output ports. Energy withdrawn from `bus0` at snapshot `t` arrives at output ports after a delay measured in `snapshot_weightings.generators` (best choice I could make wo introducing a new field) units. This enables modeling transport delays (e.g. gas pipelines, shipping).

**New Link attributes:**
- `delay` / `delay2`, `delay3`, ... — delay in generator weighting time units per port
- `cyclic_delay` / `cyclic_delay2`, `cyclic_delay3`, ... — whether delayed energy wraps cyclically or is lost at boundaries

**Implementation:**
- `Links.get_delay_source_indexer()` computes per-snapshot source mappings using cumulative generator weightings, supporting both cyclic and non-cyclic modes
- `Links.split_by_port_delay()` groups links by delay/cyclicity for efficient constraint generation
- Nodal balance constraints apply the delay
- Post-solve result extraction applies matching shifts to `p1`, `p2`, ... DataFrames
- `check_link_delays` consistency check validates delay values are non-negative and within the snapshot horizon

**Bug fix:** Link-port attribute expansion (`bus2`, `bus3`, `efficiency2`, ...) no longer mutates shared Link defaults across Network instances. A targeted test was added for this.  

## Checklist


docs/user-guide/components/links.md
```diff
@@ -50,4 +50,33 @@ plant).
 
     Suppose a link representing a methanation process takes as inputs one unit of hydrogen and 0.5 units of carbon dioxide, and gives as outputs 0.8 units of methane and 0.2 units of heat. Then `bus0` connects to hydrogen, `bus1` connects to carbon dioxide with `efficiency=-0.5` (since 0.5 units of carbon dioxide is taken for each unit of hydrogen), `bus2` connects to methane with `efficiency2=0.8` and `bus3` to heat with `efficiency3=0.2`. [This example](../../examples/biomass-synthetic-fuels-carbon-management.ipynb) illustrates many modelling processes with multiple inputs and outputs using links.
 
+## Time-Delayed Energy Delivery
+
+Links can model time-delayed energy transport via the `delay` and `cyclic_delay` attributes. This is useful for representing transport delays in gas pipelines, hydrogen shipping, long-distance HVDC transmission, or any process where energy withdrawn at `bus0` arrives at output ports after a configurable time lag.
```
euronion:
```suggestion
Links can model time-delayed energy transport via the `delay` and `cyclic_delay` attributes. This is useful for representing transport delays in pipelines, shipping and other transport modes, or any process where energy withdrawn at `bus0` arrives at output ports after a configurable time lag.
```

Not sure why HVDC was mentioned, I think the electrons in those lines are also pretty fast.

docs/user-guide/components/links.md
```diff
@@ -50,4 +50,33 @@ plant).
...
+
+- **`delay`**: The delay in units of elapsed time, referencing `n.snapshot_weightings.generators`. Energy withdrawn from `bus0` at snapshot `t` arrives at `bus1` at `t + delay`. That is, for uniform hourly snapshots with unit weightings, `delay=3` means a 3-hour delay. In case of snapshot weightings greater than 1, the energy arrives at `bus1` the first snapshot at whose time step at least `delay` units later.
```
euronion: a) I don't understand the second part about snapshot weightings > 1
b) What is we have snapshots with weighting = 0? Are they skipped?
c) What if `port0` is activated in snapshot `t` with a `delay=1`, and `t+1` has a different weighting. How are e.g. `marginal_costs` or `efficiency` affected? Assume efficiency is snapshot dependent? Would that give an incentive to operate the link at `port0` in `t`? Not sure if that is fine or not.
d) In many instances we have started to allow `pd.Series` instead of scalar values for properties. Would it be possible or meaningful to do the same here? I.e. allow for a Series of delays, specifying different delays during different snapshots.
e) If the users changes their snapshots, they will also need to adjust the delay, right? I think it should fail rather than trying to automatically fix it and aligning it with a guessed snapshot.

docs/user-guide/optimization/energy-balance.md
```diff
@@ -27,6 +27,12 @@ where the **decision variables** are represented by:
...
 bus $n$. If $p_{l,t}>0$ it withdraws from the starting bus. The time-varying incidence matrix $L_{n,l,t}$ takes non-zero values $-1$ if the link $l$ starts at bus $n$ and efficiency $\eta_{n,l,t}$ if it ends at bus $n$. If $f_{l,t}>0$ it withdraws from `bus0` and feeds in $\eta_{n,l,t} f_{l,t}$ to `bus1`. For a link with more than two outputs (e.g. a combined heat and power plant), the incidence matrix $L_{n,l,t}$ has more than two non-zero entries with efficiencies $\eta_{n,l,t}$ for `bus2`, `bus3`, etc.. The entries may also be negative to denote additional inputs rather than multiple outputs.
 
+When a link has a non-zero `delay` attribute, the output flow at snapshot $t$ corresponds to the input at an earlier source snapshot $s(t)$ rather than at $t$ itself. The source snapshot $s(t)$ is the latest snapshot such that $\tau(s) \leq \tau(t) - \delta$, where $\tau$ denotes the cumulative start time derived from `snapshot_weightings.generators` and $\delta$ is the delay value. The delayed contribution to the energy balance at output bus $n$ becomes:
```
euronion, line 30: I think we should enforce that the delay ends on an existing snapshot, not matching it to the next snapshot. The user should explicitly need to handle this.

pypsa/components/_types/links.py
```diff
@@ -47,6 +47,13 @@ class Links(Components):
 
     _operational_variables = ["p"]
 
+    class DelayGroup(NamedTuple):
```
lkstrp: This should be internal and not nested in the class. And if further look at it, its not even necessary, is it?

pypsa/components/_types/links.py
```diff
@@ -116,3 +123,97 @@ def additional_ports(self) -> list[str]:
             for col in self.static.columns
             if (match := RE_PORTS_GE_2.search(col))
         ]
+
+    def split_by_port_delay(
+        self, port: str, *, subset: pd.Index | None = None
+    ) -> tuple[pd.Index, list[DelayGroup]]:
+        """Split links of a port into non-delayed assets and delayed groups.
```
lkstrp: The method shouldn't be exposed and I think this is not even a good class method and I would move this to the optimize code. It's only a helper there and wouldn't called otherwise and it also doesn't really tell you something meaningful on the Linksf

pypsa/components/_types/links.py
```diff
@@ -116,3 +123,97 @@ def additional_ports(self) -> list[str]:
...
+
+        Parameters
+        ----------
+        port : str
+            Port id like ``"1"``, ``"2"``, ...
```
lkstrp: This is no markdown formatting

pypsa/components/_types/links.py
```diff
@@ -116,3 +123,97 @@ def additional_ports(self) -> list[str]:
...
+        subset : pd.Index, optional
+            Optional subset of link names to process.
+
+        Returns
+        -------
+        tuple[pd.Index, list[DelayGroup]]
+            First element contains non-delayed link names. Second element contains
+            delay-groups, where each group has common ``delay`` and
+            ``is_cyclic`` values.
```
lkstrp: markdown

pypsa/optimization/constraints.py
```diff
@@ -942,7 +943,7 @@ def define_nodal_balance_constraints(
     if buses is None:
         buses = n.c.buses.static.index.unique("name")
 
-    links = as_components(n, "Link")
+    links = cast("Links", as_components(n, "Link"))
```
lkstrp: See comment above. This is not clean since only needed because we use a subclass specific method via general interface. Should rather be restructured instead of casted

Also as_components you can kick for internals wherever you see it

pypsa/optimization/constraints.py
```diff
@@ -954,14 +955,45 @@ def define_nodal_balance_constraints(
...
+            eff_attr = f"efficiency{i_suffix}"
             eff = links.da[eff_attr].sel(snapshot=sns)
-            args.append(["Link", "p", f"bus{i}", eff])
+
+            non_delayed, delayed_groups = links.split_by_port_delay(i)
+            if isinstance(non_delayed, pd.MultiIndex):
+                non_delayed = non_delayed.get_level_values("name").unique()
+            non_delayed = non_delayed.intersection(active)
+            if not non_delayed.empty:
+                args.append(
```
lkstrp: I think this can be simplified, can I just dug in here ? Just give me a go so we don't have merge conflicts

docs/user-guide/components/links.md
FabianHofmann: they definitely do

docs/user-guide/components/links.md
FabianHofmann: a) there is a "as" too much, I will add an example to clearify. 
b) weighting of 0? interesting use case. they are not skip, the would just not contribute to the cumulative sum of the snapshot weightings  when comparing delays and time elapsed. if you want to have that case bullet proof I can add a test for that. It could be that atm the a delay of 1 and a weighting of 0 would lead to energy traveling back in time. 
c) marginal_cost is always referencing `p0` so this is implicitly handled by the optimization, efficiency is only touching outputs `p1+`. if the efficiency is time-dependent, it will reference the efficiency at the arrival time, ie. the target snapshot. after thinking about it, I found this cleaner then the other way round, but open to change if this seems counter-intuitive.
d) honestly I wouldn't do it. I would be quite a complex thing. the cumulative comparison is already complex and this would make it even more complicated to review later. 
e)  do you mean if the user change the snapshots or their weightings? Since the delay takes the time elapsed into account, changing the snapshot alongside with the snapshot weighting is no problem. suppose delay is 3 and snapshot resolution is hourly (sn weights = 1) . now you are resampling to 3 hourly resolution (adjusting the sn weights to 3), everything works out of the box without touching the delay.

FabianHofmann:
> I love seeing this feature.
> 
> Left some comments without looking at it in more detail.
> 
> Another concern: How does this combine with the statistics module, especially if `aggregate_time=False`. Do we then need to specify `at_port` to get the right results for time-series, as they will be different depending on which port is selected.

thanks for the review @euronion. before we make the decision of killing the flexibility of having `delay != sn weight` or rather `delay != sn weight * n`, I would like to advocate for the convention I introduced dealing with these cases (last two sections) in link-delay.ipynb

run it with
``` 
gh pr checkout 1569
uv run jupyter notebook docs/examples/link-delay.ipynb
```

or have a look at the rdt built in a minute when it (hopefully) runs through


About the stastics module. that is working right away as it uses the already delayed dispatch time series at `p{port}`

lkstrp:
> before we make the decision of killing the flexibility of having delay != sn weight or rather delay != sn weight * n, I would like to advocate for the convention I introduced dealing with these cases (last two sections) in link-delay.ipynb

I am not too sure about the use cases. The convention looks clean and is also explained in a good way, but is it just convenience and it would always be better to make delays match with weights? I feel that this can be hard to investigate/ understand. If we keep the convention we could add warnings that some delay rounding is going on

FabianHofmann:
@lkstrp should we then introduce a consistency check? I still think `delay != weights * n` is valid as long as you think about snapshots with weights != 1 as time spans.

lkstrp:
I changed a couple of things and added multi horizon support (raised by @Irieo). @FabianHofmann If you could have another look?

Otherwise two more things:
- Should we also disallow negative delays? For new attributes I wanna be as strict as possible so we can have quick great user feedback with the validation layer without breaking things
- What about #1333 ? Do we wanna support delays there as well?

euronion:
> Otherwise two more things:
> 
> * Should we also disallow negative delays? For new attributes I wanna be as strict as possible so we can have quick great user feedback with the validation layer without breaking things
> * What about [Add processes components as multi-link alternative #1333](https://github.com/PyPSA/PyPSA/pull/1333) ? Do we wanna support delays there as well?

1. I think it would be good to allow for negative delays. That allows users to flexibly use the different buses as input and output. If we only allow for positive delays, then the user needs to start manually preprocessing data again.
2. Yes!

Related to the convention / automatic matching:
Consistency check + automatic matching would be an ok compromise for me.
I think the documentation is clear, although I don't look forward debugging this feature with non-uniform weightings. ;)

But I would be keen on hearing the comments from others.

FabianHofmann:
1. I also tend to keep it strictly zero/positive for now to avoid confusion in the models. formulation is already quite complex. but if you are saying this will be really useful we can ship that later @euronion?    
2. yes, I would definitely expand this to processes. but let's do it one by one though.  completely separate pr

@lkstrp fine with pulling in now?

lkstrp:
> I also tend to keep it strictly zero/positive for now to avoid confusion in the models. formulation is already quite complex. but if you are saying this will be really useful we can ship that later @euronion?

This would still need to be checked then

FabianHofmann:
> > I also tend to keep it strictly zero/positive for now to avoid confusion in the models. formulation is already quite complex. but if you are saying this will be really useful we can ship that later @euronion?
> 
> This would still need to be checked then

I it already checked by the consistency check but only warns. should we make that particular check always strict, ie. always raising? it is currently inside a function with another check, so setting the `strict` default to True would also affect the other check

PR #1570, opened 2026-02-18
Add example: market power model

Irieo:
Adds example with theoretical background and PyPSA implementation of Cournot-Nash oligopoly equilibrium using the "fictitious objective" approach. The method stems from 1990s operations research literature (see [Slade 1994 seminal paper](https://www.jstor.org/stable/2950588)) and was explored in the energy context by [Hobbs 2001](https://hobbsgroup.johnshopkins.edu/docs/papers/IEEE_Hobbs_2001_LCP00918286.pdf), but perhaps remains under-appreciated in modern energy system modeling. It bypasses formulating KKT conditions with complementarity constraints, instead modeling strategic behavior (imperfect competition) by appending a quadratic markup term to PyPSA's cost-minimization objective.

## Checklist


docs/examples/imperfect-competition.ipynb
```diff
@@ -0,0 +1,665 @@
...
+   "id": "01ca0016",
+   "metadata": {},
+   "source": [
+    "# Imperfect Competition in Energy Markets with PyPSA\n",
+    "\n",
+    "This example demonstrates how to model oligopolistic behaviour in energy markets using PyPSA. The notebook provides some theoretical background and a quick neat PyPSA implementation.\n",
+    "\n",
+    "Spoiler: we implemented Cournot-Nash equilibrium without KKT conditions, complementarity solvers, big-M constraints, or any structural changes to cost-minimization problem instance created by PyPSA. Instead, we use a trick known as the \"fictitious objective\" in the economics/OR literature, perhaps underappreciated in energy system modelling stream.\n",
+    "\n",
+    "*IR, Feb 2026*"
```
lkstrp:
```suggestion
```

We don't have those contributions anywhere else so I would remove it. We can list authors based on git automatically below, wanted to do that anyway at some point

lkstrp (approved):
My imperfect brain thinks it's beautiful

docs/examples/imperfect-competition.ipynb
Irieo: Sure

fneum (approved):
If you shorten the name of the example to "Imperfect Competition" it's perfect.

FabianHofmann:
@Irieo this is such a wonderful notebook and explanation! I learned a lot and after all this math I need a coffee :D

Here are some notes, that you might want to consider. but no blocker at all. feel free the merge from my side:

- add abbreviations for LCP and KKT
- make clear that `P(Q)` in section "Setup" is a price. When I read first it felt to me it is a demand "Inverse demand is linear:"; I guess for pedagogical reasons for people like me it could benefit from something like "Assuming a  market price in terms of an inverse demand, where the price drops with increased supply:" (if that is correct). 
- you are saying "The nodal price (shadow price of the bus power balance constraint) clears the market:" but the formula does not include any nodal reference. is that rather the cupper-plated market price or should we think of this as the "market inside a bus"
- there is two cournot markups (one linear, the other quadratic). that's fine I guess?
- we could probably for the sake of ease just add the cournot markup to `marginal_cost_quadratic` - that would lead to the same outcome or? 
- I am still struggling to understand that concept of inverse demand `P(Q) = a - b Q` . does that mean we are talking about a special case of market where the price per MWh decreases with supply?  where do they appear in the real world. perhaps short term wholesale market? 
 
 EDIT: confusion about inverse demand was cleared up offline :)

FabianHofmann (approved):
and if you want to touch the notebook again, you can remove the second `import pypsa` and perhaps move the pandas import up, to have them all in one place

Irieo:
> @Irieo this is such a wonderful notebook and explanation! I learned a lot and after all this math I need a coffee :D
> * we could probably for the sake of ease just add the cournot markup to `marginal_cost_quadratic` - that would lead to the same outcome or?

Thanks @FabianHofmann!

- wrt abbreviations and hints for the reader -- I address via a small patch
- wrt the highlighted comment -- yes! Here's the proof: https://iriepin.com/uploads/pypsa_cournot_two_ways.py

PR #1576, opened 2026-02-23
Maintenance optimization

### LINKED ISSUE #1003: Add optimization for planned generator maintenance (outage)
A subset of power plants has to undergo maintenance once a year, but **when** is an endogenous decision. The decision is **when** and **which** asset undergoes maintenance, and the duration is fixed to something like X timesteps. This feature is particularly useful for better dispatch modelling and CEM modelling where a "typical" summer capacity reduction is not feasible because the model captures multiple latitudes. 

Some variables (e.g. "planned maintenance duration as a fraction of time steps") and constraints (= active, if the variable is non-zero) could be added to PyPSA such that the optimization will optimize when the maintenance (outage) takes place. 

The LP and MILP formulations are possible. @Irieo had experience with the latter. 
Here is an inspiration of how it could work as LP formulation:
image
image
image

We will eventually test this & then contribute it if successful within the following year.

Update (other examples):
- GridPath has linear and binary [retirement and expansion](https://gridpath.readthedocs.io/en/latest/functionality.html#module-gridpath.project.capacity.capacity_types.gen_ret_lin)

FabianHofmann:
Closes #1003

Add maintenance scheduling optimization for `Generator`, `Link` and `Process` components. PyPSA now co-optimizes when to schedule maintenance windows, see #1003 for the motivation.

Maintenance is configured per asset with `maintainable=True`, `maintenance_duration` (elapsed time per event in units of `snapshot_weightings.generators`), `maintenance_events` (required number of events per horizon) and `maintenance_pu` (share of the asset taken out).

**Formulation** (reworked following the review):

- Only the maintenance start variable $ms_t \in \{0,1\}$ is binary; the maintenance status $m_t \in [0,1]$ is continuous with integrality implied.
- A single window-coverage equality $m_t = \sum_{t':\, t \in \text{cov}(t')} ms_{t'}$ enforces contiguity, non-overlap (via $m \le 1$) and integrality of $m$ in one go.
- Coverage windows are precomputed per potential start snapshot as the minimal run of consecutive snapshots whose weightings accumulate to at least `maintenance_duration` (round-up semantics: each event lasts at least the requested elapsed time). With this, the former total-duration constraint is implied and dropped; only the event count $\sum_t ms_t = E$ remains.
- Starts whose window would run past the horizon or span snapshots in which the asset is inactive (multi-investment periods) are forbidden.
- For extendable assets, the bilinearity $\hat{p} \cdot m$ is linearized **exactly** with a McCormick envelope, since $m$ only takes integral values. The auxiliary variable is internal and not written to network outputs. This requires a finite `p_nom_max`, enforced by a consistency check.

Going one step beyond the review suggestion (which proposed snapshot counting as in unit commitment), `maintenance_duration` honors snapshot weightings, consistent with the planned direction of making unit commitment weighting-aware as well.

**For the reviewer:**
There is an example notebook (`docs/examples/maintenance-scheduling.ipynb`) explaining the feature. We might not pull this in later, but it serves as a good entry point into the topic.

**Open questions:**

- Do we want to make `maintainable` an official attribute or derive it solely from a non-NaN `maintenance_duration`?
- Committable + maintenance interplay: the current dispatch bounds effectively force `status=1` while in maintenance (documented as a caveat). A clean decoupling via $u_t + m_t \le 1$ is exact only for `maintenance_pu=1`; fractional maintenance on a committed unit needs a design decision.

## Checklist


pypsa/data/component_attrs/generators.csv
```diff
@@ -29,6 +29,10 @@ discount_rate,float,per unit,NaN,"Discount rate for annuity calculation when usi
 fom_cost,float,currency/MW,0,"Fixed period operation and maintenance costs per MW, added to annuitized investment cost.",Input (optional)
 efficiency,static or series,per unit,1,"Ratio output and primary energy carrier input (e.g. 0.4 MWh~elec~/MWh~fuel~). This is required for global constraints on primary energy in optimization.",Input (optional)
 committable,boolean,n/a,False,Apply unit commitment constraints.,Input (optional)
+maintainable,boolean,n/a,False,Apply maintenance scheduling constraints.,Input (optional)
+maintenance_duration,int,snapshots,0,Consecutive snapshots per maintenance event. Only if `maintainable=True`.,Input (optional)
+maintenance_pu,float,per unit,1.0,Fraction of capacity unavailable during maintenance (0-1). Only if `maintainable=True`.,Input (optional)
+maintenance_events,int,n/a,1,Number of maintenance events per optimization horizon. Only if `maintainable=True`.,Input (optional)
```
fneum:
```suggestion
maintenance_events,int,n/a,1,Number of required maintenance events per optimization horizon. Only if `maintainable=True`.,Input (optional)
```

pypsa/data/component_attrs/generators.csv
```diff
@@ -29,6 +29,10 @@ discount_rate,float,per unit,NaN,"Discount rate for annuity calculation when usi
 fom_cost,float,currency/MW,0,"Fixed period operation and maintenance costs per MW, added to annuitized investment cost.",Input (optional)
 efficiency,static or series,per unit,1,"Ratio output and primary energy carrier input (e.g. 0.4 MWh~elec~/MWh~fuel~). This is required for global constraints on primary energy in optimization.",Input (optional)
 committable,boolean,n/a,False,Apply unit commitment constraints.,Input (optional)
+maintainable,boolean,n/a,False,Apply maintenance scheduling constraints.,Input (optional)
+maintenance_duration,int,snapshots,0,Consecutive snapshots per maintenance event. Only if `maintainable=True`.,Input (optional)
+maintenance_pu,float,per unit,1.0,Fraction of capacity unavailable during maintenance (0-1). Only if `maintainable=True`.,Input (optional)
```
fneum, line 34: Maybe you already thought about it: Why not just `maintenance_p_max_pu` and implement it as fraction of capacity *available* during maintenance?

The name `maintenance_pu` is sweet and short, but perhaps not very clear on what the per-unit refers to.

I can imagine both ways, just maybe worth discussing briefly.

pypsa/data/component_attrs/generators.csv
```diff
@@ -29,6 +29,10 @@ discount_rate,float,per unit,NaN,"Discount rate for annuity calculation when usi
 fom_cost,float,currency/MW,0,"Fixed period operation and maintenance costs per MW, added to annuitized investment cost.",Input (optional)
 efficiency,static or series,per unit,1,"Ratio output and primary energy carrier input (e.g. 0.4 MWh~elec~/MWh~fuel~). This is required for global constraints on primary energy in optimization.",Input (optional)
 committable,boolean,n/a,False,Apply unit commitment constraints.,Input (optional)
+maintainable,boolean,n/a,False,Apply maintenance scheduling constraints.,Input (optional)
+maintenance_duration,int,snapshots,0,Consecutive snapshots per maintenance event. Only if `maintainable=True`.,Input (optional)
```
fneum:
```suggestion
maintenance_duration,int,snapshots,0,Consecutive snapshots per maintenance event during which capacity reduction is applied. Only if `maintainable=True`.,Input (optional)
```

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+
+$$m_{*,t} \leq \sum_{t'=t-d+1}^{t} ms_{*,t'}$$
+
+Constraint name: `*-maint-duration`
+
+## Horizon Restriction
+
+Maintenance cannot start in the last $d-1$ snapshots (to ensure there are enough remaining snapshots to complete the maintenance block):
+
+$$\sum_{t \in \text{last}(d-1)} ms_{*,t} \leq 0$$
```
fneum: Why not just?:

$ms_{*,t} = 0 \forall t \in \text{last}(d-1)$

Cannot start maintenance in last d-1 snapshots. Avoids dense sum.

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Maintenance Scheduling
+
+Maintenance scheduling constraints are implemented for the [`Generator`][pypsa.components.Generators] and [`Link`][pypsa.components.Links]
```
fneum: What about `Process`?

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Committable Components
+
+When combined with unit commitment (`committable=True`), the dispatch bounds incorporate both the commitment status $u$ and the maintenance status $m$:
+
+=== "Generator (fixed capacity)"
+
+    | Constraint | Name |
+    |-------------------|------------------|
+    | $g_{n,s,t} \geq \underline{g}_{n,s,t} \cdot \hat{g}_{n,s} \cdot u_{n,s,t} - \underline{g}_{n,s,t} \cdot \hat{g}_{n,s} \cdot \alpha_{n,s} \cdot m_{n,s,t}$ | `Generator-com-p-lower` |
```
fneum: This can go negative if $u_{n,s,t} = 0$ and $m_{n,s,t}=1$. Is the dispatch bounded to 0 elsewhere?

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+
+Constraint name: `*-maint-event-count`
+
+## Total Duration
+
+The total weighted time in maintenance equals the duration per event $d$ times the number of events $E$, where $w_t$ are the snapshot weightings:
+
+$$\sum_{t} w_t \cdot m_{*,t} = d \cdot E$$
+
+Constraint name: `*-maint-total-duration`
```
fneum: Isn't this already implied through the other constraints (i.e. the window constraint below)? I think this can be dropped because both the number of maintenance starts and their duration is known.

If we follow the rest of unit commitment, we do not take into account of $w_t$ but just the number of snapshots (as component documentation also states).

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Maintenance Scheduling
+
+Maintenance scheduling constraints are implemented for the [`Generator`][pypsa.components.Generators] and [`Link`][pypsa.components.Links]
+components. For components marked as `maintainable=True`, new binary maintenance variables
+$m_{*,t} \in \{0,1\}$ are introduced, which indicate whether the component is
```
fneum: My change suggestions below build on redefining $m$ as continuous:

$m_{*,t} \in \[0,1\]$

and only the start points $ms$ as binary

$ms_{*,t} \in \{0,1\}$

The integrality of $m_{*,t}$ is implied through $ms$, bounds and (adapted) contiguity constraint.

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+
+$$\sum_{t} w_t \cdot m_{*,t} = d \cdot E$$
+
+Constraint name: `*-maint-total-duration`
+
+## Contiguity
+
+Each maintenance snapshot must be covered by a maintenance start event within the previous $d$ snapshots, enforcing contiguous maintenance blocks:
+
+$$m_{*,t} \leq \sum_{t'=t-d+1}^{t} ms_{*,t'}$$
```
fneum: This version ensures contiguity, non-overlap and integrality of $m$ in one go:

$$m_{...,t} - \sum_{t'=t-d+1}^{t}  ms_{...,t'} = 0$$

No further constraints such as 

$$\sum_{t'=t-d+1}^{t}  ms_{...,t'} \leq 1 $$

required.

Maximum one maintenance start in each window.

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+
+The McCormick envelope bounds on $z$ are:
+
+| Constraint | Name |
+|-------------------|------------------|
+| $z_{*,t} \leq \hat{g}_{*}$ | `*-maintcap_upper` |
+| $z_{*,t} \leq \hat{g}^{\max}_{*} \cdot m_{*,t}$ | `*-maintcap_upper_nommax` |
+| $z_{*,t} \geq \hat{g}_{*} + \hat{g}^{\max}_{*} \cdot m_{*,t} - \hat{g}^{\max}_{*}$ | `*-maintcap_lower_nommax` |
+
+Together with $z \geq 0$, these form the convex hull of $z = \hat{g} \cdot m$ for $m \in \{0,1\}$ and $0 \leq \hat{g} \leq \hat{g}^{\max}$.
```
fneum: How do we know this relaxation is good enough?

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+??? note "Mapping of symbols to component attributes"
+
+    === "Generator"
+
+        | Symbol | Attribute | Type |
+        |-------------------|-----------|-------------|
+        | $g_{n,s,t}$       | `n.generators_t.p` | Decision variable |
+        | $m_{n,s,t}$       | `n.generators_t.maintenance` | Decision variable |
+        | $ms_{n,s,t}$      | `n.generators_t.maintenance_start` | Decision variable |
+        | $z_{n,s,t}$       | `n.generators_t.maintenance_capacity` | Decision variable |
```
fneum: Should this perhaps just be a 'hidden' auxiliary variable? Not sure.

pypsa/optimization/variables.py
```diff
@@ -199,6 +199,95 @@ def define_shut_down_variables(
...
+    """
+    c = n.c[c_name]
+    maint_i = c.maintainables.difference(c.inactive_assets)
+
+    if maint_i.empty:
+        return
+
+    active = c.da.active.sel(name=maint_i, snapshot=sns)
+    n.model.add_variables(
+        coords=active.coords, name=f"{c.name}-maintenance", mask=active, binary=True
```
fneum: See suggestion in user guide to make this `binary=False`, but requires bounds of [0,1]

fneum (changes requested):
I like the implementation as part of the least-cost optimisation and the documentation was pretty clear. One would build on top later some reserve constraints that limit decision freedoms of generators going into maintenance.

I think one can do some rewiring of variables and constraints to simplify and minimise the computational impact and number of binary variables.

In particular, keeping only the event start as binary variable, dropping the event duration constraint, and reformulating the contiguity constraint so that it captures both contiguity and non-overlap.

pypsa/data/component_attrs/generators.csv
fneum: After having reviewed the rest of the code, I like the current implementation.

fneum:
Ah, and I briefly thought about whether the maintenance optimisation could be integrated with the committable `status` variable, i.e. constraints on `status=0` number of events with given lengths. But that doesn't allow for fractional maintenance.

MaykThewessen:
Hi @FabianHofmann — just opened #1697 (draft) for a small helper handling **deterministic, user-supplied outage schedules** (write known windows into `p_max_pu_t` / `p_min_pu_t` pre-solve, no variables added). Issue #1696 explains the scope and why I think it's orthogonal to the MILP maintenance-optimization work here rather than a subset — different lifecycle, different sub-package (`pypsa.utils` vs `pypsa.optimization`), and the use cases are mostly historical replay / ENTSO-E unavailability records where the windows are known a priori. Happy to fold into a single PR or module if you'd prefer one entry point for everything outage-related; otherwise I'd welcome your sign-off on the orthogonality call so #1697 can move ahead independently. No rush.

FabianHofmann:
@fneum your reformulation idea was great and worked out! It was missing the snapshot weighting but it should be solid now (I would rather follow the convention of elapsed time when we talk about duration - UC min/up time is still an exception that we should fix). 

- About the "relaxation": it is not really a relaxation in that sense since we are applying it on $m$ which is bound by 0 and 1. The docs is updated and make it clearer. 
- the `maintenance_capacity` is no longer written to the network outputs.
- Processes are now supported as well


I still need to figure out the connection to shut down generators in UC. doing that now

FabianHofmann:
@fneum I prepared a notebook for showcasing the features. run it with 

```
gh pr checkout 1576
uv run jupyter notebook docs/examples/maintenance-scheduling.ipynb
```

would be nice to pull this in soon :)

docs/examples/maintenance-scheduling.ipynb
```diff
@@ -0,0 +1,827 @@
...
+    "\n",
+    "plt.tight_layout()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c2d19c2a",
+   "metadata": {},
+   "source": [
+    "!!! note\n",
```
lkstrp: This doesn't render

docs/examples/maintenance-scheduling.ipynb
```diff
@@ -0,0 +1,827 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cab44da6",
+   "metadata": {},
+   "source": [
+    "## Links and Processes\n",
+    "\n",
+    "The same attributes (`maintainable`, `maintenance_duration`, `maintenance_pu`, `maintenance_events`) apply to [`Link`][pypsa.components.Links] and [`Process`][pypsa.components.Processes] components, where the identical formulation constrains their power flow $p$. The `Process` component follows the same formulation as the `Link`. See the [user guide](../../user-guide/optimization/maintenance-scheduling.md) for the full mathematical description."
```
lkstrp: Links to other components are broken

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,219 @@
...
+
+    | Constraint | Name |
+    |-------------------|------------------|
+    | $g_{n,s,t} \geq \underline{g}_{n,s,t} \cdot \hat{g}_{n,s} \cdot (u_{n,s,t} - \alpha_{n,s} \cdot w_{n,s,t})$ | `Generator-com-p-lower` |
+    | $g_{n,s,t} \leq \bar{g}_{n,s,t} \cdot \hat{g}_{n,s} \cdot (u_{n,s,t} - \alpha_{n,s} \cdot w_{n,s,t})$ | `Generator-com-p-upper` |
+    | $w_{n,s,t} \leq u_{n,s,t}$ | `Generator-maint-status-le-status` |
+    | $w_{n,s,t} \leq m_{n,s,t}$ | `Generator-maint-status-le-maint` |
+    | $w_{n,s,t} \geq u_{n,s,t} + m_{n,s,t} - 1$ | `Generator-maint-status-lb` |
+
+    Together with $w \geq 0$, these McCormick inequalities give $w = u \cdot m$ exactly, since $m$ is binary. This holds for binary status (standard unit commitment) and for the relaxed continuous status of `linearized_unit_commitment=True`. The auxiliary variable $w$ is internal and not written to the network outputs.
```
lkstrp: $ms$ is binary, $m$ is not? And $w$ is also already used for snapshot weightings, this confuses my overstretched brain

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,219 @@
...
+
+Maintenance cannot start where the coverage window does not fit, i.e. where the
+remaining weighted horizon is shorter than $d$ or where the window would span
+snapshots in which the component is inactive:
+
+$$ms_{*,t'} = 0 \quad \forall t' : \text{cov}(t') \text{ incomplete or partially inactive}$$
+
+Constraint name: `*-maint-start-horizon`
+
+These constraints are defined in the functions `define_maintenance_variables()`, `define_maintenance_start_variables()`, `define_maintenance_capacity_variables()`, and `define_maintenance_constraints()`.
```
lkstrp: There is one missing here, I am not even sure if we should mention them. Those are not exposed in the API, so I would drop the full line. If a user dives into code, they will manage with just the docstrings. And Bryn is coming anyway

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,219 @@
...
+        | $w_{l,t}$         | internal, not written to outputs  | Decision variable |
+        | $z_{l,t}$         | internal, not written to outputs  | Decision variable |
+        | $\hat{f}_{l}$     | `n.links.p_nom`  | Parameter |
+        | $\hat{f}^{\min}_{l}$ | `n.links.p_nom_min` | Parameter |
+        | $\hat{f}^{\max}_{l}$ | `n.links.p_nom_max` | Parameter |
+        | $\underline{f}_{l,t}$| `n.links_t.p_min_pu`  | Parameter |
+        | $\bar{f}_{l,t}$   | `n.links_t.p_max_pu`  | Parameter |
+        | $\alpha_{l}$      | `n.links.maintenance_pu`  | Parameter |
+        | $d$               | `n.links.maintenance_duration`  | Parameter |
+        | $E$               | `n.links.maintenance_events`  | Parameter |
```
lkstrp, line 209: Processes are missing here. I guess that's our fault of holding it back that long, probably worth to send an agent to find all missing processes refs, also in example notebook

pypsa/components/components.py
```diff
@@ -907,6 +907,26 @@ def committables(self) -> pd.Index:
...
+    def maintainables(self) -> pd.Index:
+        """Get the index of maintainable elements of this component.
+
+        Returns
+        -------
+        pd.Index
+            Single-level index of maintainable elements.
+
+        """
+        if "maintainable" not in self.static:
```
lkstrp, line 920: Shouldn't instead this just be available for Generators, Links, Processes? The rest is not maintainable. It also referes to #1665, I am pretty sure if you remove it we run into problems in the constraint functions again since those are also looping components which are not touched. But this also applies for the other methods in here. Fine to keep for now

EDIT: Nevermind, if I look at constraint code this will complicate things way too much

pypsa/optimization/constraints.py
```diff
@@ -599,6 +719,123 @@ def define_operational_constraints_for_committables(
...
+        z = n.model[f"{c.name}-maintenance_capacity"].sel(name=maint_ext_i)
+        active_me = active.sel(name=maint_ext_i)
+
+        n.model.add_constraints(
+            z + p_nom_min <= p_nom + p_nom_min * m,
+            name=f"{c.name}-maintcap_upper",
+            mask=active_me,
+        )
+        n.model.add_constraints(
+            z <= p_nom_max * m,
```
lkstrp, line 821: `p_nom_max` defaults `inf` here which leads to `inf` constraints. This is checked by the consistency checks, but just logged, so most likely ignored. Highs is than declining to solve something, and still returns `ok`. Should the consistency check just be strict as default? 

``` python
import pypsa

n = pypsa.Network()
n.set_snapshots(range(2))
n.add("Bus", "b")
n.add("Generator", "g", bus="b", p_nom_extendable=True, capital_cost=1,
      maintainable=True, maintenance_duration=1)  # p_nom_max defaults to inf
n.add("Generator", "backup", bus="b", p_nom=100, marginal_cost=50)
n.add("Load", "l", bus="b", p_set=50)

print("status   :", n.optimize())                            # ('ok', 'unknown')
print("objective:", n.objective)                             # 0.0
print("generated:", n.c.generators.dynamic.p.sum(axis=1).values, "vs load 50")
```

pypsa/optimization/optimize.py
```diff
@@ -697,6 +702,10 @@ def create_model(
             define_status_variables(n, sns, c, linearized_unit_commitment)
             define_start_up_variables(n, sns, c, linearized_unit_commitment)
             define_shut_down_variables(n, sns, c, linearized_unit_commitment)
+            define_maintenance_variables(n, sns, c)
```
lkstrp, line 705: As already mentioned above, should we not just start calling them on the components for which they matter? EDIT: as above, never mind

test/test_lopf_maintenance.py
```diff
@@ -0,0 +1,703 @@
...
+    n.add("Load", "load", bus="bus", p_set=load)
+
+    status = n.optimize()
+    assert status[0] == "ok"
+
+    maint = n.c.generators.dynamic.maintenance["gen"].values > 0.5
+    weights = n.snapshot_weightings.generators.values
+    assert weights[maint].sum() >= 3
+    assert set(np.where(maint)[0]) <= set(range(5, 10))
+    assert maint.sum() == 3
```
lkstrp: There is no test with active attribute in here, you think that's worth to add?

lkstrp (commented):
I like this quite a lot! And thanks for the fantastic example notebook!

Next to all the comments one question: Having multiple `maintenance_events`, doesn't allow me to have different `maintenance_durations` per event. We don't do that because it will not be needed or the API get's too tricky? I guess the formulation should be fine? Not voting for it, just curious.

pypsa/optimization/constraints.py
FabianHofmann: good one, will do that!

docs/user-guide/optimization/maintenance-scheduling.md
FabianHofmann: yeah, let's remove that

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,219 @@
...
+and [`Process`][pypsa.components.Processes] components. The `Process` component follows the same formulation as the `Link`,
+applied to its internal power $p$. For components marked as `maintainable=True`, two new variables are
+introduced: binary maintenance start variables $ms_{*,t} \in \{0,1\}$, indicating
+that a maintenance event begins in snapshot $t$, and continuous maintenance status
+variables $m_{*,t} \in [0,1]$, indicating whether the component is undergoing
+maintenance (1) or available (0). The integrality of $m$ is implied through the
+window coverage equality below, so only $ms$ enters the model as a binary. This
+turns the model into a mixed-integer linear programme (MILP).
+
+## Dispatch Coupling
```
fneum: I would consider moving this section towards the end of the page, mainly because the (necessarily) complicated setup with McCormick and Big-M comes quite early here before the main maintenance scheduling constraints are defined.

I would also name it "Effect on Dispatch Limits" so it's closer the "Dispatch Limits" section in optimisation user guide.

docs/user-guide/optimization/maintenance-scheduling.md
```diff
@@ -0,0 +1,219 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Maintenance Scheduling
```
fneum, line 7: I would really like it if we could link this formulation up with 1-2 canonical literature/textbook references (as we do for unit commitment and stochastic optimisation).

Perhaps the following get kind-of close:

- Andrade (2025) On the tightness of linear relaxations of alternative mixed integer programming formulations for the generator maintenance scheduling problem https://arxiv.org/abs/2502.08855 - formulation 6/7 in review

- Fabrice Fourcade, Ellis Johnson, Mourad Bara, Philippe Cortey-Dumont,
Optimizing nuclear power plant refueling with mixed-integer programming,
European Journal of Operational Research, Volume 97, Issue 2,
1997, Pages 269-280, ISSN 0377-2217, https://doi.org/10.1016/S0377-2217(96)00197-X. - multiple maintenance events

- J. F. Dopazo and H. M. Merrill, "Optimal generator maintenance scheduling using integer programming," in IEEE Transactions on Power Apparatus and Systems, vol. 94, no. 5, pp. 1537-1545, Sept. 1975, https://doi.org/10.1109/T-PAS.1975.31996. - very original from 1975

fneum (approved):
Very nice! The math is quite elegant now (and I am happy you went through all the more complicated cases of committable and extendable components).

Also docs and example are excellent (and @lkstrp already mentioned the finishing touches regarding rendering an `Process` component).

I had two comments on perhaps A) reordering the subsections in the user guide and B) finding 1-2 canonical literature references to link elements of the PyPSA formulation to. The *generator maintenance scheduling* is quite a standard term in OR literature.

docs/user-guide/optimization/maintenance-scheduling.md
FabianHofmann: wonderful stuff thanks. I guess I should have looked at these earlier. from what I see, they pretty much cover the functionality as we implemented it. done.

test/test_lopf_maintenance.py
FabianHofmann: worth it. done and passes

docs/user-guide/optimization/maintenance-scheduling.md
FabianHofmann: done

docs/user-guide/optimization/maintenance-scheduling.md
FabianHofmann: you are totally right, I used `v` no for te maintenance status

FabianHofmann:
@fneum and @lkstrp thanks for the reviews. I addressed everything, the doc structure improved by the suggestions of you. I'll wait to see if it renders correctly and then merge this in

lkstrp (approved):

PR #1578, opened 2026-02-24
fix: properly assign controllable branches

FabianHofmann:
Quite some bug, but nothing severe (hopefully downstream too)

### DIFF
--- a/pypsa/networks.py
+++ b/pypsa/networks.py
@@ -1137,17 +1137,14 @@ class Network(
         Examples
         --------
         >>> n.controllable_branches() # doctest: +ELLIPSIS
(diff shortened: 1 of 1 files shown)


lkstrp:
Good catch! But where is my release note, this is still fixing/ changing directly exposed behavior

And these days we would even add a test for that

FabianHofmann:
> Good catch! But where is my release note, this is still fixing/ changing directly exposed behavior
> 
> And these days we would even add a test for that

I have to remind myself that the days of lazy one-liner PRs are officially over. it is actually tested by the doc test. could of course be better with checking the first level index against `n.controllable_branch_components`.

FabianHofmann:
fine if I append the release note in #1333?

PR #1593, opened 2026-03-03
docs: reduce peak memory in scigrid-redispatch example

### LINKED ISSUE #1586: Scigrid-redispatch example: 3x memory from holding all network copies simultaneously
## Description

The [scigrid-redispatch example](https://docs.pypsa.org/latest/examples/scigrid-redispatch/) demonstrates the standard two-stage market + redispatch pattern. This pattern creates two full `network.copy()` calls (zonal market + redispatch model) and holds all three networks in memory simultaneously.

For the small SciGRID network this is fine, but when users apply this pattern to **production-scale networks** (500+ buses, 3500+ generators, 8760 snapshots), peak memory becomes 3× the network size (~1.2 GB just for network objects, before solver allocations). This can cause OOM on systems with limited RAM.

## The pattern in the example

```python
# All three held in memory at the same time:
network = pypsa.Network("solved_network.nc")   # ~400 MB
m = network.copy()   # zonal market copy        ~400 MB
# ... solve m ...
n = network.copy()   # redispatch model copy    ~400 MB  (m still alive!)
# ... use m.generators_t.p to fix dispatch in n ...
# ... solve n ...
```

After solving the zonal model `m`, only the **dispatch result** (`m.generators_t.p`) and the **objective** (`m.objective`) are needed to build the redispatch model. The full zonal network copy can be freed before building `n`.

## Proposed improvement

Update the example to demonstrate extracting lightweight results and freeing memory:

```python
# Solve zonal market
m = network.copy()
# ... aggregate buses, solve ...
zonal_dispatch = m.generators_t.p.loc[snapshots].copy()
zonal_cost = m.objective
zonal_prices = m.buses_t.marginal_price.loc[snapshots].copy()
del m  # Free ~400 MB before building redispatch model

# Build redispatch using lightweight DataFrames instead of full Network
n = network.copy()
# ... fix dispatch using zonal_dispatch (DataFrame), not m (Network) ...
```

Additionally, `network.copy(snapshots=snapshots)` could be highlighted as a way to create a smaller copy when only a subset of snapshots is needed for the analysis.

**Effort**: Documentation/example update only — no changes to PyPSA core needed.
**Benefit**: ~50% memory reduction for the redispatch analysis pattern on large networks. Prevents OOM when running full-year redispatch analysis.

## Context

We run this pattern on networks with 500 buses, 3500 generators, and 8760 hourly snapshots. With the original pattern, the process OOM'd during batch processing of multiple scenarios. After extracting zonal results early and freeing `m`, memory usage dropped from ~1.2 GB to ~800 MB for the network objects alone.

## Environment

- PyPSA: 0.31.x / 1.0+
- Network scale: 500 buses, 3500 generators, 8760 snapshots
- OS: macOS / Linux

MaykThewessen:
## Summary

- Defer creation of the redispatch network copy `n` until after the zonal market model `m` is solved and freed
- Extract lightweight DataFrames (`market_dispatch`, `market_p_max_pu`, `market_p_nom`) from `m` before `del m`
- Ensures at most **2 network copies** coexist at any time instead of **3**

## Motivation

The current example creates all three network copies (`o`, `m`, `n`) upfront in cell 5, keeping them in memory simultaneously throughout the entire workflow. On production-scale networks (500+ buses, 8760 snapshots), this causes ~3x peak memory usage when only ~2x is necessary.

The fix follows the same pattern already used in real-world redispatch implementations: extract the few DataFrames needed from the zonal market result, free the full network, then build the redispatch model.

Closes #1586

## Changes

| Before | After |
|--------|-------|
| `n = o.copy()` created upfront (cell 5) | `n = o.copy()` created after `del m` |
| `m` held in memory until end | `m` freed after extracting results |
| 3 networks coexist during redispatch build | Max 2 networks coexist |
| References `m.generators_t.p`, `m.generators.p_nom` | References `market_dispatch`, `market_p_nom` DataFrames |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/examples/scigrid-redispatch.ipynb
+++ b/docs/examples/scigrid-redispatch.ipynb
@@ -79,7 +79,9 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
-   "source": "m = o.copy()  # for market model"
+   "source": [
+    "m = o.copy()  # for market model"
+   ]
   },
   {
    "cell_type": "code",
@@ -141,7 +143,9 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
-   "source": "zones = (o.buses.y > 51).map(lambda x: \"North\" if x else \"South\")"
+   "source": [
+    "zones = (o.buses.y > 51).map(lambda x: \"North\" if x else \"South\")"
+   ]
   },
   {
    "cell_type": "markdown",
@@ -220,15 +224,20 @@
   },
   {
    "cell_type": "markdown",
-   "source": "Before building the redispatch model, we extract the market results we need and free the market model. This avoids holding all three network copies (`o`, `m`, `n`) in memory simultaneously. On production-scale networks, this reduces peak memory by ~33%.",
-   "metadata": {}
+   "metadata": {},
+   "source": "Before building the redispatch model, we extract the market results we need and free the market model. This avoids holding all three network copies (`o`, `m`, `n`) in memory simultaneously. On production-scale networks, this reduces peak memory by ~33%."
   },
   {
    "cell_type": "code",
-   "source": "market_dispatch = m.generators_t.p.copy()\nmarket_p_max_pu = m.get_switchable_as_dense(\"Generator\", \"p_max_pu\").copy()\nmarket_p_nom = m.generators.p_nom.copy()\ndel m",
-   "metadata": {},
    "execution_count": null,
-   "outputs": []
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "market_dispatch = m.generators_t.p.copy()\n",
+    "market_p_max_pu = m.get_switchable_as_dense(\"Generator\", \"p_max_pu\").copy()\n",
+    "market_p_nom = m.generators.p_nom.copy()\n",
+    "del m"
+   ]
   },
   {
    "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


Irieo:
@MaykThewessen I am struggling to see the purpose of this work. The PyPSA example suite contains many notebooks with simple pedagogical toy models. Many of these were written before LLM era by humans, for humans, and focused on illustration of a concept rather than optimizing memory usage within the example (which is obv different to the approach in PyPSA core optimization module or linopy module).

With current LLM tools, code is cheap (aside environmental consideration). It is immediate to generate “improvements” and open an issue or PR. However, reviewing PR and validating the change cost more human attention than was spent producing the PR itself. This raises asymmetry of push-and-pull resources, a big problem in the OS community these days.

This concern applies to a dozen of the issues and PRs you have opened across the PyPSA ecosystem within yesterday. Please keep in mind that each of these requires real human effort to judge, test, and merge. That effort should be justified by meaningful intent.

MaykThewessen:
okay, will only try to raise proposals when i think they are really valuable, and will try to save time for everyone to read

docs/examples/scigrid-redispatch.ipynb
```diff
@@ -226,26 +225,41 @@
...
-    "Next, based on the market outcome with two bidding zones `m`, we build a secondary redispatch market `n` that rectifies transmission constraints through curtailment and ramping up/down thermal generators.\n",
-    "\n",
-    "First, we fix the dispatch of generators to the results from the market simulation. (For simplicity, this example disregards storage units.)"
+    "## Build redispatch model `n`"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": "Next, based on the market outcome with two bidding zones, we build a secondary redispatch market `n` that rectifies transmission constraints through curtailment and ramping up/down thermal generators.\n\nWe create the redispatch model here (after freeing `m`) rather than upfront, so that at most two network copies coexist at any time.\n\nFirst, we fix the dispatch of generators to the results from the market simulation. (For simplicity, this example disregards storage units.)"
```
lkstrp, line 252: Yeah as @Irieo already says. We are happy for contributions. But this is just stuff which needs to be cleaned up before you file in PRs in any case. It is an issue if we are the first ones who read over Claudes output. Please keep that in mind

Issue #1482, opened 2025-12-10
Unit purchase cost (cost incurred per built decision not capacity)


fneum:
Idea is to add some cost when capacities bumps from 0 to >0, in addition to cost/MW. This requires a binary variable.

Ties in nicely with #1473.

brynpickering:
This would be solved as a first pass by giving an example of applying it using piecewise segments

brynpickering:
Now that piecewise has been merged in, I can confirm that this _isn't_ possible with our piecewise implementation. This is because our piecewise attributes (capital cost, marginal cost, efficiency) are defined per unit. So, once denormalised y=0 at x=0, always. 

To represent this, we need the equivalent of a `stand-by-cost` for unit purchase. Not sure what the best name is here. Some options (with equivalent `overnight_cost_...` for `capital_cost_...` options):

- `purchase_cost`
- `capital_cost_mod` 
- `capital_cost_unit`
- `purchase_capital_cost`
- `capital_cost_fixed`

Any preference @fneum ?

PR #1603, opened 2026-03-04
feat: Add piecewise linear costs and constraints

### LINKED ISSUE #1002: Load dependent efficiency curves
Hi, 

Big fan of this package. 

Are there any plans to implement load dependent efficiency curves? For example, a link might have a higher efficiency at 80% load than 20% load. I appreciate there was a response [here](https://groups.google.com/g/pypsa/c/ViJcmf9Xh9Y) a while ago but was wondering if the view has changed. 

I have seen it linearised in the following two ways and appreciate both incur a computational overhead. 

**Slope and offset** 
- https://oemof-solph.readthedocs.io/en/v0.5.3/usage.html#offsetconverter-component

**Piecewise linearisation** 
- https://github.com/spine-tools/SpineOpt.jl/blob/master/docs/src/advanced_concepts/user_constraints.md
- https://github.com/FZJ-IEK3-VSA/FINE/blob/master/examples/11_Partload/11_Partload_Example.ipynb

Let me know. 

Many thanks,

Harry

### LINKED ISSUE #1020: Add Option for Marginal Cost Offer Curve
It would be extremely helpful to add multipart offer curves to generators, rather than just having a static marginal cost or quadratic marginal cost. In many balancing authorities, generators as well as loads are allowed to submit stepwise increasing offer curves rather than a single offer for their whole capacity. This feature would make PyPSA more robust and better able to capture unit economic dynamics. 

This feature would require a new variable that may or may not be time varying. This variable would be represented as a series that would break generator's marginal cost into chunks based on percentage of generation. For example, a 100mw generator might be represented in this manner:
20mw for $20/mw
40mw for $30/mw
35mw for $35/mw
5mw for $50/mw
This could be captured in a series of tuples formatted like ((.2, 20), (.4, 30), (.35, 35), (.05, 50))

In addition to the actual implementation, several consistency checks would need to be added. First would be a check to make sure that the total capacity being broken up summed to one. The marginal cost curve would also need to be monotonically increasing and may be limited to a set number of offers, ten is likely more than enough. 

Files that would need changes:
variables.csv
variables.py
consistency.py
optimize.py

It may also be prudent to change the logic in define_objective to include an enum to store the offer type for a generator, rather than using a boolean is_quadratic variable.

### LINKED ISSUE #1473: Piecewise costs and constraints
This builds on #1020 and SOS constraints being introduced in linopy (https://github.com/PyPSA/linopy/pull/495).

We want to be able to represent non-linear relationships, including:

- incremental bid levels (#1020) (links `p` with `marginal_cost`)
- non-linear part-load efficiency (changing ratio between input and output energy flows). These need to be 3D SOS2 constraints to link `p_nom`, `p` and an intermediate variable that tracks dispatch.
- economies of scale (links `p_nom` and `capital_cost`). 

There may be more, but these are the first to start with.

All of these can be described by piecewise linear curves, but how to introduce that data into PyPSA?

Two options we can currently see:

1. different components representing different cost/efficiency levels. An user-defined attribute links them together in a given order. Under the hood, PyPSA collapses these components into the required piecewise mathematical representation. Pro: no need to change/add a new data structure to the PyPSA network object. Con: heavily relies on the user to know what they're doing. Also could introduce unnecessary decision variables in the mathematical problem and requires limiting the parameters that can be changed between the linked components.
2. **preferred** A new network attribute to track piecewise data which has an extra index level for the segments. We would probably limit this to the three examples above, at least to begin with. Pro: easy data input (open question, how to include it in `network.add`?), Con: new data structure that needs to be maintained. 

Internally, we would have auxiliary decision variables to track the resolved SOS2 values (`marginal_cost`, `capital_cost`, `p * efficiency`), which are otherwise simply parameters or expressions.

## workarounds that already exist

- **economies of scale**: unit-wise components of different `p_nom_mod` values that have different capital costs associated with them and the very large units are cheaper per `p_nom` than the smaller units.

## Keeping it linear

In some cases, it's possible to keep a piecewise curve entirely linear, if:

1. The curve is convex and the constraint is a lower bound (costs, which the model is trying to minimise)
2. The curve is concave and the constrain is an upper bound (efficiency, which the model would like to maximise)

In both cases, you would have a series of linear constraints that are tangents of the nonlinear curve. Automatically, the decision variables would sit on the curve defined by these tangents (effectively an equality constraint) as that's the most optimal place to sit.

We could check this automatically and apply SOS2/tangent constraints as necessary (those linked to cost are tangent constraints if the curve is convex. Those linked to efficiency are tangent constraints if the curve is concave)

coroa:
Closes #1473 
closes #1020 
closes #1002 

Adds support for piecewise linear `marginal_cost` and `capital_cost` curves on generators, storage units, stores, links, and processes using linear tangent constraints + piecewise linear `efficiency` curves on links and processes.

#### Definition
Costs are specified as marginal values (slopes) at breakpoints — xarray integrates trapezoidally to recover total cost:

```python
# Piecewise marginal cost (per-unit x-axis, fixed p_nom required)
n.add("Generator", "gen", bus="bus0", p_nom=100,
      marginal_cost={0.0: 40.0, 0.5: 60.0, 1.0: 70.0})

# Piecewise capital cost (absolute x-axis, extendable)
n.add("Generator", "gen", bus="bus0", p_nom_extendable=True, p_nom_max=100,
      capital_cost=pd.DataFrame({"p_nom": [0.0, 100.0], "capital_cost": [0.5, 1.5]}))
```

#### Results
Resolved costs in solution are accessible via attrs the `_opt` suffix:

```python
n.c.generators.dynamic.marginal_cost_opt
n.c.generators.static.capital_cost_opt
```

These have been normalised to per unit capacity/dispatch to match their equivalent non-piecewise equivalents.

#### Optimisation config
On optimisation, default piecewise formulations can be overridden using `piecewise_options`. E.g., the fix the marginal cost to the curve rather than the default (">="):

```python
n.optimize(piecewise_options=[{"component": "Generator", "attribute": "marginal_cost", "operator": "=="}])
```

To do so just for one generator for which piecewise marginal costs have been defined:

```python
n.optimize(piecewise_options=[{"component": "Generator", "attribute": "marginal_cost", "operator": "==", name="gen"}])
```

This is a list of options, so can include an arbitrary number of overrides.

### Open todos

- [x] Add implementation for pwl efficiencies and rates
- [x] Check for convexity (ie. monotonically increasing slopes) and fall back to sos otherwise
- [x] Apply piecewise efficiency to generators
- [x] Apply piecewise efficiency to storage units, and stores(?)
- [x] Add segment data to im-/exporters
- [x] unit test `n.add` with all segmentable attrs
- [x] unit test `piecewise` module
- [x] integration test `optimize.create_model` with all combinations of attrs
- [x] integration test `assign_solution` with all segmentable attrs
- [x] Add docs page
- [x] Resolve infeasibilities when applying piecewise curve to process `rate0`.

### Open questions
- How do we handle the first y value not being 0% and the final y value not being 100% (both effectively apply a `p_<in/max>_pu` to the component)?
  - Do we force covering the full range? 
  - Do we move from defining per unit to absolute x values for dispatch piecewise constraints?
- Why is applying piecewise curve to process `rate0` leading to infeasibilities (see xfailing test)?
- How to document marginal vs non-marginal piecewise curves? I've made an attempt but I'm not sure it hits the mark.
- Do we want piecewise outputs in the static columns by default (e.g. `capital_cost_piecewise_opt`), given that dynamic piecewise attrs only appear when they are needed (e.g. `marginal_cost_piecewise_opt`)?
- Should p1...N always be negative for multiports in the outputs?

## Checklist


brynpickering:
@coroa I notice that my updates leads to the tests you introduced failing. Two reasons for this:

## x-points per-unit or not

We consider `marginal_cost` as per unit `p` and we are making different assumptions for what this means when `p = 0`. Your initial tangent constraints assumed that `marginal_cost` at `p = 0` was the starting cost, I think?

This raises a wider question around 1. whether we define these piecewise attrs against `p` or `p_pu` and 2. whether to force segments to cover the full range (eff/mc: `[0, p_nom]`, capital cost: `[0, p_nom_max]`).

1. If set relative to `p` it would better match `capital_cost`, but would require the user to change the x breakpoints to match any changes made in `p_nom`. The alternative would be to use `p_nom_pu` or similar for the capital cost case, so all `x` breakpoints are defined in the range `[0, 1]`.
2. If we don't enforce the full range, we effectively add an additional constraint on the component. If either marginal cost or efficiency have x_points covering the range `[0.1, 0.9]` it will implicitly force the component to have a p_min_pu of 0.1 and p_max_pu of 0.9.

## y-points per-unit or not

I was considering capital cost as nominal, not per unit. I realise this isn't correct when considering all others uses of `capital_cost`. I can fix this quite easily (just multiply by the `y` breakpoint values by the `x` breakpoint values) but it does mean that we can't ever have a non-zero cost/consumption at zero capacity/operation. 

If you set, say, `capital_cost={0: 40, ...}`, this doesn't mean 40 Euros at zero capacity; it will just set a point at the origin `[0, 0]` since y points are multiplied by x points when creating the piecewise constraint (0 x 40 = 0). We _could_ allow the zero point to be a special case, that allows an offset value to be defined but it would have a different meaning to all other segments, so could be confusing.

To achieve costs/energy consumption at zero `p`/`p_nom`, we'd need to be able to define `y` values as nominal. But then calling it `capital_cost`/`marginal_cost`/`efficiency` doesn't make sense. Perhaps `capex`/`opex`/`p{port}` would work?

## Cumulative y-values or not

I see now that your piecewise tangent approach was based on calculating the integral of the curve up to the given operating point. The new approach I've introduced sets the cost at face value, as given by the input dict. 

In physical terms, for e.g. capital cost, the cumulative approach is the expenditure in capacity considering a single component as being composed of an infinite number of incremental units, with incremental cost `dy` for each incremental capacity increase `dx`. For decreasing capital costs, it means you always pay the high cost for the first X% of capacity, only benefitting from reduced costs for the final Y%. 

I can see this making sense for costs, in some instances. However, non-cumulative costs also make sense. E.g., you might be deciding between a 10MW and 100MW solar park and you get a fixed price for both; you don't pay the same upfront cost for the first 10MW in both cases. Then again, how do you define the capital cost rate of a 50MW plant? The capital cost rate would likely be somewhere between the two but we lose that information in the non-cumulative piecewise constraint unless we actively supply a segment. This is because our auxiliary variable represents CAPEX, so the line between two segments will be linear for CAPEX but not for capital cost (= CAPEX / x).

Not sure what the best approach to representing this ultimately is.

brynpickering:
RE cumulative curves, I've now implemented a `marginal_attr` flag that allows costs to be treated cumulatively (reflecting having to pay the marginal costs along the curve up to each final capacity / dispatch level)

pypsa/data/component_attrs/generators.csv
```diff
@@ -18,16 +18,16 @@ e_sum_max,float,MWh,inf,The maximum total energy produced during a single optimi
 q_set,static or series,MVar,0.,Reactive power set point (for power flow),Input (optional)
 sign,float,n/a,1,"Sign denoting the orientation of the dispatch variable (e.g. positive for generation, negative for consumption).",Input (optional)
 carrier,string,n/a,n/a,"Prime mover energy carrier (e.g. coal, gas, wind, solar); required for global constraints on primary energy in optimisation",Input (optional)
-marginal_cost,static or series,currency/MWh,0.,Marginal cost of production of 1 MWh.,Input (optional)
+marginal_cost,static or series,currency/MWh,0.,Marginal cost of production of 1 MWh.,Input (optional),p_pu
```
fneum: Could add?
```suggestion
marginal_cost,static or segments or series,currency/MWh,0.,Marginal cost of production of 1 MWh.,Input (optional),p_pu
```

pypsa/data/component_attrs/generators.csv
```diff
@@ -52,3 +52,5 @@ mu_lower,series,currency/MWh,n/a,Shadow price of lower `p_nom` limit,Output
 mu_p_set,series,currency/MWh,n/a,Shadow price of fixed power generation `p_set`,Output
 mu_ramp_limit_up,series,currency/MWh,n/a,Shadow price of upper ramp up limit,Output
 mu_ramp_limit_down,series,currency/MWh,n/a,Shadow price of lower ramp down limit,Output
+capital_cost_opt,float,currency/MW,0.,Optimised capital cost when using piecewise linearisation.,Output
+marginal_cost_opt,series,currency/MWh,0.,Optimised marginal cost when using piecewise linearisation.,Output
```
fneum: Judging by how it was used in the example, these are the resulting average €/MWh and €/MW cost, right? ("average marginal cost" is a bit of an oxymoron).

```suggestion
capital_cost_opt,float,currency/MW,0.,Optimised average capital cost when using piecewise linearisation.,Output
marginal_cost_opt,series,currency/MWh,0.,Optimised average marginal cost when using piecewise linearisation.,Output
```

pypsa/data/component_attrs/buses.csv
```diff
@@ -1,20 +1,20 @@
...
-"v_mag_pu_max","float","per unit","inf","Maximum desired voltage, per unit of `v_nom`.  Placeholder attribute not currently used by any functions.","Input (optional)"
-"control","string","n/a","PQ","P,Q,V control strategy for power flow, must be ""PQ"", ""PV"" or ""Slack"". Note that this attribute is an output inherited from the controls of the generators attached to the bus. Do not set by hand.","Output"
-"generator",string,"n/a","n/a","Name of slack generator attached to slack bus.","Output"
-"sub_network","string","n/a","n/a","Name of connected sub-network to which bus belongs. This attribute is set by `n.determine_network_topology()`. Do not set by hand.","Output"
-"p","series","MW",0.,"active power at bus (positive if net generation at bus)","Output"
-"q","series","MVar",0.,"reactive power (positive if net generation at bus)","Output"
-"v_mag_pu","series","per unit",1.,"Voltage magnitude, per unit of `v_nom`","Output"
-"v_ang","series","radians",0.,"Voltage angle","Output"
-"marginal_price","series","currency/MWh",0.,"Shadow price from energy balance constraint","Output"
+attribute,type,unit,default,description,status,segments_x
```
fneum: I'm fine with this, but `segments_x` now shows up in a lot of tables where it doesn't play a role. Were alternatives discussed?

test/test_lopf_piecewise_costs.py
```diff
@@ -0,0 +1,60 @@
...
+
+    n = pypsa.Network()
+    n.add("Bus", "bus0")
+    n.add("Generator", "gen0", bus="bus0", p_nom=100, marginal_cost=20)
+    n.add(
+        "Generator",
+        "gen1",
+        bus="bus0",
+        p_nom=100,
+        marginal_cost={0.1: 60, 0.5: 35.0, 1.0: 100.0},
```
fneum: Must this be a dictionary or can it also be a `pd.Series`? I guess that makes it harder to distinguish between `dynamic` and `segments`, but one could check if `pd.Index` aligns with `n.snapshots` and if not try to interpret as segments? Not sure.

test/test_lopf_piecewise_costs.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    n = pypsa.Network()
+    n.add("Bus", "bus0")
+    n.add(
+        "Generator",
+        "gen2",
+        bus="bus0",
+        p_nom_extendable=True,
+        p_nom_max=100,
+        capital_cost=pd.DataFrame({"p_nom": [0.0, 100.0], "capital_cost": [0.5, 1.5]}),
```
fneum: How would this `pd.DataFrame` approach work if multiple generators were added here? Is this supported?

pypsa/statistics/expressions.py
```diff
@@ -1670,7 +1696,7 @@ def func(n: Network, c: str, port: str) -> pd.Series:
                 ):
                     cost = n.get_switchable_as_dense(c, cost_type, inds=com_i)
                     var = n.c[c].dynamic[attr].loc[:, com_i]
-                    opex = var * cost
+                    opex = var * (cost + n.c[c].dynamic.get(f"{cost_type}_opt", 0))
```
fneum:
```suggestion
                    cost_segmented_opt = n.c[c].dynamic.get(f"{cost_type}_opt", 0)
                    opex = var * (cost + cost_segmented_opt)
```

pypsa/optimization/piecewise.py
```diff
@@ -0,0 +1,277 @@
...
+        If True, the integral of the piecewise curve will be used to define y breakpoints.
+        If False, the nominal values at each x breakpoint will be used to define y breakpoints.
+    y_var : linopy.Variable or None, optional
+        The optimisation variable for the y-axis of the piecewise constraint.
+        If None, a new auxiliary variable will be created and used instead.
+    extra_options : Iterable[PiecewiseOptions]
+        Extra options to pass to the piecewise constraint formulation.
+        If multiple instance of piecewise options are provided for the same component and attribute,
+        then multiple piecewise constraints will be created with each of the specified options.
+        If empty, default settings will be used.
```
fneum, line 112: What would be a use case for multiple instances of piecewise options?

pypsa/optimization/piecewise.py
```diff
@@ -0,0 +1,277 @@
...
+    invert_attr : bool, optional
+        Whether to invert the piecewise attribute values before multiplying by the x-axis variable.
+        If True, y_attr -> 1 / y_attr.
+        Default is False.
+    method : str, optional
+        The method to use for the piecewise constraint formulation, passed to linopy's add_piecewise_constraints method.
+        Default is "auto" (also the linopy default).
```
fneum, line 119: Could link to linopy docs what options are available?

fneum (changes requested):
Looks pretty good already! Had a few questions (some of which are TODO markers for later):

- I was a bit unsure about the `{capital,marginal}_cost_opt`. These are the resulting **average** cost per MW or MWh of the piecewise linear cost component, right?

- Second, the statistics and optimisation uses `capital_cost` + `capital_cost_opt` (or piecewise cost component). So, the piecewise component is additive to whatever is already user defined in `capital_cost`, right? Same for `marginal_cost`. This is different to the behaviour between `static` and `dynamic` values, where `dynamic` would overwrite any `static` values. This is fine to me, but must be explained carefully in the documentation to avoid confusion.

- What is the priority? `segments` > `dynamic` > `static`? Or is it difficult to say given `static` and `segments` are mixed?

- I was wondering why the `MultiIndex` was necessary for the columns of the segments DataFrame (i.e. why the x-coordinate couldn't just be the index). Is it because different components may have different number/positions of breakpoints?

- Did I understand correctly that if the marginal cost were decreasing with increasing use, there would be automatic reformulation to SOS/MILP?

- Shouldn't the index name of the segments DataFrame be `breakpoint` rather than `segment` because each row gives you a breakpoint?

- Is there a check that complains if an extendable component uses segments for efficiency or marginal cost?

- Test cases I would definitely include before merging: stochastic network, multi-horizon, global constraint.

- Do all IO variants (CSV, Excel, netCDF) work with the new segment DataFrames?

- For the final version, the new piecewise equations should be added to the user guide in the documentation for all variants (invest, efficiency, marginal cost; both tangent and SOS; cumulative and marginal; etc.)

pypsa/data/component_attrs/generators.csv
brynpickering: ~`average` should just go. Not sure why I put it there.~ OK, I _didn't_ put it there, it was your suggestion! I would say it is the final marginal cost for both since it has already progressed through the the other marginal cost rates to get to that point.

pypsa/data/component_attrs/buses.csv
brynpickering: No, we didn't discuss alternatives. Happy to implement differently. Perhaps in a `component_attrs/piecewise.csv` table?

test/test_lopf_piecewise_costs.py
brynpickering: It can be a dataframe (`segment` on the index, `marginal_cost` and `p_pu` on the columns). If it is an attribute that _can_ have piecewise attributes, this is currently considered first, then it is considered as a dynamic attr if it doesn't fit the expected format. I think some of this would be better solved by the validation work by @lkstrp

pypsa/optimization/piecewise.py
brynpickering: It allows different options per attribute and, within attributes, per component/name. E.g., If you have piecewise `capital_cost` you might want different components to use different piecewise formulations based on your knowledge of the most performant form of piecewise for the shape of the curve. And then you might also want to force marginal cost to be pinned to the curve, not be `ge`. So, you'd have:

```python
piecewise_options=[
    {"component": "Link", "attribute": "capital_cost", "method": "incremental"},
    {"component": "Generator", "attribute": "capital_cost", "method": "Disjunctive"},
    {"component": "Generator", "attribute": "marginal_cost", "operator": "=="}
]
```

I doubt there will be many instances of wanting to do this but from what @Irieo was already saying about the endogenous learning example, it seems worth leaving open to option for maximum customisability.

pypsa/optimization/piecewise.py
brynpickering: Will do once the linopy API + docs on this is stable

brynpickering:
@fneum 

> I was a bit unsure about the {capital,marginal}_cost_opt. These are the resulting average cost per MW or MWh of the piecewise linear cost component, right?

They are the marginal costs per MW or MWh linked to the optimal capacity / dispatch

> Second, the statistics and optimisation uses capital_cost + capital_cost_opt (or piecewise cost component). So, the piecewise component is additive to whatever is already user defined in capital_cost, right? Same for marginal_cost. This is different to the behaviour between static and dynamic values, where dynamic would overwrite any static values. This is fine to me, but must be explained carefully in the documentation to avoid confusion.

This is more of an expectation (once validation is sorted) that you cannot define both. So, it's safe for it to be additive since one will _always_ be zero.

> What is the priority? segments > dynamic > static? Or is it difficult to say given static and segments are mixed?

Yes, that's the priority currently. With validation this will be someone redundant as only one of the three should ever be defined without raising an exception.

> I was wondering why the MultiIndex was necessary for the columns of the segments DataFrame (i.e. why the x-coordinate couldn't just be the index). Is it because different components may have different number/positions of breakpoints?

Exactly, you can define any arbitrary number of breakpoints and at any points along the x-axis, so each entry needs to be separated. We discussed different forms (multi-index index vs multi-index columns, attrs as an index level) before settling on this one.

> Did I understand correctly that if the marginal cost were decreasing with increasing use, there would be automatic reformulation to SOS/MILP?

Correct.

> Shouldn't the index name of the segments DataFrame be breakpoint rather than segment because each row gives you a breakpoint?

I'd be inclined towards that, too. This would align better with linopy terminology. Open question is whether we should allow for segment definitions (which would allow disjoint definitions).

> Is there a check that complains if an extendable component uses segments for efficiency or marginal cost?

Yes.

> Test cases I would definitely include before merging: stochastic network, multi-horizon, global constraint.

Yup! There's one already for generator efficiency in a global constraint. Would you want one for a custom global constraint?

> Do all IO variants (CSV, Excel, netCDF) work with the new segment DataFrames?

🤷; I revert to @coroa on that.

> For the final version, the new piecewise equations should be added to the user guide in the documentation for all variants (invest, efficiency, marginal cost; both tangent and SOS; cumulative and marginal; etc.)

As more than just an example notebook, I assume? Can do.

@fneum perhaps you could also comment on my implementation detail questions in [this earlier comment](https://github.com/PyPSA/PyPSA/pull/1603#issuecomment-4353282808).

test/test_lopf_piecewise_costs.py
coroa: Yes, it then expects the columns to be a multiindex where the first level is the name and the second is the attribute:

```
          gen1            gen2            gen3
attribute p_pu efficiency p_pu efficiency p_pu efficiency
segment
0          0.0       0.00  0.0       0.00  0.0       0.00
1          0.5       0.33  0.5       0.33  0.5       0.33
2          1.0       0.50  1.0       0.50  1.0       0.50
```

coroa:
> > Second, the statistics and optimisation uses capital_cost + capital_cost_opt (or piecewise cost component). So, the piecewise component is additive to whatever is already user defined in capital_cost, right? Same for marginal_cost. This is different to the behaviour between static and dynamic values, where dynamic would overwrite any static values. This is fine to me, but must be explained carefully in the documentation to avoid confusion.
> 
> This is more of an expectation (once validation is sorted) that you cannot define both. So, it's safe for it to be additive since one will _always_ be zero.

I fear that is a strong assumption, but well depends on validation, indeed.

coroa:
> > Do all IO variants (CSV, Excel, netCDF) work with the new segment DataFrames?
> 
> 🤷; I revert to @coroa on that.

Uh, true, probably none do. I guess that is my job, i'll whip something up early next week.

pypsa/data/component_attrs/buses.csv
brynpickering: Updated to being in a `piecewise.csv` table, which does work much better IMO

pypsa/optimization/piecewise.py
brynpickering: Done.

test/test_lopf_piecewise_efficiency.py
```diff
@@ -44,3 +108,173 @@ def test_piecewise_efficiency_gen():
...
+        pd.testing.assert_frame_equal(n.c[comp].dynamic.p1, expected_p1)
+
+    @pytest.mark.xfail(
+        reason="rate0 formulation is not working correctly in the piecewise case"
+    )
+    @pytest.mark.parametrize("fixture_name", ["piecewise_two_port_process_network_r0"])
+    def test_piecewise_efficiency_two_port_in(self, request, fixture_name: str) -> None:
+        n, comp = request.getfixturevalue(fixture_name)
+        n.optimize(solver_name="gurobi")
+        # TODO: update expected when the issue with rate0 formulation is resolved.
```
fneum: Is that something that needs to be resolved before merging?

pypsa/optimization/constraints.py
```diff
@@ -1694,7 +1705,7 @@ def define_storage_unit_constraints(n: Network, sns: pd.Index) -> None:
     except ValueError:
         pass
 
-    # TODO-1603: grab segment efficiencies if available
+    # TODO-1603: grab piecewise efficiencies if available
```
fneum: Piecewise efficiencies are not supported for storage units now. Would you say that is a blocker for this PR or should we proceed without it?

It's relatively clearly documented that this combination is not implemented yet.

pypsa/optimization/expressions.py
```diff
@@ -547,14 +563,14 @@ def func(n: Network, c: str, port: str) -> pd.Series:
             elif direction != "both":
                 msg = f"Got unexpected argument direction={direction}. Must be 'supply', 'withdrawal' or 'both'."
                 raise ValueError(msg)
-            # TODO-1603: work out sign of seg_var depending on direction
-            segment_efficiency = port_efficiency(n, c, port=port, segment=True)
-            if isinstance(segment_efficiency, pd.DataFrame):
-                seg_var = n.model.variables[f"{c}-p{port}_piecewise"]
-                coeffs.loc[{"name": seg_var.coords["name"]}] = 0
+            # TODO-1603: work out sign of pw_var depending on direction
```
fneum: To be resolved. Should be straightforward to figure out, shouldn't it?

docs/user-guide/optimization/piecewise.md
fneum, line 1: The description of the piecewise linearisation is great. But one thing I would really like to see is more mathematical explanation of new variables, objective terms and (auxiliary) constraints added to the model (incl. mapping between component attributes, math formulation and model variable names). Perhaps the style of https://docs.pypsa.org/latest/user-guide/optimization/pathway-planning/ is a good blue print.


Can also link to piecewise example at the bottom.


This file needs to be integrated into docs. Perhaps under "Contingencies"?


I would drop " in PyPSA" in titles.

The doc builds are failing.

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Piecewise linearisation
+
+If you have non-linear relationships between two decision variables, you may want to represent them as a [piecewise linear function](https://en.wikipedia.org/wiki/Piecewise_linear_function).
+Common non-linear relationships found in energy systems include:
+
+- **marginal cost curves**: Where the marginal cost of generating a unit of energy changes as the quantity of energy generated changes.
+- **capital cost curves**: Where the capital cost of investing in a unit of nominal capacity changes as the quantity nominal capacity changes.
```
fneum:
```suggestion
- **capital cost curves**: Where the capital cost of investing in the next unit of nominal capacity changes as the quantity nominal capacity changes.
```

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+
+!!! question "What is a _breakpoint_?"
+
+    A breakpoint is a point along the piecewise curve where two straight lines of different gradients meet.
+    We define breakpoints in terms of integer values (breakpoint 0, 1, 2, etc.).
+
+## Inspecting piecewise data in PyPSA
+
+Piecewise data is stored in a component `piecewise` dictionary-like object, much like `dynamic` data.
+For instance, to access marginal_cost piecewise data for generators, you can call `n.c.generators.piecewise.marginal_cost`.
```
fneum: Right now, the documentation is written in the current API, in which case we have `n.generators_t.marginal_cost`. Is there a n.generators_pw.marginal_cost`? It might require some coordination so that it is not confusing for people working with current default API. @lkstrp

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+    This should be fine in most cases; the optimisation is cost-minimising, so you will not actually get costs that deviate from the piecewise curve.
+
+    If you want to pin piecewise costs to the curve, you need to use an equality sign.
+    Assuming we are applying this to marginal costs of generators:
+
+    ```python
+    n.optimize(piecewise_options=[{"component": "Generator", "attribute": "marginal_cost", "sign": "=="}])
+    ```
+
+## Merit order vs. spot curves
```
fneum: Maybe "cumulative curves" and "pointwise curves" or "direct curves"?

Merit order is perhaps too narrow for all possible use cases of PWL cost.

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Merit order curves
+
+Here, one PyPSA component might represent a fleet of assets (e.g. a single `CCGT` generator represents 100 real assets in the given zone).
+If the assets have different marginal costs, you would want to operate them from least to most expensive as you increase total generation.
+In our example, a 50% load rate would mean the 50 least expensive assets are operating at 100% load since they incur the least cost in operation.
+
+The piecewise curve in this instance is based on combining the increasing marginal costs of the fleet in increasing order.
+We start by paying the cheaper costs and then build up to paying the most expensive cost.
+The marginal operating cost at the optimal dispatch rate _is_ the marginal cost of the generator, but the total operating cost is the integral of operating costs along the piecewise curve up to that point.
```
fneum: Alternative explanation (feel free to mix in):

Here the y-value at each point describes an *incremental* contribution, so the quantity accumulates along the curve. The total at operating point $X$ is the integral of the curve up to $X$.

**Example:** For an increasing marginal cost curve, the increments are dispatched in cost order. For example, a CCGT generator might combine many units with different marginal costs, so at a 50% load rate the cheapest half run at full output and the rest stay off, since that is the least-cost way. The value at $X$ is then the marginal cost of the generator at that dispatch, the cost of the last increment, while the total operating cost is the integral along the curve up to $X$. The realised per-unit cost is the average of all increments, below the value at $X$ where the curve rises and above it where it falls.

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+### Non-merit order curves
+
+Here, one PyPSA component might represent a single asset with a non-linear operating characteristic.
+Unlike merit order curves, we do not care about the nature of the curve except at the point we are operating.
+So, at a 50% load rate, this asset is operating at 50% of its nameplate capacity and it doesn't matter what the piecewise curve says about its characteristics at 10%, 25%, etc. load rate.
+
+Therefore, the piecewise curve in this instance is _not_ integrated up to the operating point; instead, it is taken at face value at each operating point.
+
+We assume that part-load efficiencies are not merit order curves.
+It could be reasonable to also assign this curve type to marginal operating or capital costs where those curves describe costs for a single asset.
```
fneum:
```suggestion
We assume that part-load efficiency curves are direct. A cost curve can also be direct where its values are meant to be read at the operating point rather than accumulated.
```

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,84 @@
...
+
+We assume that piecewise marginal operating and capital cost curves are merit order curves.
+
+### Non-merit order curves
+
+Here, one PyPSA component might represent a single asset with a non-linear operating characteristic.
+Unlike merit order curves, we do not care about the nature of the curve except at the point we are operating.
+So, at a 50% load rate, this asset is operating at 50% of its nameplate capacity and it doesn't matter what the piecewise curve says about its characteristics at 10%, 25%, etc. load rate.
+
+Therefore, the piecewise curve in this instance is _not_ integrated up to the operating point; instead, it is taken at face value at each operating point.
```
fneum: Alternative explanation (feel free to mix in):

Here the curve is read at the operating point and no integration takes place. The value at the operating rate $X$ is the realised value directly, evaluated point by point rather than accumulated from the rates below it.

**Example:** Consider a component with a part-load efficiency curve: at a 50% load rate the asset operates at 50% of its nameplate capacity and the efficiency is the curve value at 50%, regardless of what it says at load rates of 10%, 25%, and so on. If a direct curve is used for a cost instead, the total is the value at $X$ multiplied by the full dispatch $X$, so that value acts as a per-unit cost applied across all dispatch rather than as the cost of the last increment.

fneum (changes requested):
Code-wise everything looks good to me and there are heaps of tests. 👍 

As for your Qs:

> How do we handle the first y value not being 0% and the final y value not being 100% (both effectively apply a p_<in/max>_pu to the component)?

Could be fine the way it is, applying p_min/max_pu implicitly

> Do we force covering the full range?

Only if someone says that the implicit p_min/max_pu is not good. Then yes.

> Do we move from defining per unit to absolute x values for dispatch piecewise constraints?

I would say no because this gets more difficult for extendable components.

> Why is applying piecewise curve to process rate0 leading to infeasibilities (see xfailing test)?

Do you have some IIS to look at?

> How to document marginal vs non-marginal piecewise curves? I've made an attempt but I'm not sure it hits the mark.

See suggestions below.

> Do we want piecewise outputs in the static columns by default (e.g. capital_cost_piecewise_opt), given that dynamic piecewise attrs only appear when they are needed (e.g. marginal_cost_piecewise_opt)?

For now, yes. @lkstrp is working on making optional attributes appear dynamically when used separately.

> Should p1...N always be negative for multiports in the outputs?

Depends on the efficiency, but the convention in PyPSA is p{i} > 0 if withdrawing from bus i and p{i} < 0 if injecting into bus i.

FabianHofmann:
@brynpickering the xfail test was falling into infeasibility due to lower bounds in the y aux variables preventing it to go negative. For withdrawing rates (<0) this is however required. I also uplifted the capacity of the generators in the base_network to supply all demands (last timestep surpassed 100MW)

FabianHofmann:
compat issue with linopy v0.8 (current reason for test failures) is solved here https://github.com/PyPSA/PyPSA/pull/1722

pypsa/optimization/expressions.py
brynpickering: I think so, yeah, assuming that `p{port}` is equivalent to `coeff[port] * p`, so Process `p0` will be negative, because `rate0` is negative. So `p0_piecewise` will be negative. So, we an apply the same `sign` multiplication and `withdrawal` negation as is done for non-piecewise expressions here.

docs/user-guide/optimization/piecewise.md
brynpickering: This is complicated by the fact that the mathematical explanation (aux vars & constrs) is different depending on which piecewise formulation is used. These are also not within PyPSA's control, so the docs could go out of date easily if anything about how piecewise constraints are stored is changed in linopy.

docs/user-guide/optimization/piecewise.md
brynpickering: Aha. There isn't a `n.generators_pw` or similar attribute. I can add it.

brynpickering: Actually, since `processes_t` wasn't added, I don't think we should be introducing `generators_pw` or similar at this stage. I would just use the new API terminology in the docs.

docs/user-guide/optimization/piecewise.md
brynpickering: Updated to align with your alternative

docs/user-guide/optimization/piecewise.md
brynpickering: Updated to align with your alternative

docs/user-guide/optimization/piecewise.md
brynpickering: Updated to your suggestion

pypsa/optimization/constraints.py
brynpickering: No, have removed this.

pypsa/optimization/expressions.py
brynpickering: Resolved (but highlighted a bug with how we use latest linopy in our expression generation!) in #1739

brynpickering:
note: tests are passing locally (incl. after a `uv sync`). Not sure why the CI tests are failing.

FabianHofmann (approved):

FabianHofmann:
this is finalized from our side. @fneum, since you requested changes, do you want to take another look or defer to @lkstrp ?

fneum:
Yes, I'd like to take a final look and try for today together with maintenance scheduling.

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,124 @@
...
+-->
+
+# Piecewise linearisation
+
+If you have non-linear relationships between two decision variables, you may want to represent them as a [piecewise linear function](https://en.wikipedia.org/wiki/Piecewise_linear_function).
+Common non-linear relationships found in energy systems include:
+
+- **marginal cost curves**: Where the marginal cost of generating a unit of energy changes as the quantity of energy generated changes.
+- **capital cost curves**: Where the capital cost of investing in the next unit of nominal capacity changes as the quantity nominal capacity changes.
+- **part-load efficiency curves**: Where the efficiency of energy generation / conversion changes as the load rate ($\frac{\text{generation}{\text{nominal capacity}}}$) changes.
```
fneum:
```suggestion
- **part-load efficiency curves**: Where the efficiency of energy generation / conversion changes as the load rate ($\frac{\text{generation}}{\text{nominal capacity}}$) changes.
```

docs/user-guide/optimization/piecewise.md
```diff
@@ -0,0 +1,124 @@
...
+
+!!! question "What is a _breakpoint_?"
+
+    A breakpoint is a point along the piecewise curve where two straight lines of different gradients meet.
+    We define breakpoints in terms of integer values (breakpoint 0, 1, 2, etc.).
+
+## Inspecting piecewise data
+
+Piecewise data is stored in a component `piecewise` dictionary-like object, much like `dynamic` data.
+For instance, to access marginal_cost piecewise data for generators, you can call `n.c.generators.piecewise.marginal_cost`.
```
fneum:
```suggestion
For instance, to access `marginal_cost` piecewise data for generators, you can call `n.c.generators.piecewise.marginal_cost`.
```

fneum (approved):
I mostly checked functionality, documentation, examples and API. I quite like how it works and support merging!

As for the code, I'm happy if either @lkstrp or @FabianHofmann confirm their approval on the technical side. It's quite many code changes/additions (I couldn't quite review all of it comprehensively), but perhaps that's just necessary to get the PWL in.

docs/api/networks/network.md
```diff
@@ -6,6 +6,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ::: pypsa.Network
     options:
+        inherited_members: true
```
lkstrp: This is missing on purpose since all the Mixins are showing their methods on own docs pages. Now we just have them duplicated

docs/release-notes.md
```diff
@@ -22,7 +22,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - Fixed spurious infeasibility in [`optimize_with_rolling_horizon()`][pypsa.optimization.OptimizationAccessor.optimize_with_rolling_horizon] when a network mixed committable and non-committable generators with ramp limits. At a window seam, non-committable components (which carry no commitment status) were assigned `status=0`, corrupting their start-up/shut-down ramp terms. (<!-- md:pr 1644 -->)
```
lkstrp, line 34: A release note for this banger is missing

pypsa/components/_types/mixin/multiports.py
```diff
@@ -58,6 +59,41 @@ def additional_ports(self) -> list[str]:
...
+    def _piecewise_attrs(self) -> pd.DataFrame:
+        """Get the piecewise attributes for this component.
+
+        Returns
+        -------
+        pd.DataFrame
+            DataFrame with piecewise attributes for this component.
+
+        """
+        filtered_attrs = PIECEWISE_ATTRS.query(
```
lkstrp: Get rid of query, see comment in `network/transform.py`

Code can probably also be reduced a fair bit

pypsa/components/components.py
```diff
@@ -376,7 +398,66 @@ def _get_data_containers(ct: ComponentType) -> tuple[pd.DataFrame, Dict]:
             df.columns.name = "name"
             dynamic[k] = df
 
-        return static, dynamic
+        # Piecewise breakpoint data: one empty DataFrame per piecewise attribute
+        # defined by the piecewise_x column in the component's attribute CSV.
+        piecewise = Dict()
+        for y_attr in PIECEWISE_ATTRS.query(
```
lkstrp: Get rid of query, see comment in `network/transform.py`

pypsa/components/components.py
```diff
@@ -376,7 +398,66 @@ def _get_data_containers(ct: ComponentType) -> tuple[pd.DataFrame, Dict]:
...
+    def _piecewise_attrs(self) -> pd.DataFrame:
+        """Get the piecewise attributes for this component.
+
+        Returns
+        -------
+        pd.DataFrame
+            DataFrame with piecewise attributes for this component.
+
+        """
+        filtered_attrs = PIECEWISE_ATTRS.query(
```
lkstrp: Get rid of query, see comment in `network/transform.py`

pypsa/components/components.py
```diff
@@ -376,7 +398,66 @@ def _get_data_containers(ct: ComponentType) -> tuple[pd.DataFrame, Dict]:
...
+
+    def has_piecewise(self, attr: str) -> bool:
+        """Whether this component instance has piecewise breakpoint data for `attr`."""
+        pw_df = self.piecewise.get(attr)
+        return pw_df is not None and not pw_df.empty
+
+    def _piecewise_schema(self, **attrs: str) -> pd.Series:
+        """Get the schema row of the piecewise definition for `attr` (empty if undefined)."""
+        query = " and ".join(f"{k} == '{v}'" for k, v in attrs.items())
+        series = self._piecewise_attrs.query(query).squeeze()
```
lkstrp: Get rid of query, see comment in `network/transform.py`

pypsa/data/component_attrs/processes.csv
```diff
@@ -55,3 +55,4 @@ mu_upper,series,currency/MW,n/a,Shadow price of upper `p_nom` limit $f \\leq F$.
 mu_p_set,series,currency/MWh,n/a,Shadow price of fixed dispatch `p_set`,Output
 mu_ramp_limit_up,series,currency/MWh,n/a,Shadow price of upper ramp up limit,Output
 mu_ramp_limit_down,series,currency/MWh,n/a,Shadow price of lower ramp down limit,Output
+p_piecewise,series,MWh,0.,Optimised dispatch at remote bus.,Output
```
lkstrp: I think alignment with links was missed here, the result is never written back since the output container is not defined

```
import pandas as pd
import pypsa

n = pypsa.Network()
n.add("Bus", ["b0", "b1"])
n.add("Generator", "src", bus="b0", p_nom=100)
n.add("Load", "load", bus="b1", p_set=50.0)
n.add(
    "Process",
    "proc",
    bus0="b0",
    bus1="b1",
    rate0=-1,
    rate1=1,
    p_nom_extendable=True,
    p_nom_max=100,
    capital_cost=pd.DataFrame({"p_nom": [0.0, 100.0], "capital_cost": [0.5, 1.5]}),
)
n.optimize()

print("objective:", n.objective)  # 75.0
print("stats.capex():", n.stats.capex().sum())  # 0.0
```

This also needs a test

pypsa/network/io.py
```diff
@@ -1989,6 +2132,118 @@ def _import_series_from_df(
...
+            if None, it is inferred from the component's extendables.
+        overwrite : bool, default False
+            If True, replace existing breakpoint data for the component.
+
+        """
+        idx_name = "breakpoint"
+        col_names = ["name", "attribute"]
+        c = self.c[cls_name]
+        search_attr = c._get_base_coeff(attr) if isinstance(c, _Multiport) else attr
+        pw_attr = PIECEWISE_ATTRS.query(
```
lkstrp: Get rid of query, see comment in `network/transform.py`

pypsa/network/io.py
```diff
@@ -1989,6 +2132,118 @@ def _import_series_from_df(
...
+            "component == @name and y == @search_attr",
+            local_dict={"name": cls_name, "search_attr": search_attr},
+        ).squeeze()
+        if pw_attr.empty:
+            valid_attrs = (
+                PIECEWISE_ATTRS.query("component == @cls_name").y.unique().tolist()
```
lkstrp: Get rid of query, see comment in `network/transform.py`

pypsa/network/transform.py
```diff
@@ -267,22 +271,80 @@ def add(
...
-            # If index/ columnes are passed (pd.DataFrame or pd.Series)
+            # Enable default lookup for attributes with port suffixes, e.g. efficiency1, rate2.
+            search_attr = c._get_base_coeff(k) if isinstance(c, _Multiport) else k
+
+            # Intercept piecewise breakpoint data: a DataFrame whose columns are the curve
+            # attributes (e.g. ["p_pu", "efficiency"]), not component names.
+            # These are identified by the attribute name being in piecewise_attrs and
+            # the value being a DataFrame whose columns include the x-axis attribute.
+            if not (
+                pw_attr := PIECEWISE_ATTRS.query(
```
lkstrp: This query leads to roughly twice the runtime of `n.add`, even for non piecewise adds. This should just be a dictionary lookup instead of a query. The dict can be created in `constants.py`

@brynpickering I know that `df.query` can be elegant and that it has many fans (e.g. @fneum), but aside from being slow here, it is generally not a good idea to use it. As the [docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html) say:
> This method can run arbitrary code which can make you vulnerable to code injection if you pass user input to this function.

Here the input is not user data but comes from the package. But it is generally easy to lose track of that distinction. We also still use `query` somewhere in the statistics module.

I am not sure what you are using for the text based YAML formulation feature, but just to keep that in mind. Mitigation is quite easy with stuff like `simpleeval` or `ast.parse`

pypsa/network/transform.py
```diff
@@ -267,22 +271,80 @@ def add(
...
+                        v.columns.names = ["name", "attribute"]
+                        v.index.name = "breakpoint"
+                        piecewise[k] = v
+                        continue
+            elif isinstance(v, dict):
+                msg = (
+                    "Dictionaries are not supported as dynamic attribute values. "
+                    "Please use pandas.Series or pandas.DataFrame instead."
+                )
+                raise NotImplementedError(msg)
```
lkstrp:
```suggestion
                raise TypeError(msg)
```

pypsa/optimization/optimize.py
```diff
@@ -961,33 +1081,15 @@ def assign_solution(self) -> None:
...
-
                 else:
                     _set_dynamic_data(n, c.name, attr, df)
             # Ignore `n_mod`
             elif attr == "n_mod":
                 pass
             else:
-                c.static.update(df.rename(attr + "_opt"), overwrite=True)
-
+                c.static.update(
```
lkstrp, line 1113: This silently drops piecewise output, see comment above

pypsa/optimization/piecewise.py
```diff
@@ -0,0 +1,322 @@
...
+                f"Problematic components: {bad_entries}."
+            )
+            raise ValueError(msg)
+        x_da = x_da * nom_attr_da
+    x_da = x_da.where(valid_breakpoints)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=EvolvingAPIWarning)
+        x_breakpoints = breakpoints(x_da)
+        if cumulative_attr:
+            slopes = y_da.shift({BREAKPOINT_DIM: -1})
```
lkstrp, line 263: Isn't this just dropping the first breakpoint of our piecewise curve?

EDIT:
Yes, or is this a expected users problem fault? I would still expect this to make a difference/ give any feedback. It's nowhere mentioned in the docs
``` python
import pypsa
objectives = []
for first_y in [40, 100]:
    n = pypsa.Network()
    n.add("Bus", "b")
    n.add("Load", "load", bus="b", p_set=25.0)
    n.add("Generator", "gen", bus="b", p_nom=100,
          marginal_cost={0.0: first_y, 0.5: 60.0, 1.0: 70.0})
    n.optimize()
    objectives.append(n.objective)

assert objectives == [1500.0, 1500.0]
print("first y-value is silently ignored")
```

lkstrp (changes requested):
Fantastic feature! As discussed, very rough to stuff into the current PyPSA data model. We want to converge to something more simple and reliable

I am not fully done yet, but here is what I found so far

pypsa/network/transform.py
brynpickering: @lkstrp I only used `query` as it seems well loved within pypsa projects. I had not used it before last year (and the text math does not use it at all - there I use pyparsing). I do like its elegance but also understand the concerns. However, since pandas runs the text through its own AST parser I'd say it's much safer than a standard `eval` (i.e., I expect it's about as safe as `ast.parse`). 

The runtime hit is the biggest issue. I hadn't noticed that. We should definitely revert to more verbose standard pandas df filtering.

docs/api/networks/network.md
brynpickering: That's fine - as you wish! I didn't find it at all intuitive though when writing a cross-ref to the API elsewhere in the docs (the user never sees the mixins so it makes sense to have the API docs be really clear about the public methods in public classes)

pypsa/data/component_attrs/processes.csv
brynpickering: I actually don't think it's needed, so should just be removed from here. We pass the values from the piecewise variable to `p{N}` when assigning the solution

pypsa/optimization/piecewise.py
brynpickering: This is a general issue we haven't resolved: what to do when the `y` value at `x=0` is non-zero. At the moment, it will _always_ be ignored. Here, it is shifted away and then reintroduced with `y0=0` in the next line. In the `else` block it is implicitly deleted since we multiple `y` and `x` values so whatever you have for `y` will go to zero. 

I raised this issue in the main PR description. It requires some separate flag at addition to tell pypsa to create a special, binary constraint for `x=0` (same as for #1482).

I think my preference right now would be to validate the data here and raise a warning if the first breakpoint isn't `{0.0: 0.0}`. 
We could also set `y0` to `first_y` in `Slopes` but I don't know whether that is suitable when `x0` is not zero...

brynpickering: Alright, so I investigated this some more.

The way that the cumulative ones work is that you define `y` as `cost/unit` from `x-1` to `x`. So, if you have marginal cost of `{0.0: 10.0, 0.5: 60.0, 1.0: 70.0})` for a generator with a `p_nom=100`, it is 10EUR/MW @ 0 - 0MW, 60EUR/MW @ 0 - 50MW and 70EUR/MW @ 50-100MW, creating the `(x, y)` piecewise breakpoints of `(0, 0)`, `(50, 3000)`, and `(100, 6500)` (60*50 + 70*50). 

Basically, if you have an initial breakpoint of zero then the `y` value must _always_ be ignored since the final piecewise curve always represents `x * y`. 

If your first breakpoint is >0, that's a different matter and is more complex. If the first x-point is `0.1`, does that mean it is 10EUR/MW @ 0 - 10MW or 10EUR/MW @ 10 - 10MW? The former is implicit and we'd need to account for it as we don't currently. The latter is effectively the same as the first x-point being `0.0` since it's only true at that one point, which means it takes zero area under the curve. Not sure what the best convention to impose here is.

docs/api/networks/network.md
lkstrp: > I didn't find it at all intuitive though when writing a cross-ref to the API elsewhere in the docs

I do agree. But this would rather need to be cleaned up in an extra just docs PR. We just don't wanna have a single list of all methods which would be massive, and the mixins are a nice way to keep it seperated without hardcoding methods. But probably there is a better way

pypsa/network/transform.py
lkstrp: It's not as bad as `eval` because it uses `ast.parse` only to filter the query. But the whitelist pandas uses is not very strict, which in the worst case, makes it equivalent to `eval`

pypsa/optimization/piecewise.py
lkstrp: > If your first breakpoint is >0, that's a different matter and is more complex. If the first x-point is 0.1, does that mean it is 10EUR/MW @ 0 - 10MW or 10EUR/MW @ 10 - 10MW? The former is implicit and we'd need to account for it as we don't currently. The latter is effectively the same as the first x-point being 0.0 since it's only true at that one point, which means it takes zero area under the curve. Not sure what the best convention to impose here is.

The former is implicit, which I would argue against. The user should just pass `{0.0: 0.0, 0.1: 10, ...}`. The latter is much worse isn't it? There is no width, so it's the same then just starting at 0. But the linopy formulation also bounds x to the curves range. Which would mean `0.1: 10` is effectively not just cost, but a must run at 10% constraint. 

Which means I would tend to just straight up forbid a first x-point which is `>0`, and warn if if `y>0` for `x=0`. I'll added that as the last commit


``` python
import linopy
from linopy import breakpoints

m = linopy.Model()
x = m.add_variables(lower=0, upper=100, name="x")
y = m.add_variables(name="y")
m.add_piecewise_formulation(
    (y, breakpoints([0.0, 1500.0, 5000.0]), "="),
    (x, breakpoints([25.0, 50.0, 100.0])),
    name="pwl",
)
m.add_objective(1 * x)
m.solve()
assert float(x.solution) == 25
```

lkstrp (approved):
I applied the review and did some refactoring. See comments

@brynpickering Happy for you to merge, if you agree with everything

brynpickering:
@lkstrp I refactored a bit to move these data checks much earlier in the process: to where the data is added rather than where the constraint is built. Fast failure should be preferable! I've also extended the errors/warnings to both cumulative and non-cumulative cases, since both are at risk of bounding the problem in a way the user doesn't expect. 

Final point, I became aware that my representation of the cumulative piecewise curves in the example notebook was incorrect. It should be considered a step-wise curve, with each breakpoint defined by the user defining a new step of the normalised data (EUR/MW etc.). It highlights a much more drastic deviation in the curve definition between the continuous curve and the piecewise one but I've double-checked it and it does seem to be correct once denormalised.
Screenshot 2026-08-04 at 09 00 48


If you're happy with it all then I'll go ahead and merge.

lkstrp:
Thanks for both! Let's do it 💃

LukasFrankenQ:
@brynpickering thank you for this so much, very needed, super happy I wont need to fix this locally in the future :pray:

coroa:
Congrats

PR #1614, opened 2026-03-10
Fix market_value and get_operation for multi-port components

FabianHofmann:
Closes #971

**Problem:** `market_value` used identical `at_port` kwargs for both revenue and supply, so for multi-links (e.g., electrolysis with bus0=electricity, bus1=hydrogen), only port 0 was considered — missing revenue from other ports entirely. Additionally, `get_operation` returned `p0` for all branch components, but for Processes `p ≠ p0` (since `p0 = -p * rate0`).

- **Fix `get_operation`**:  Links and Processes now correctly return `dynamic.p` (the reference variable) instead of `dynamic.p0`. Includes a fallback for legacy networks where `p` may be empty. 

- **Fix `market_value`**: Replaced previous `revenue / supply` to divide by the reference operational variable (via `get_operation`) instead. This gives units of `currency/operation` of the component's reference carrier. If a `bus_carrier` is selected, the function returns the contribution to market value from that bus carrier. Market values from different bus carriers can be summed up as they no longer depend on different units. The entity hold that all contributions from bus carriers yields the resutls from `bus_carrier=None`.  

## Checklist

### DIFF
--- a/pypsa/statistics/expressions.py
+++ b/pypsa/statistics/expressions.py
@@ -2680,8 +2680,7 @@ class StatisticsAccessor(AbstractStatisticsAccessor):
         variable.
 
         The market value is always calculated relative to the component's
-        reference operational variable from
-        [get_operation][pypsa.statistics.expressions.get_operation]. Filters such as
+        reference operational variable from `get_operation`. Filters such as
         `bus_carrier` and `at_port` only restrict the revenue contribution in the
         numerator.
 


test/test_statistics.py
```diff
@@ -434,3 +434,144 @@ def test_energy_balance_carrier_nice_name_filter(network_with_nice_name):
         carrier="residential rural heat", nice_names=False
     )
     assert result.empty
+
+
+@pytest.fixture
+def multiport_network():
```
brynpickering: Can you add a third port to this example to better catch possible edge cases

test/test_statistics.py
```diff
@@ -434,3 +434,144 @@ def test_energy_balance_carrier_nice_name_filter(network_with_nice_name):
...
+    n.add("Load", "h2_load", bus="h2", carrier="H2", p_set=50)
+
+    def _df(data):
+        df = pd.DataFrame(data, index=n.snapshots)
+        df.columns.name = "name"
+        return df
+
+    p = _df({"electrolysis": [80.0, 60.0, 70.0]})
+    n.c.links.dynamic["p"] = p
+    n.c.links.dynamic["p0"] = p.copy()
```
brynpickering: This is because the reference port is set as `p0`, right? Probably need an example where the reference port is `p1` (is that currently possible? I don't know where the other PR on multiports is at)

test/test_statistics.py
```diff
@@ -434,3 +434,144 @@ def test_energy_balance_carrier_nice_name_filter(network_with_nice_name):
...
+    pd.testing.assert_frame_equal(get_operation(n, "Process"), p)
+
+
+def test_market_value_multiport(multiport_network):
+    n = multiport_network
+    mv = n.statistics.market_value(nice_names=False, round=None, drop_zero=False)
+
+    p = np.array([80.0, 60.0, 70.0])
+    price_elec = np.array([50.0, 40.0, 45.0])
+    price_h2 = np.array([100.0, 90.0, 95.0])
```
brynpickering: set these as fixtures and use them each time they're needed in each test/fixture, so that it's easier to track that they are meant to be the same value and to make it easier to maintain these tests in future

test/test_statistics.py
brynpickering: Maybe add a third port and a different reference port in a separate, otherwise simpler fixture and add separate tests for each

pypsa/statistics/expressions.py
```diff
@@ -2697,7 +2730,37 @@ def market_value(  # noqa: D417
...
+                carrier=carrier,
+                bus_carrier=None,
+                nice_names=nice_names,
+                drop_zero=False,
+                round=None,
+            )
+            df = rev / denom.abs()
+
         df.attrs["name"] = "Market Value"
         df.attrs["unit"] = "currency / MWh"
```
brynpickering: this isn't per MWh in cases where the reference port has a different unit, right? Do we have access to that metadata at all?

brynpickering: Since we're already generalising currency, you could generalise this to `energy` or `reference_flow_sum` or something like that?

pypsa/statistics/expressions.py
```diff
@@ -32,12 +32,35 @@
...
+    pd.DataFrame
+        Time series of the reference operational variable for the component.
+
+    """
+    if c in n.passive_branch_components:
         return n.c[c].dynamic.p0
     if c == "Store":
         return n.c[c].dynamic.e
-    return n.c[c].dynamic.p
+    p = n.c[c].dynamic.p
```
brynpickering, line 60: Does this call pass in legacy networks (i.e., will accessing `p` succeed or will it raise a KeyError?)

pypsa/statistics/expressions.py
```diff
@@ -2697,7 +2730,37 @@ def market_value(  # noqa: D417
...
-        df = self.revenue(**kwargs) / self.supply(**kwargs)
+
+        if bus_carrier is not None:
+            df = self.revenue(**kwargs) / self.energy_balance(**kwargs).abs()
+        else:
+            revenue_kwargs = {**kwargs, "at_port": "all"}
+            rev = self.revenue(**revenue_kwargs)
+
+            @pass_empty_series_if_keyerror
+            def ref_func(n: Network, c: str, port: str) -> pd.Series:
```
brynpickering: All others instances in this class use `func`, not `ref_func` for this method name

pypsa/statistics/expressions.py
```diff
@@ -2697,7 +2730,37 @@ def market_value(  # noqa: D417
...
+                weights = n.snapshot_weightings.generators
+                return self._aggregate_timeseries(p, weights, agg=groupby_time)
+
+            denom = self._aggregate_components(
+                ref_func,
+                components=components,
+                agg=groupby_method,
+                aggregate_across_components=aggregate_across_components,
+                groupby=groupby,
+                at_port=[0],
```
brynpickering: ok, maybe this answers my question in another comment, that the reference port is zero by default. No scope for user-defined reference port (yet)

pypsa/statistics/expressions.py
```diff
@@ -2621,8 +2644,18 @@ def market_value(  # noqa: D417
...
+        Currency is currency/MWh or currency/unit_{bus_carrier} where
+        unit_{bus_carrier} is the unit of the bus carrier.
+
+        The behavior depends on whether `bus_carrier` is specified:
+
+        - **Default (no `bus_carrier`)**: Returns total revenue across all ports
+          divided by the reference operational variable. Units are
+          `currency/MWh` of the component's reference carrier (e.g., €/MWh_el
+          for electrolysis consuming electricity at bus0).
+        - **With `bus_carrier`**: Returns revenue at the specified bus carrier's
```
brynpickering:
```suggestion
        - **With `bus_carrier`**: Returns revenue at the specified bus carriers'
```

pypsa/statistics/expressions.py
brynpickering, line 2749: A very high quality example /s

pypsa/statistics/expressions.py
```diff
@@ -2697,7 +2730,37 @@ def market_value(  # noqa: D417
...
+            @pass_empty_series_if_keyerror
+            def ref_func(n: Network, c: str, port: str) -> pd.Series:
+                sign = (
+                    -1.0 if c in n.branch_components else n.c[c].static.get("sign", 1.0)
+                )
+                p = (sign * get_operation(n, c)).abs()
+                weights = n.snapshot_weightings.generators
+                return self._aggregate_timeseries(p, weights, agg=groupby_time)
+
+            denom = self._aggregate_components(
```
brynpickering: Why not just call `self.energy_balance` with a kwarg update to limit the port?

pypsa/statistics/expressions.py
brynpickering: As it is, any other `kwargs` supplied be the user effectively get ignored here, whereas they are being accounted for when `bus_carrier` is not None.

test/test_statistics.py
FabianHofmann: exactly, it is not possible for links though, the process component is the one which is more flexible in that regard. that's perhaps a good, point. I can change the entire test setup to use a process with ref bus at `bus1`

pypsa/statistics/expressions.py
FabianHofmann: yes, this one would work. pypsa auto-creates all templated attributes. since this one was recently introduced, the field will be there but empty for legacy networks. But I guess you raise a good point and I think we should address this in the IO directly, overwriting setting `p` to `p0` if it remains empty from the IO data. @lkstrp I guess you have an opinion about this as well?

lkstrp: Yes, fine for now. We have a couple of those legacy migration steps (not many though), which I am moving to IO. Afterwards this should be a strict `n.links_t.p == n.links_t.p0` condition without the need to handle it in the logic

pypsa/statistics/expressions.py
FabianHofmann: indeed :D

FabianHofmann:
thanks for the reveiw @brynpickering  

I have addressed the comments about the tests. it is now using a process component with three ports. As you love fixtures and seem to be experienced with that I would take one more iteration on this. I think I am not fully there yet. happy if you take another look. 


About the denominator. The idea is to make the market value always referencing the **operation** of the asset independent of the `bus_carrier` selection. The operation is defined as `p` for links and processes where `p == p0` for links and `p == p{ref_port}` for processes. Indeed the unit of that operation remains not strictly defined in pypsa, this is a bigger issue. 

The advantage of that choice is that market values from different bus_carriers can be summed for the same components as the denominator stays the same. In pseudo-code:

```
market_value() == sum( market_value(bus_carrier=c) for c in bus_carriers) )   (for all individual assets)
```

this would not be the case if we chose energy_balance in the denominator. 

The `func` in the demoninator function ignores the `port` argument, the `at_port=[0]` in the outer function is essentially just a placeholder to trigger the loop over the `func` in the internal (only ones). 


Having said all that and looking at the code again, it still using energy_balance for `bus_carrier=None`. I will "correct" this to using the operation as well.  

does this make sense?

FabianHofmann:
okay, everything I now based on using the internal operation as a reference. docstrings and pr description were updated accordingly. if you want to have another look @brynpickering

docs/release-notes.md
```diff
@@ -35,6 +35,11 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ### Bug Fixes
 
+- Fix `get_operation` to correctly return the reference operational variable (`p`) for Link and Process components instead of `p0`.
+
+- Fix `market_value` to correctly handle multi-port components (Links, Processes) by always normalizing revenue with the reference operational variable while allowing `bus_carrier` to filter the revenue contribution in the numerator.
```
brynpickering:
```suggestion
- Fix `get_operation` to correctly return the reference operational variable (`p`) for Link and Process components instead of `p0`. (<!-- md:pr 1614 -->)

- Fix `market_value` to correctly handle multi-port components (Links, Processes) by always normalizing revenue with the reference operational variable while allowing `bus_carrier` to filter the revenue contribution in the numerator. (<!-- md:pr 1614 -->)
```

pypsa/statistics/expressions.py
```diff
@@ -2697,9 +2731,36 @@ def market_value(  # noqa: D417
             "drop_zero": drop_zero,
             "round": round,
         }
-        df = self.revenue(**kwargs) / self.supply(**kwargs)
+
+        revenue_kwargs = {**kwargs, "at_port": at_port}
+        if bus_carrier is None and at_port == "all":
+            revenue_kwargs["at_port"] = "all"
```
brynpickering: This seems unnecessary. It only passes if `at_port == "all"` but you would have already set the kwarg to `at_port == "all"`  in the preceding line.

pypsa/statistics/expressions.py
```diff
@@ -2697,9 +2731,36 @@ def market_value(  # noqa: D417
...
+        rev = self.revenue(**revenue_kwargs)
+
+        @pass_empty_series_if_keyerror
+        def func(n: Network, c: str, port: str) -> pd.Series:
+            sign = -1.0 if c in n.branch_components else n.c[c].static.get("sign", 1.0)
+            p = (sign * get_operation(n, c)).abs()
```
brynpickering: If you're taking the `abs` then you never need to know the sign. Just do `get_operation(n, c).abs()`

pypsa/statistics/expressions.py
```diff
@@ -2697,9 +2731,36 @@ def market_value(  # noqa: D417
...
+            aggregate_across_components=aggregate_across_components,
+            groupby=groupby,
+            at_port=[0],
+            carrier=carrier,
+            bus_carrier=None,
+            nice_names=nice_names,
+            drop_zero=False,
+            round=None,
+        )
+        df = rev / denom.abs()
```
brynpickering: No need for `abs` in denominator since you already set that in `func`.

test/conftest.py
```diff
@@ -254,6 +254,52 @@ def networks_including_solved(request):
 # Network collections
 
 
+@pytest.fixture
+def multiport_process_network():
```
brynpickering: Move into `test_statistics.py`. I would only put fixtures into `conftest.py` if you plan to use them across multiple files.

test/test_statistics.py
```diff
@@ -434,3 +434,120 @@ def test_energy_balance_carrier_nice_name_filter(network_with_nice_name):
...
+    pd.testing.assert_frame_equal(get_operation(n, "Link"), n.c.links.dynamic["p0"])
+    pd.testing.assert_frame_equal(get_operation(n, "Line"), n.c.lines.dynamic["p0"])
+    pd.testing.assert_frame_equal(
+        get_operation(n, "Generator"), n.c.generators.dynamic["p"]
+    )
+    pd.testing.assert_frame_equal(
+        get_operation(multiport_process_network, "Process"),
+        multiport_process_network.c.processes.dynamic["p"],
+    )
+
```
brynpickering, line 455: Let's bring some order and best practice to this test script, at least when adding new tests! Move tests for different top-level methods/classes into their own test class and separate out assertions into their own tests as much as possible. Here (written in GH, so untested/unlinted on my end):


```suggestion

class TestGetOperation:
    @pytest.mark.parametrize(("component", "expected_attr"), [("Link", "p0"), ("Line", "p0"), ("Generator", "p"), ("Store", "e")])
    def test_get_operation(self, ac_dc_network_r, component, expected_attr):
        n = ac_dc_network_r
        pd.testing.assert_frame_equal(
            get_operation(n, component),
            n.components[component].dynamic[expected_attr]
         )
         
    def test_get_operation_multi_port(self, multiport_process_network):
        pd.testing.assert_frame_equal(
        get_operation(multiport_process_network, "Process"),
        multiport_process_network.c.processes.dynamic["p"],
    )

```

test/test_statistics.py
```diff
@@ -434,3 +434,120 @@ def test_energy_balance_carrier_nice_name_filter(network_with_nice_name):
...
+    np.testing.assert_allclose(mv.loc["load"], expected, rtol=1e-10)
+    assert mv.loc["load"] < 0
+
+
+def test_market_value_grouped_branches_is_finite(ac_dc_network_r):
+    n = ac_dc_network_r
+    mv = n.statistics.market_value(nice_names=False, round=None, drop_zero=False)
+
+    assert np.isfinite(mv.loc[("Link", "DC")])
+    assert np.isfinite(mv.loc[("Line", "AC")])
```
brynpickering: move into `TestMarketValue` test class

test/conftest.py
```diff
@@ -254,6 +254,52 @@ def networks_including_solved(request):
 # Network collections
 
 
+@pytest.fixture
```
brynpickering:
```suggestion
@pytest.fixture(scope="module")
```

As you currently use it, it only needs to be built once during testing since you don't change its contents in tests. The time/memory impact is likely negligible but it's good practice to use the broadest possible scope as (a) it will sometimes reduce overall memory/time impact and (b) it can catch cases where the network changes but you _don't expect it to_.

pypsa/statistics/expressions.py
FabianHofmann: oh man, I could have seen all that and the above myself, was rather a relic from the previous energy_balance approach thanks for having another look.

test/test_statistics.py
FabianHofmann: aha! I see there is a lot from you to learn about properly abstracting the tests.

test/conftest.py
FabianHofmann: that was the plan, I think it is a sensible fixture which will have wider application cases, so I would leave it there

FabianHofmann:
@brynpickering all done! final look?

PR #1615, opened 2026-03-10
add reset and update merge network functions 

### LINKED ISSUE #1606: `network.merge(...)` seems to add default values for outputs

### Issue Description

Not 100% on this one.

Calling `n_merged = n.merge(n_other)` seems to restore some values to their default values, even if they were not present in either `network`.

In my case, `n` and `n_other` had both `<n|n_other.>c.links.dynamic["p0"].empty == True`.
However then `n_mergedc.links.dynamic["p0"].empty == False`, with all valeus set to `0`.

This is not bad per-se. The problem is in combination with other functionality that determines if a network has already been solved or not and behaves differently depending on which attributes are set / not set.
In my case that is the `network.cluster.temporal.segmentation(...)` function that throws a warning as it encounters values in `p1` and others:

https://github.com/PyPSA/PyPSA/blob/2e22ebd6bd8315a5e6c51f7581cbb56aa7f1e4af/pypsa/clustering/temporal.py#L28-L41

### Reproducible Example

```python
-
```

### Expected Behavior

`n.merge(...)` should not re-add attributes with their default values if they were not present in the originating networks.

### Installed Versions

<details>
Replace this line.
</details>

FabianHofmann:
Closes #1606

- new property `n.has_ouput` - checks whether network has non-empty output fields
- `Network.reset()` — clears all output data from the network, resetting component output attributes to their
 defaults and clearing the objective value, model, and related internal state. Returns the network to a pre-solve
 state while preserving all input data.
- `merge_output` parameter on `Network.merge()` — controls whether output data is included in the merged result. When  `None` (default) and either network has output, a `FutureWarning` is emitted. Set to True to keep output data or `False` to discard it. 

## TODO

- [ ] align on `n.reset` vs `n.sanitize.clear_outputs`
- [ ] align on usage of output clearing in merge function. alternatively and potentially leaner, we only add non-output attributes in the first place 

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -22,6 +22,8 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - Add `merge_output` parameter to `Network.merge()` to control whether output data is included in the merged network. Defaults to `None` which warns and includes output data; will default to `False` in version 2.
 
+- Add configurable numerical tolerance for consistency checks via the new `params.consistency.numerical_tolerance` option (default `1e-9`). This prevents false warnings from floating-point noise when comparing attributes like `p_min_pu` vs `p_max_pu`, `p_nom_min` vs `p_nom_max`, and `e_sum_min` vs `e_sum_max`.
+
 - New Process component mirroring the behavior of a multi-port Link component with explicit rates (efficiency equivalent to the Link) at each bus, including `bus0`. The component allows to flexibly change the reference unit used for associated costs by adjusting the rates. (<!-- md:pr 1333 -->)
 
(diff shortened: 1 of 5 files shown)


euronion:
Small idea. How about putting this function under the `pypsa.sanitize` module, i.e. instead of calling `network.reset()` something like `network.sanitize.clear_outputs()`?

FabianHofmann:
> Small idea. How about putting this function under the `pypsa.sanitize` module, i.e. instead of calling `network.reset()` something like `network.sanitize.clear_outputs()`?

I like that! Let's hear what @lkstrp says

PR #1619, opened 2026-03-13
Fix port number parsing in consistency check

mgrabovsky:

Port number should be parsed from all trailing digits rather than the last one only. The latter would cause warnings to be emitted in the case bus10 or bus11 contained empty strings.

## Checklist


pypsa/consistency.py
```diff
@@ -74,8 +75,16 @@ def check_for_unknown_buses(
...
+        if component.name in n.branch_components:
+            # extract port number from column name
+            port_match = re.search("\\d+$", attr)
+            if port_match is None:
+                msg = f"Attribute name '{attr}' of {component.name} must contain port number"
+                raise UnexpectedError(msg)
+            port = int(port_match[0])
+
+            if port > 1:
+                missing &= component.static[attr] != ""
```
lkstrp:
```suggestion
        if component.name in n.branch_components and RE_PORTS_GE_2.match(attr):
            missing &= component.static[attr] != ""
```

We have a regex for that already

mgrabovsky: Great! Thanks for the suggestion.

mgrabovsky:
This should be ready now.

PR #1652, opened 2026-04-13
ci: release workflow simplification and move example networks to bucket

lkstrp:
Simplifies the complicated commit again on master release workflow, which lead to the yanked 1.1.1 release. Instead use our bucket for example networks and a user cache.

### DIFF
--- a/pypsa/examples.py
+++ b/pypsa/examples.py
@@ -35,22 +35,14 @@ def _check_url_availability(url: str) -> bool:
         return False
 
 
-def _example_url(name: str) -> str:
-    """Resolve the bucket URL for an example network."""
-    versioned = f"{_EXAMPLES_BASE_URL}/v{__version_base__}/{name}.nc"
-    if _check_url_availability(versioned):
-        return versioned
-    return f"{_EXAMPLES_BASE_URL}/latest/{name}.nc"
-
-
 def _cache_root() -> Path:
     """Return the cache root for example networks."""
     return Path(user_cache_dir("pypsa")) / "examples"
 
 
 def _load_example(name: str) -> Network:
-    url = _example_url(name)
-    cache = _cache_root() / url.removeprefix(_EXAMPLES_BASE_URL + "/")
+    url = f"{_EXAMPLES_BASE_URL}/v{__version_base__}/{name}.nc"
+    cache = _cache_root() / f"v{__version_base__}" / f"{name}.nc"
 
     if not cache.exists():
         if not options.get_option("general.allow_network_requests"):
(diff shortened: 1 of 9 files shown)


test/test_collection.py
```diff
@@ -311,11 +311,12 @@ def test_collection_repr(network1, network2, network3):
     assert "... and 5 more" in repr_str
 
 
-def test_collection_init_with_paths():
+def test_collection_init_with_paths(tmp_path):
     """Test initialization with string paths and Path objects."""
-    # Use example networks from the examples directory
-    example_path1 = "examples/networks/ac-dc-meshed/ac-dc-meshed.nc"
-    example_path2 = "examples/networks/scigrid-de/scigrid-de.nc"
+    example_path1 = str(tmp_path / "ac_dc_meshed.nc")
```
FabianHofmann, line 316: why converting to a string?

pypsa/examples.py
```diff
@@ -41,10 +35,34 @@ def _check_url_availability(url: str) -> bool:
...
+def _cache_root() -> Path:
+    """Return the cache root for example networks."""
+    return Path(user_cache_dir("pypsa")) / "examples"
+
+
+def _load_example(name: str) -> Network:
+    url = _example_url(name)
+    cache = _cache_root() / url.removeprefix(_EXAMPLES_BASE_URL + "/")
+
+    if not cache.exists():
```
FabianHofmann, line 55: this function requires internet connection right at the start without checking first for an existent cache. I don't think this is intended or? feels like we should add a fallback check first that skip the url logic entirely if cache is existent.

test/test_examples.py
```diff
@@ -40,16 +40,51 @@ def test_carbon_management():
...
     assert not _check_url_availability("")
-    assert not _check_url_availability("https://google.com/invalid-url")
-
-    # Test valid URL format (should return True for valid URLs)
-    assert _check_url_availability("https://google.com")
-    assert _check_url_availability("https://google.com/search?q=pypsa")
+    assert not _check_url_availability("https://data.pypsa.org/nonexistent")
+    assert _check_url_availability(
+        "https://data.pypsa.org/networks/examples/latest/ac_dc_meshed.nc"
+    )
```
FabianHofmann, line 90: does this make sense in general for offline test execution?

FabianHofmann (changes requested):
if i am not mistaken there is an issue with offline use and existent cache

pypsa/examples.py
lkstrp: Good catch, thanks !

test/test_examples.py
lkstrp: No it does not. Offline test execution was failing so far as well, but might be worth to migrate to. I can change this part to handle that already

PR #1661, opened 2026-04-23
feat(optimization): support phase_shift in LOPF incl. optimisable taps

### LINKED ISSUE #456: Phase Shifting Transformers (PST)

*Please give a clear and concise description and provide context why the feature would be useful.*
*Also, we'd appreciate any implementation ideas and references you already have.*

> There was this PhD thesis from 2008 by Verboomen who looked at 
integrating PST into optimisation:

> https://repository.tudelft.nl/islandora/object/uuid%3Ac9c72936-380e-4810-ac73-95e684b031db

> There are no dedicated components in PyPSA, but you could consider 
replacing lines with PST with a link to represent the controllability, 
but you need to understand the degrees of freedom properly I think.

### LINKED ISSUE #1220: Update needed on tutorial example "Transformer with non-trivial phase shift and tap ratio"
### Location

https://pypsa.readthedocs.io/en/latest/examples/transformer_example.html

### Documentation problem

Hi there, the last section of the tutorial is a bit confusing because the phase shifting introduced by a transformer doesn't appear in the LOPF result, and the results on the webpage demonstrate this as well.

### LINKED ISSUE #1716: Treatment of n.transformers.phase_shift

### Issue Description

The example above is created with the help of codex and the printed information shows that the optimal solution from running n.optimize() won't change under different values of phase_shift of the transformer. However, n.lpf() and n.pf() do respect the phase_shift values from the transformers and fix them during the calculation. 

```
 phase_shift_deg opf_status  pf_converged  optimize_line_p0_mw  optimize_pst_p0_mw  lpf_line_p0_mw  lpf_pst_p0_mw  pf_line_p0_mw  pf_pst_p0_mw  lpf_minus_opt_pst_mw  pf_minus_opt_pst_mw  opf_gen_mw  pf_gen_mw
             0.0 ok/optimal          True               32.615              67.385          32.615         67.385         32.615        67.386                 0.000                0.000       100.0    100.001
            10.0 ok/optimal          True               32.615              67.385        1171.077      -1071.077       1164.216     -1063.819             -1138.463            -1131.205       100.0    100.397
            20.0 ok/optimal          True               32.615              67.385        2309.540      -2209.540       2261.541     -2159.966             -2276.925            -2227.352       100.0    101.574
           -20.0 ok/optimal          True               32.615              67.385       -2244.311       2344.311      -2195.684      2297.259              2276.925             2229.873       100.0    101.574
```

### Reproducible Example

```python
"""
A09_debug.py
============
Small PyPSA example to illustrate fixed transformer phase-shift behaviour.

The network has two buses joined by:
  * one normal AC line, and
  * one transformer with a fixed phase_shift angle.

The phase-shifting transformer creates a circulating flow around the parallel
path.  PyPSA's optimize() and lpf() use the linear/DC approximation, while
pf() runs the nonlinear AC power flow.  The phase_shift is fixed in all three;
it is not optimized.
"""

import logging
import warnings

import pandas as pd
import pypsa


SNAPSHOT = 0

logging.getLogger("linopy").setLevel(logging.WARNING)
logging.getLogger("pypsa").setLevel(logging.WARNING)
warnings.filterwarnings("ignore", category=FutureWarning)


def build_parallel_network(phase_shift_deg):
    n = pypsa.Network()
    n.set_snapshots([SNAPSHOT])

    n.add("Carrier", "AC")
    n.add("Carrier", "gen")

    n.add("Bus", "A", v_nom=220.0, carrier="AC")
    n.add("Bus", "B", v_nom=220.0, carrier="AC")

    # Cheap generator at A supplies the load at B.
    n.add(
        "Generator",
        "G_A",
        bus="A",
        carrier="gen",
        control="Slack",
        p_nom=1000.0,
        marginal_cost=10.0,
    )
    n.add("Load", "L_B", bus="B", p_set=100.0, q_set=30.0)

    # Parallel path 1: ordinary line.
    n.add(
        "Line",
        "Line_A_B",
        bus0="A",
        bus1="B",
        r=0.01,
        x=5.0,
        b=0.0,
        s_nom=1000.0,
    )

    # Parallel path 2: phase-shifting transformer.
    n.add(
        "Transformer",
        "PST_A_B",
        bus0="A",
        bus1="B",
        r=0.0001,
        x=0.05,
        s_nom=1000.0,
        tap_ratio=1.0,
        phase_shift=phase_shift_deg,
        model="t",
    )

    return n


def fixed_dispatch_copy(n):
    if getattr(n, "model", None) is not None:
        n.model.solver_model = None
    c = n.copy()
    c.generators_t.p_set = c.generators_t.p.copy()
    c.generators.p_set = c.generators_t.p.iloc[0]
    return c


def branch_flows(n):
    return {
        "line_p0_mw": float(n.lines_t.p0.loc[SNAPSHOT, "Line_A_B"]),
        "pst_p0_mw": float(n.transformers_t.p0.loc[SNAPSHOT, "PST_A_B"]),
        "gen_p_mw": float(n.generators_t.p.loc[SNAPSHOT, "G_A"]),
    }


def run_case(phase_shift_deg):
    # 1) OPF: phase_shift is fixed, dispatch is optimized.
    opf = build_parallel_network(phase_shift_deg)
    status, condition = opf.optimize()
    opf_flows = branch_flows(opf)

    # 2) LPF on the OPF dispatch: should be close to optimize() branch flows.
    lpf = fixed_dispatch_copy(opf)
    lpf.lpf()
    lpf_flows = branch_flows(lpf)

    # 3) AC PF on the same OPF dispatch: nonlinear AC validation.
    ac = fixed_dispatch_copy(opf)
    pf_result = ac.pf(distribute_slack=True, slack_weights="p_set")
    ac_flows = branch_flows(ac)

    return {
        "phase_shift_deg": phase_shift_deg,
        "opf_status": f"{status}/{condition}",
        "pf_converged": bool(pf_result["converged"].all().all()),
        "optimize_line_p0_mw": opf_flows["line_p0_mw"],
        "optimize_pst_p0_mw": opf_flows["pst_p0_mw"],
        "lpf_line_p0_mw": lpf_flows["line_p0_mw"],
        "lpf_pst_p0_mw": lpf_flows["pst_p0_mw"],
        "pf_line_p0_mw": ac_flows["line_p0_mw"],
        "pf_pst_p0_mw": ac_flows["pst_p0_mw"],
        "lpf_minus_opt_pst_mw": lpf_flows["pst_p0_mw"] - opf_flows["pst_p0_mw"],
        "pf_minus_opt_pst_mw": ac_flows["pst_p0_mw"] - opf_flows["pst_p0_mw"],
        "opf_gen_mw": opf_flows["gen_p_mw"],
        "pf_gen_mw": ac_flows["gen_p_mw"],
    }


cases = [0.0, 10.0, 20.0, -20.0]
df = pd.DataFrame([run_case(angle) for angle in cases])

pd.set_option("display.width", 180)
pd.set_option("display.max_columns", None)

print("\nFixed phase-shift comparison")
print("============================")
print(df.round(3).to_string(index=False))

print("\nNotes")
print("-----")
print("* phase_shift is not optimized here; it is a fixed input parameter.")
print("* In this installed PyPSA run, optimize() branch flows do not change with phase_shift.")
print("* lpf() and pf() both react strongly to phase_shift, using the value in degrees.")
print("* That means an OPF dispatch can look benign, while later lpf()/pf() shows large phase-shift circulation.")
```

### Expected Behavior

The optimal solution when running n.optimize() will vary with different n.transformers.phase_shift values. 

### Installed Versions

<details>
1.12.1
</details>

MaykThewessen:
## Summary

closes #1220 
closes #1716 
closes #456 

This PR lands two related changes on `Transformer` components in linear optimal power flow (LOPF):

1. **Bug fix (closes #1220)**: include the fixed `phase_shift` in the cycle-based Kirchhoff Voltage Law constraint. Previously LOPF dropped the term, so `n.optimize()` results diverged from subsequent `n.lpf()` / `n.pf()` verification runs.

2. **Feature (closes #456)**: setting `phase_shift_min < phase_shift_max` on a `Transformer` turns its phase shift into a per-snapshot decision variable bounded by those two attributes (degrees). The optimiser re-taps the PST at every snapshot to redistribute cycle flows, modelling TSO operational tap control of phase-shifting transformers (PSTs). The realised per-snapshot angle is written to the new dynamic output `n.transformers_t["phase_shift_opt"]`. The `phase_shift` input keeps its existing name, so this is non-breaking (see Follow-up below).

The two changes share the same KVL modification so they ship as one unit; reviewers can inspect the fixed-shift path by reading the first branch of the new logic and ignoring the variable term.

## Mathematical formulation

Around each cycle $k$ in the network:

$$\sum_{l \in \text{cycle}_k} C_{l,k} \cdot (x_{pu,l} \cdot P_{l} + \varphi_{l}) = 0$$

where $C_{l,k} \in \{-1, 0, +1\}$ is the cycle-direction sign, $x_{pu,l}$ is the per-unit series reactance, $P_l$ is the branch active-power flow (MW), and $\varphi_l$ is the phase shift in radians (zero for lines and transformers without a phase shift). Prior to this PR the $\varphi_l$ term was absent from the LOPF constraint.

Implementation: an unweighted cycle matrix is computed alongside the existing weighted one; fixed $\varphi$ contributions land on the RHS as a constant, optimised $\varphi$ variables enter the LHS with the same cycle-direction coefficients. Fixed and optimised PSTs can coexist in the same network.

## New / changed attributes on `Transformer`

| Attribute | Type | Default | Description |
|---|---|---|---|
| `phase_shift` | static or series (deg) | `0` | Fixed phase-shift setpoint (input). Unchanged from earlier releases, so non-breaking. |
| `phase_shift_min` | float (deg) | `0` | Lower bound. When `phase_shift_min < phase_shift_max` the shift becomes a per-snapshot decision variable. |
| `phase_shift_max` | float (deg) | `0` | Upper bound. |
| `phase_shift_opt` | series (deg) | `0` | Realised / optimised per-snapshot phase shift (output). Equals `phase_shift` for fixed transformers. |

The optimisation variable is `Transformer-phase_shift` and its result is written to `phase_shift_opt`, following the `p_nom` / `p_nom_opt` convention. The `phase_shift_extendable` flag from an earlier revision was dropped: activation is implicit in `phase_shift_min < phase_shift_max`.

## Usage

```python
import pypsa
n = pypsa.Network()
# ... add buses, lines, generators, loads ...
n.add(
    "Transformer", "PST_Breukelen_Lage_Weide",
    bus0="NL_Breukelen-150", bus1="NL_Lage_Weide-150",
    x=0.01, s_nom=644,
    phase_shift_min=-35.0, phase_shift_max=35.0,  # min < max => optimised per snapshot
)
n.optimize(solver_name="highs")
n.transformers_t.phase_shift_opt["PST_Breukelen_Lage_Weide"]
```

## Tests

`test/test_phase_shift_transformer.py` (15 tests, 19 cases):

- **Fixed phase shift in KVL (3)**: baseline equal split, +10° redistributes flow by the analytically derived amount, sign flip on -10° vs +10°.
- **Variable phase shift (7)**: dynamic `phase_shift_opt` writeback, bounds respected, `phase_shift` input preserved after solve, no variable registered when `min == max`, fixed transformer reports its setpoint, unbounded variable rejected, KVL row kept for a variable-only cycle.
- **Mixed fixed + variable (1)**: both kinds coexist in a single network.
- **Multi-investment period (2)**: fixed shift isolated per period; variable shift optimised per period.
- **LOPF matches linear power flow (2, one parametrised)**: LOPF flows pinned to `n.lpf()` across a range of angles, and the variable result reproduced by `lpf` at the optimised angle (issue #1220 regression).

## Test plan for reviewers

- [ ] Run `pytest test/test_phase_shift_transformer.py -v`
- [ ] Run broader LOPF regression: `pytest test/test_lopf_*.py`
- [ ] Verify mathematical correctness: for a 2-bus parallel-path network with equal `x_pu = 0.01` on both branches and a 10° fixed phase shift on the transformer, LOPF flow split should be 33.73/16.27 (derived from `ΔP = φ_rad / x_pu`).
- [ ] Check that an existing model with fixed `phase_shift ≠ 0` now returns different (correct) LOPF results, a behavioural change fixing #1220.

## References

- Verboomen et al., [Analytical Approach to Grid Operation With Phase Shifting Transformers](https://doi.org/10.1109/TPWRS.2007.913197), IEEE Transactions on Power Systems, 23 (1), 41-46 (2008). Section III-A eq. (3): the PST DC-load-flow relation.
- Hörsch et al., [Linear optimal power flow using cycle flows](https://doi.org/10.1016/j.epsr.2017.12.034), EPSR 158, pp. 126-135 (2018). The cycle-based LOPF formulation used in this PR.

## Real-world motivation

Developed while modelling the Dutch Flevopolder-Gelderland-Utrecht (FGU) 150 kV congestion area where TenneT operates two DRTs on the Breukelen Kortrijk pocket. TenneT's own congestion studies show their DRT-optimised dispatch cuts Breukelen 380/150 kV coupling-transformer hours-over-limit from 2064 h to 672 h in 2033 (a 3x reduction), a gap that a fixed-tap PyPSA model cannot reproduce. With `phase_shift_min < phase_shift_max` the LP solver (HiGHS in our case) re-taps the PST per snapshot and closes that gap natively, without any extra_functionality callbacks.

## Follow-up (v2.0)

To stay non-breaking in 1.x, the input keeps its existing name `phase_shift` and the new output is `phase_shift_opt`. In v2.0, align with the standard input/output convention (`p_set` / `p`):

- [ ] rename input `phase_shift` → `phase_shift_set`
- [ ] rename output `phase_shift_opt` → `phase_shift`
- [ ] add a deprecation alias in the last 1.x release so old code warns instead of breaking

Tracked per review discussion with @FabianHofmann and @lkstrp.


MaykThewessen:
Heads-up for reviewers: while validating PR #1661 end-to-end in a downstream model (psa_mee, Dutch 150 kV grid), I hit a related pre-existing issue worth flagging separately — `SpatialClusteringAccessor.cluster_by_busmap` in [pypsa/clustering/spatial.py](https://github.com/PyPSA/PyPSA/blob/master/pypsa/clustering/spatial.py) has no `transformer_strategies` and no Transformer handling, so any Transformer component is silently dropped by the clustering step. This means even a working static `phase_shift` model is wiped if clustering runs after the transformer is added. Not introduced by this PR; mentioning so it's on the radar. Workaround: run Transformer additions AFTER clustering. Happy to open a separate issue if useful.

fneum:
Incidentally, we discussed phase-shifting transformers recently internally.

As for your changes, I have three questions for now:

1. I don't think `phase_shift_extendable` is a good idea, because it doesn't come with any investment costs. I think you can just work with `phase_shift_{min/max/set}` as attributes. No need for `phase_shift_opt`, can just write into `phase_shift`.

2. Do you have any literature reference for how the phase-shift is added to KVL (or to another software library following this approach)? While it sounds intuitive to me - as $x_l f_l$ gives voltage angle difference - I am not quite sure it's that easy. Doesn't the reactance also change with phase shift?

3. Why is the code extension for the phase shifting in KVL so much longer than original constraint? It seems long for just adding a variable term for phase-shiftable transformers. There could be potential to trim.

Other folks have suggested to model PST just with a controllable link.

And yes, the clustering was never built with transformers in mind (historical explanation).

MaykThewessen:
@fneum thanks for the review, and apologies for the delay and the AI-formatted reply above — I'll keep the verbose comparison in a collapsed block from here on per the contributing guidelines.

Short version of where I'd like to land:

I agree `phase_shift_extendable` is the wrong vocabulary — it belongs with capex-coupled variables, and a PST tap has no investment cost. I'd like to refactor along the operational-variable convention instead: drop `phase_shift_extendable` and `phase_shift_opt`, keep `phase_shift_{min,max,set}` only, and write the solver result into `phase_shift` (with `phase_shift_set` as the frozen user input, mirroring the `p_set` / `p` split on generators). Activation rule becomes unambiguous: `min < max` ⇒ variable, otherwise fixed. Three attrs instead of four, and the KVL block trims down accordingly.

On reactance: for the DC-LOPF approximation `x` is treated as constant w.r.t. tap, same convention as PSS/E and PowerWorld. The variation is small relative to the DC approximation itself; happy to add a short note in the docstring.

Reference for the cycle-KVL formulation with phase shifters: Wood, Wollenberg & Sheblé, *Power Generation, Operation, and Control* (3rd ed., 2014), §6.7 eq. 6.32; also Bergen & Vittal *Power Systems Analysis* §6.4–6.5.

I also realised this PR bundles two scopes that should land separately — the static `phase_shift` cycle-KVL bugfix (#1220) is independent of the new variable feature (#456). I'd like to split the bugfix into its own small PR (3 tests, ~30 lines) so it merges quickly, and rework this PR to Option B on top of that. Does that work for you?

<details>
<summary>Detailed comparison of the three design options I considered (verbose)</summary>

### Option A — keep the PR as-is (`phase_shift_extendable`, `phase_shift_opt`)

- ➕ Mirrors `_extendable` / `_opt` so users get an input/output split (re-solving from a saved network doesn't clobber the user-set bound).
- ➖ Semantic mismatch: `_extendable` is reserved for capex-coupled investment variables (`p_nom_extendable`, `e_nom_extendable`, `s_nom_extendable`), all of which couple to a `capital_cost`. PSTs have none.
- ➖ Four new attrs is heavy. Ambiguous activation if a user sets `extendable=True` but leaves `min == max`.

### Option B — collapse to `phase_shift_{min,max,set}`, write the optimum into `phase_shift`

- ➕ Matches the operational-variable convention (`p_min_pu` / `p_max_pu`, no flag). Activation unambiguous.
- ➕ Three attrs. Smaller diff, easier review, less code in `constraints.py`.
- ➖ Loses the input/output separation. Mitigation: `phase_shift_set` = frozen user input, `phase_shift` = solver writeback, same as `p_set` vs `p`.

### Option C — model PST as a controllable Link

- ➕ Zero core changes.
- ➖ Doesn't model the angle: a Link constrains power directly, but a PST physically constrains the voltage-angle shift across the transformer. In a meshed network the angle shift redistributes flow on parallel cycles via KVL; the Link approach loses that effect.
- ➖ Doesn't address #1220 — static `phase_shift` would remain silently dropped in LOPF.
- ➖ Per-scenario user boilerplate instead of a first-class component.

</details>

Separately, I'll fix the CI on this PR once we agree on the direction: `Check types` is a missing `import pypsa` in the new test file, `Check docs` is a doctest in `pypsa.networks` that needs its expected output regenerated after the new transformer attrs, and the release-notes conflict is already resolved locally and ready to push.

MaykThewessen:
Split the static-`phase_shift` cycle-KVL bugfix (#1220) into its own focused PR: #1692. Local checks green (4 new tests, 13/13 LOPF regressions, mypy clean, ruff clean, pre-commit clean).

This PR (#1661) will be reworked to Option B (`phase_shift_{min,max,set}`, no `_extendable` flag, writeback into `phase_shift`) once you confirm direction.

coroa:
I just now went over: Wood, Wollenberg & Sheblé, Power Generation, Operation, and Control (3rd ed., 2014), §6.7 eq. 6.32 .

eq. 6.32 relates to the construction of the decoupled power flow formulation, which is part of the angles formulation of the DC power flow formulation.

I do not find in the book any mention cycle flow formulations (and checking the toc and scanning it quickly, i don't see any formulation other than the default angles one). More crucially, i don't find any active flow control devices like PST there.

Please check the references!

MaykThewessen:
@coroa You're right, apologies. Wood-Wollenberg-Sheblé §6.7 is the nodal-angle DC-PF derivation, not cycle/loop-flow, and the book does not cover PSTs. The citation was AI-generated and I failed to verify it against the source before posting. That's on me, and I'll tighten the loop on reference checks going forward. Withdrawing it.

The cycle-flow LOPF basis itself is from Hörsch et al., *Linear optimal power flow using cycle flows*, EPSR 158 (2018), already cited in the PR. The PST addition is a direct extension of their eq. (7): KVL around each cycle requires the sum of voltage drops to return to zero, and a PST contributes a constant angular drop φ alongside the standard x·P term. So the cycle-KVL with PSTs is

  Σ_{l ∈ cycle_k} C_{l,k} · (x_{pu,l} · P_l + φ_l) = 0

which is what the PR implements. I'm not aware of a published primary source that writes this in the cycle basis explicitly with PSTs: textbook treatments (Glover/Sarma, Bergen/Vittal, Verboomen's TU Delft thesis on PSTs) stay in the nodal-angle basis. If you'd prefer, I can replace the bad citation with a short derivation in the PR description showing the cycle-basis extension from Hörsch eq. (7), and keep Verboomen (2008) as the PST physics reference.

Does that work?

coroa:
@MaykThewessen While the equation is intuitive, i am unsure whether it is correct. The main thing, we would need a reference for, is how to treat PST in a linearized opf fashion (in which formulation is much less important) and while i agree that Verboomen probably contains that and there might also be textbooks, but I am quite unfamiliar with these.

Can you please find a concrete equation where this is implemented in the thesis or a textbook or other published reference!

MaykThewessen:
@coroa Here is a concrete published equation, plus how it maps onto the cycle-basis constraint in this PR.

### Published reference and equation

> **J. Verboomen, D. Van Hertem, P. H. Schavemaker, W. L. Kling, R. Belmans, "Analytical Approach to Grid Operation With Phase Shifting Transformers", IEEE Transactions on Power Systems, vol. 23, no. 1, pp. 41–46, Feb. 2008.** DOI [10.1109/TPWRS.2007.913197](https://doi.org/10.1109/TPWRS.2007.913197). Open-access PDF: [TU Delft repository](https://resolver.tudelft.nl/uuid:13ecde36-8156-47ee-b165-d0c6477e23c5).

This is the journal write-up of the DC-load-flow chapter of Verboomen's PhD thesis. Section III-A "Linearized Power Flow", eq. (3): for a single PST with phase shift `α` in series with the line between buses `i` and `j`,

```
P_ij = y_ij · (δ_ij + α)              (Verboomen 2008, eq. 3)
```

with `y_ij = 1/x_ij` the series susceptance and `δ_ij = δ_i − δ_j`. In system form, eq. (2) `P = Y·δ` becomes eq. (4) `P = Y·δ + Y_α·α`, where `Y_α` holds `+y_ij` at row `i` and `−y_ij` at row `j`.

### Why no separate cycle-basis reference is needed

You are right that the published form stays in the nodal-angle basis. But the PR's cycle constraint is the **exact algebraic image** of eq. (3), not a new model. KVL in the cycle basis is a graph identity for nodal potentials (Hörsch et al. 2018, eq. 4):

```
Σ_ℓ C_{ℓc} · θ_ℓ = 0        ∀ cycles c        where θ_ℓ = θ_i − θ_j
```

The PST changes only the branch constitutive relation. Rearranging eq. (3) to `θ_ℓ = x_ℓ·P_ℓ − α_ℓ` (susceptance form) and substituting:

```
Σ_ℓ C_{ℓc} · (x_ℓ·P_ℓ − α_ℓ) = 0        ∀c
```

The `θ` terms telescope to zero around the cycle (Hörsch eq. 5 is this with `α = 0`), leaving the phase-shift term as a per-cycle constant (static) or variable (extendable). So the cycle form is just eq. (3) inserted into the standard KVL identity already used in PyPSA.

### Sign convention (important, and the PR is correct)

PyPSA's own `n.pf()` / `n.lpf()` use the **opposite** sign to Verboomen. In `power_flow.py::SubNetwork.calculate_B_H`:

```python
self.p_branch_shift = np.multiply(-b, phase_shift, ...)   # b = 1/x, phase_shift in rad
```

so the realised flow is `P_ℓ = (θ_i − θ_j − φ_ℓ) / x_ℓ`, i.e. `θ_i − θ_j = x_ℓ·P_ℓ + φ_ℓ`. Substituting into the KVL identity gives

```
Σ_ℓ C_{ℓc} · (x_ℓ·P_ℓ + φ_ℓ) = 0
```

which is exactly what this PR assembles (static term on the RHS, extendable term on the LHS both reduce to `Σ C (xP + φ) = 0`). So the `+φ` sign is correct **for matching `lpf`/`pf` and closing #1220**. Note this means `φ_PyPSA = −α_Verboomen`; the model is identical, only the sign convention of the attribute differs. Worth a one-line note in the `phase_shift` docstring pinning it to the `pf` convention so nobody later "corrects" it toward the paper and re-opens #1220.

I verified the oracle empirically on the two-branch test network: `n.lpf()` with `phase_shift = +10°` gives `L1 = 33.727 MW`, `T1 = 16.273 MW` (`L1 − T1 = +17.453 MW`), i.e. a positive shift moves flow **off** the shifted transformer onto the parallel line. The PR's analytic expectation `17.453 MW` matches in magnitude.

### One test-coverage gap

The current tests assert only `abs(L1 − T1)` and that `±φ` flip the sign relative to each other. Both are invariant under a wrong-sign fix, which is the exact failure mode of #1220. Two suggestions:

1. Add an absolute-direction assertion to `test_positive_phase_shift_shifts_flow`:

```python
# +phi moves flow off the shifted branch (T1) onto L1, per PyPSA's
# P = (theta0 - theta1 - phi)/x convention (power_flow.calculate_B_H).
assert L1 > T1
```

2. Add a regression class that pins LOPF to PyPSA's own power flow (the authoritative oracle, which respected `phase_shift` before this PR):

```python
class TestLOPFMatchesLinearPowerFlow:
    """Issue #1220 regression: LOPF flows must match n.lpf() / n.pf()."""

    @pytest.mark.parametrize("phase_shift", [-15.0, -5.0, 0.0, 5.0, 15.0])
    def test_lopf_flows_match_lpf(self, phase_shift):
        n = _build(phase_shift=phase_shift)
        n.optimize(solver_name="highs")
        lopf_L1 = n.lines_t.p0["L1"].copy()
        lopf_T1 = n.transformers_t.p0["T1"].copy()
        n.lpf()  # trusted DC oracle; overwrites p0 in place
        np.testing.assert_allclose(n.lines_t.p0["L1"].values, lopf_L1.values, atol=1e-3)
        np.testing.assert_allclose(n.transformers_t.p0["T1"].values, lopf_T1.values, atol=1e-3)

    def test_lopf_flows_match_pf(self):
        # Mirrors the #1220 symptom: optimise, then verify with the AC power flow.
        n = _build(phase_shift=10.0)
        n.optimize(solver_name="highs")
        lopf_L1 = n.lines_t.p0["L1"].copy()
        lopf_T1 = n.transformers_t.p0["T1"].copy()
        conv = n.pf()
        assert conv.converged.all().all()
        np.testing.assert_allclose(n.lines_t.p0["L1"].values, lopf_L1.values, atol=1.0)
        np.testing.assert_allclose(n.transformers_t.p0["T1"].values, lopf_T1.values, atol=1.0)

    def test_extendable_result_consistent_with_lpf(self):
        # Locks the extendable (LHS variable) path to the static (RHS) path and to lpf.
        n = _build(extendable=True, phase_shift_min=-20.0, phase_shift_max=20.0)
        n.optimize(solver_name="highs")
        opt_angle = float(n.transformers_t.phase_shift_opt["T1"].iloc[0])
        lopf_L1 = n.lines_t.p0["L1"].copy()
        check = _build(phase_shift=opt_angle)
        check.lpf()
        np.testing.assert_allclose(check.lines_t.p0["L1"].values, lopf_L1.values, atol=1e-2)
```

coroa:
@MaykThewessen Very nice, eq. 3 in Verboomen 2008 was exactly what i needed, and i am very happy it is indeed as simple. Agreed, that the kirchhoff formulation is only a reformulation and does not need additional evidence. I'll need some time today, to check the sign conventions and will then give you a full code review.

FabianHofmann:
@MaykThewessen I saw there is some weird things going on with multi-investment period optimization. do you mind if I push a fix? also I am not sure about the data model. I would also not keep the `phase_shift_opt` output - dynamic output data does not have a _opt suffix by convention.

FabianHofmann:
@MaykThewessen I have some more plans on addressing the reviewers comments on the pr. fine if I push here?

MaykThewessen:
Hey yes go ahead!

MaykThewessen:
Downstream validation, in case it helps with the merge: I ported our PyPSA-based Dutch grid model (psa_mee) onto the redesigned API here and ran it end to end.

Build under test: this branch (fc8a17f) plus the two Arrow-string IO fixes our model needs under pandas 3, since fc8a17f's master base predates the #1687 merge: #1687 (import-level coercion) and the `create_model`-level coercion from #1690. (We have since re-pinned the model to v1.2.4 with this PR squashed on top, where #1687 is already present.)

What I checked:
- Phase-shift unit + integration tests: green on the new API (`phase_shift_set` input, activation via `phase_shift_min < phase_shift_max`, `phase_shift` output).
- Full weekly-chunked DC-OPF over the real network: all three cross-border PSTs (Meeden, Maasbracht, Zandvliet) optimise within their bounds, and the per-snapshot `phase_shift` writes back through our chunked solver and round-trips via netCDF.
- Wider fast test suite on the new build: 0 failures attributable to this PR (the remaining errors are an unrelated topology gap on our side).

The `min < max` activation rule and the `phase_shift_set` / `phase_shift` split behaved exactly as documented. Thanks @FabianHofmann for taking this over and doing the rework.

docs/release-notes.md
```diff
@@ -6,15 +6,27 @@ SPDX-License-Identifier: CC-BY-4.0
...
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
--->
+
+### Features
+
+- Phase-shifting transformers (PSTs, also known as _dwarsregeltransformatoren_) are now fully supported in linear optimal power flow. Setting `phase_shift_min < phase_shift_max` on a `Transformer` turns the voltage phase-angle shift into a per-snapshot decision variable bounded by those two attributes (degrees); the optimiser re-taps the PST each snapshot to redistribute flow around cycles, modelling TSO operational tap control. The optimised per-snapshot angle is written to the dynamic output `n.transformers_t["phase_shift"]`, mirroring the `p_set`/`p` input/output split. (<!-- md:pr 1661 -->)
```
fneum:
```suggestion
- Phase-shifting transformers (PSTs) are now supported in linear optimal power flow. Setting `phase_shift_min < phase_shift_max` on a `Transformer` turns the voltage phase-angle shift into a per-snapshot decision variable bounded by those two attributes (degrees); the optimiser re-taps the PST each snapshot to redistribute flow around cycles, modelling TSO operational tap control. The optimised per-snapshot angle is written to the dynamic output `n.transformers_t["phase_shift_opt"]`, mirroring the `p_set`/`p` input/output split. (<!-- md:pr 1661 -->)
```

No Dutch please ;)

docs/release-notes.md
```diff
@@ -6,15 +6,27 @@ SPDX-License-Identifier: CC-BY-4.0
...
+
+### Breaking Changes
+
+- The `Transformer` input attribute `phase_shift` has been renamed to `phase_shift_set`. The name `phase_shift` is now a dynamic *output* holding the realised/optimised per-snapshot shift (analogous to `p_set`/`p`). Setting `phase_shift` via `n.add(...)` or loading networks saved with the old attribute still works but emits a `DeprecationWarning` and is aliased to `phase_shift_set`; this fallback will be removed in 2.0. (<!-- md:pr 1661 -->)
+
+### Bug Fixes
+
+- The fixed `phase_shift_set` on `Transformer` components is now included in the cycle-based Kirchhoff Voltage Law constraint in `n.optimize()`. Previously the phase shift was silently dropped in LOPF (only `n.lpf()` and `n.pf()` respected it), causing optimisation results to diverge from subsequent non-linear power-flow verification. (<!-- md:pr 1661 -->)
```
fneum:
```suggestion
- The fixed `phase_shift_set` on `Transformer` components is now included in the cycle-based Kirchhoff Voltage Law constraint in `n.optimize()`. Previously the phase shift was not considered in LOPF (only `n.lpf()` and `n.pf()` respected it), causing optimisation results to diverge from subsequent non-linear power-flow verification. (<!-- md:pr 1661 -->)
```

fneum (changes requested):
Elegant! Thanks for your work! There are some final fixes to do for the docs, but approved in principle.

Final steps: Make sure that the changes are reflected in the user guide (`docs/user-guide/optimization/power-flow.md`). It can be compact, but please:

A) adapt the cycle-KVL constraint with the phase-shift variable
B) add the variable bounds of the phase-shift variable
C) reference the source [1] on that page (linking to `Section III-A "Linearized Power Flow", eq. (3)`).
D) summarise the purpose of a phase-shifting transformer in one sentence

[1] J. Verboomen, D. Van Hertem, P. H. Schavemaker, W. L. Kling, R. Belmans, "Analytical Approach to Grid Operation With Phase Shifting Transformers", IEEE Transactions on Power Systems, vol. 23, no. 1, pp. 41–46, Feb. 2008. DOI [10.1109/TPWRS.2007.913197](https://doi.org/10.1109/TPWRS.2007.913197).

MaykThewessen:
Thanks for the review @fneum!

**A-D (user guide):** Added a compact "Phase-Shifting Transformers (PSTs)" section to `docs/user-guide/optimization/power-flow.md`:

- one-sentence purpose of a PST (controllable phase-angle shift to redistribute flows in a meshed grid without adding/removing active power),
- the KVL cycle constraint adapted with the phase-shift term: $\sum_l C_{l,c} \left( x_l p_{l,t} + \alpha_{l,t} \right) = 0 \quad \forall\, c,t$,
- the variable bounds $\alpha_l^{min} \leq \alpha_{l,t} \leq \alpha_l^{max}$ from `phase_shift_min` / `phase_shift_max` (degrees, converted to radians internally),
- Verboomen et al. (2008) as footnote `[^4]` with an explicit pointer to Section III-A "Linearized Power Flow", eq. (3), DOI 10.1109/TPWRS.2007.913197.

I also added a short admonition on the sign convention ($\alpha^{PyPSA} = -\alpha^{Verboomen}$, consistent with `pf`/`lpf`, cf. #1220) so nobody flips the sign towards the paper.

**Release notes:** Applied both suggestions. Dutch term dropped (also from the `variables.py` docstring and the `transformers.csv` attribute description for consistency), and "silently dropped" softened to "not considered".

**One question on the first suggestion:** it renames the output to `n.transformers_t["phase_shift_opt"]`, but the implementation writes the dynamic output `n.transformers_t["phase_shift"]`, mirroring the `p_set`/`p` split for dynamic outputs (`*_opt` is currently only used for static nominal capacities like `s_nom_opt`). I kept `phase_shift` in the release note so the docs match the code. If you prefer `phase_shift_opt` for the output here, I'm happy to rename it throughout (output attr, IO, tests, docs); just confirm which way you want it.

fneum:
Excellent, that's good to go!

FabianHofmann:
great, thanks @fneum. @MaykThewessen Ddo you feel comfortable with merging the latest master and trying to make the CI run green? Then we can pull this in. happy to support as well

MaykThewessen:
@FabianHofmann heads-up before I push anything: `master` has moved and the branch is now conflicting (`mergeable_state: dirty`), which is also why pre-commit.ci reports "error during mergeable check" rather than actually running. I merged `upstream/master` locally to see the scope: 4 conflicts.

Two are mechanical: `docs/release-notes.md` (just combine our PST entries with master's new ones) and `pypsa/optimization/constraints.py` (an import block only).

Two touch the data model you reworked, so I'd like your call before pushing:

1. `pypsa/network/power_flow.py`: #1759 added `phase_shift` to the standard-type override `attrs` list plus a `_warn_type_param_override(...)` call. Since we renamed the input to `phase_shift_set`, I'd resolve both to `phase_shift_set`. OK?

2. `pypsa/data/component_attrs/transformers.csv`: #1757 reworded `tap_position` (absolute, relative to `tap_neutral`). I'd keep that wording and our `phase_shift_set` / `phase_shift_min` / `phase_shift_max` inputs plus the `phase_shift` output row (master still carries the old plain `phase_shift` input, which our rename replaces).

Are you planning to do the master-merge yourself, or shall I push my resolution? Either is fine, just don't want to collide with your in-flight work.

FabianHofmann:
Thanks for the clear report. Feel free to go ahead!

MaykThewessen:
Pushed the `master` merge (`90014e40`), resolving the four conflicts as described above:

- `constraints.py`: kept `import numpy as np` (used by the KVL degree-to-radian conversion), dropped the now-unused `import linopy`; took master's `LinearExpression` import.
- `power_flow.py`: kept `phase_shift` out of the standard-type override `attrs` list (the type's `phase_shift` feeds `phase_shift_set` in the separate assignment just below) and adopted #1759's `_warn_type_param_override` call.
- `transformers.csv`: took #1757's `tap_position` wording; kept the `phase_shift_set` / `phase_shift_min` / `phase_shift_max` inputs plus the `phase_shift` output.
- `release-notes.md`: combined the PST entries with master's under unified Upcoming Release headers; kept the v1.2.4 section.

Verified against the merged tree: 20/20 in `test_phase_shift_transformer.py`, plus 111 passed / 0 failed across the LOPF / LPF / pf-vs-pandapower / IO suites (including the #1220 lopf-vs-pf regression and the multi-investment-period cases). pre-commit hooks pass locally. I also refreshed the PR description to match the shipped API (`phase_shift_set` input, `phase_shift` output, no `phase_shift_extendable` flag).

FabianHofmann:
sorry, another round required. the maintenance pr made it in the meantime....

fneum (approved):

docs/release-notes.md
```diff
@@ -18,12 +18,19 @@ SPDX-License-Identifier: CC-BY-4.0
...
+
 ### Enhancements
 
 - Speed up [`create_model()`][pypsa.optimization.OptimizationAccessor.create_model] for large networks by computing the bus membership filter in `define_nodal_balance_constraints` with a pandas hash join instead of `xarray.isin` over object-dtype arrays. (<!-- md:pr 1770 -->)
 
+### Breaking Changes
+
+- The `Transformer` input attribute `phase_shift` has been renamed to `phase_shift_set`. The name `phase_shift` is now a dynamic *output* holding the realised/optimised per-snapshot shift (analogous to `p_set`/`p`). Setting `phase_shift` via `n.add(...)` or loading networks saved with the old attribute still works but emits a `DeprecationWarning` and is aliased to `phase_shift_set`; this fallback will be removed in 2.0. (<!-- md:pr 1661 -->)
```
lkstrp: Even with the deprecation helpers, this will still break if a user is reading or writing to  `n.transformers["phase_shift"]`. We don't do breaking changes before major releases

This can be mitigated by avoiding the naming collision:
Now `v1.3`:
- `phase_shift` stays as is, static input
- Introduce `phase_shift_min`, `phase_shift_max` with new functionality
- Introduce `phase_shift_<something>` for dynamic output (the new/old `phase_shift` in here)

`v2.0`:
- rename `phase_shift` to `phase_shift_set`
- potentially also rename `phase_shift_<something>` to `phase_shift`

pypsa/network/io.py
```diff
@@ -1791,6 +1791,21 @@ def _import_components_from_df(
             If True, overwrite existing components.
 
         """
+        if (
```
lkstrp: See comment above, this patching is not allowing for full backwards compatibility and is not clean. With proper renaming per version, it can be fully removed

pypsa/network/io.py
```diff
@@ -2177,7 +2192,9 @@ def import_from_pypower_ppc(
             )  # NB: PYPOWER has strange default of 0. for tap ratio
             | (pdf["branches"].phase_shift != 0)
         )
-        pdf["transformers"] = pd.DataFrame(pdf["branches"][transformers])
+        pdf["transformers"] = pd.DataFrame(pdf["branches"][transformers]).rename(
```
lkstrp: See above

pypsa/network/transform.py
```diff
@@ -216,6 +217,14 @@ def add(
             return_names = options.params.add.return_names
 
         c = as_components(self, class_name)
+        if c.name == "Transformer":
```
lkstrp: See above

pypsa/optimization/variables.py
```diff
@@ -404,6 +404,44 @@ def define_spillage_variables(n: Network, sns: Sequence) -> None:
     n.model.add_variables(0, upper_aligned, name=f"{c.name}-spill", mask=active)
 
 
+def define_phase_shift_variables(n: Network, sns: Sequence) -> None:
+    """Define per-snapshot phase-shift variables for phase-shifting transformers.
+
+    When a ``Transformer`` has ``phase_shift_min < phase_shift_max``, its voltage
```
lkstrp, line 410: Markdown formatting

lkstrp (changes requested):
Thanks @MaykThewessen,

it seems with the formulation you made everyone happy here.

This is still breaking though, even with the patch applied during import. We can fix that by simply avoiding the naming collision, which also removes the need for the patch. See my first comment. @fneum @FabianHofmann 

@MaykThewessen, I am looking here to the most verbose comments, which clearly violate our [contribution guidelines](https://docs.pypsa.org/latest/contributing/contributing/#ai-based-contributions) and make the PR discussion a mess. I see some normal comments from you as well. Please stick to those.

MaykThewessen:
Thanks @lkstrp, agreed: keeping it non-breaking for 1.x is the right call. I'll rework it so `phase_shift` stays the static input (untouched), `phase_shift_min` / `phase_shift_max` add the optimisation bounds, and the optimised per-snapshot value goes to a new dynamic output. That removes the naming collision, so the `_import_components_from_df` / `add` / pypower patches all go away.

One thing to settle: the output attribute name. `_opt` was nixed earlier (it's for static optimised capacities) and you're deferring the `phase_shift` -> `phase_shift_set` rename to 2.0, so `phase_shift` isn't free for the output yet in 1.3. What would you and @FabianHofmann prefer for the 1.3 output name? Happy to go with your pick.

And noted on the comment style, apologies for the noise.

docs/release-notes.md
FabianHofmann: I guess we are in a grey zone. from a data flow perspective things are not breaking 

`n.transformers["phase_shift"]` is so far only used in the pf module and still works as is (with the deprecation).  the time series was never used and is still not used here. 

the optimization module only reads from the new `phase_shift_set` (static+dynamic) and only overwrites the dynamic `phase_shift` which is not used in any previous release. 


The only thing breaking is that the attributing setting anymore: `n.transformers["phase_shift"] = ...` ; Fabian pre-2024 would say "nobody does that, so don't mind" but I understand if that is already breaking enough...

FabianHofmann:
yeah, see my comment above; I don't have a good solution. I was very happy when getting rid of the ugly `phase_shift_opt` output time series. I guess this leaves us with no other option

docs/release-notes.md
lkstrp: If you guys vouch for nobody is noticing it, I am fine with breaking this. We just can't sell stability which everyone is telling us is the most important thing, more important than any new features, and then squeeze in a couple of hope nobody is using it breaking things from time to time.

And I also don't see much on the downside, the workflow above with `phase_shift_opt` until 2.0 is fine, and then you get a clean communicated list of attribute renames. The `_opt` suffix syntax is also used across other attributes. Why rename and break now, while it's not even bringing consistency? I would rather argue to find all of those inconsistencies/ cleaner names and apply them up all at once with `2.0`

FabianHofmann: This is why we are in 2026 now :) Let's do it as you suggested.

MaykThewessen:
Done in b6a19ac1. Non-breaking now: `phase_shift` stays the static (or series) input untouched, `phase_shift_min` / `phase_shift_max` are unchanged, and the optimised per-snapshot value is written to the new dynamic output `phase_shift_opt`. That removes the naming collision, so the deprecation/import shims are gone: the `_import_components_from_df` rename block and the pypower rename in `io.py`, and the `n.add` kwarg-rename hook in `transform.py`. The `phase_shift` -> `phase_shift_set` rename is deferred to 2.0.

Tests green: the phase-shift suite plus the LOPF / LPF / pf-vs-pandapower / IO / consistency regressions; ruff clean.

FabianHofmann:
@MaykThewessen wonderful and thanks for being so flexible. would you mind opening a tracking issue for the renaming after this is merged? I would expect we can merge this tomorrow if @lkstrp agrees

FabianHofmann:
one remaining small thing anticipating the future rename, we can use the `phase_shift` name for the optimization variable in the linopy already today. the last commit renamed the variable to `phase_shift_opt` . this is something that we also don't do for similar cases with `_opt` suffix in the attribute level, ie. `p_nom` is the variable name but the result is written to `p_nom_opt`. would vote for aligning this

MaykThewessen:
Good call, aligned in e69e5a91: the optimisation variable is `Transformer-phase_shift` again (base name, like `p_nom`), and the solved angle is written to the `phase_shift_opt` output (like `p_nom_opt`). Tests still green.

And yes, happy to open the tracking issue for the 2.0 rename (`phase_shift` -> `phase_shift_set`, `phase_shift_opt` -> `phase_shift`) once this is merged.

MaykThewessen:
Added the v2.0 rename plan as a "Follow-up (v2.0)" checklist in the PR description (rename `phase_shift` -> `phase_shift_set`, `phase_shift_opt` -> `phase_shift`, with a deprecation alias in the last 1.x release). Happy to split it into a standalone tracking issue once this merges if you'd prefer. Also refreshed the description to match the shipped API (`phase_shift` input, `phase_shift_opt` output, variable `Transformer-phase_shift`).

lkstrp (approved):
Thank you @MaykThewessen !

MaykThewessen:
Nice!

PR #1662, opened 2026-04-24
fix: regression of n.statistics.transmission with bus_carrier

coroa:
Fixes https://github.com/PyPSA/pypsa-eur/issues/2121 .

Related to https://github.com/PyPSA/PyPSA/pull/1592 .

Fixes regression introduced by #1386 , which aligned to
```
if at_port is None:
    at_port = "bus0" if bus_carrier is None else "all"
```

In the case of transmission which always connects buses with the same carrier,
at_port = "all" leads to double counting each line once with + and once with - leading to 0 everywhere.

Here we fix this single case with at_port = "bus0" in transmission.

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,13 +6,17 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Release Notes
 
-<!--## Upcoming Release
+## Upcoming Release
 
 !!! info "Upcoming Release"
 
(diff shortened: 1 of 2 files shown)


lkstrp (approved):
Can we not just completely changed the default? As in my last commit?

coroa:
> Can we not just completely changed the default? As in my last commit?

Probably identical, yes

PR #1663, opened 2026-04-24
fix: datetime-like support for CFTimeIndex snapshots

### LINKED ISSUE #1090: Datetime-like support for CFTimeIndex snapshots
PyPSA offers some extra functionality / quality of life when using `DatetimeIndex` snapshots. It would be convenient to offer the same functionality for [`CFTimeIndex`](https://docs.xarray.dev/en/stable/generated/xarray.CFTimeIndex.html) where you want the snapshots on a [CF calendar](http://cfconventions.org/cf-conventions/cf-conventions#calendar).

`CFTimeIndex` overlaps a lot with `DatetimeIndex`, but it is not a subclass of it, therefore PyPSA will treat a `CFTimeIndex` generically rather than datetime-like.

darsh-mishr:
Closes #1090.

`xarray.CFTimeIndex` (used for CF calendars like `noleap` and `360_day`, which underpin most CMIP6 climate datasets) is a `pd.Index` subclass but not a `pd.DatetimeIndex` subclass. PyPSA's `isinstance(..., pd.DatetimeIndex)` checks missed it, causing two silent failures:

1. `set_snapshots` rebuilt the index via `pd.Index(snapshots, name="snapshot")`, collapsing `CFTimeIndex` to a generic object `Index` and losing the calendar metadata.
2. `weightings_from_timedelta=True` was silently skipped — the snapshot spacing was ignored and default weights of `1.0` were used, even though `CFTimeIndex` supports the same `.to_series().diff()` arithmetic.

This PR:

- Adds `_is_datetime_like(idx)` in `pypsa/common.py`, covering `pd.DatetimeIndex` and `xarray.CFTimeIndex`.
- In `set_snapshots`, renames in place when the input is already a `pd.Index` subclass instead of rebuilding through `pd.Index(...)`, so `CFTimeIndex` (and any future pandas index subclass) survives the call.
- Replaces the `isinstance(..., pd.DatetimeIndex)` guards on the `weightings_from_timedelta` path with `_is_datetime_like(sns)`.
- Adds parametrized tests (`noleap`, `360_day`) asserting both type preservation and correct timedelta-derived weights.

Note: `n.cluster.temporal.resample()` still refuses `CFTimeIndex` because pandas' `.resample()` does not support it natively — routing through xarray's resampler is a larger design change and is left as a follow-up.

## Why this might be worth prioritising

The failure is silent: no exception, just a single `INFO` line, and `snapshot_weightings` quietly defaults to `1.0` regardless of the real snapshot spacing.

Because downstream quantities are computed as `Σ(weight × power)`, an N-hourly input (3h or 6h is typical for CMIP6) under-reports energy, variable costs and emissions by a factor of N, and the objective under-prices operations relative to new capacity — biasing capacity expansion toward less storage and more peaker dispatch. The total objective barely shifts (capital costs dominate), which makes it easy to miss by eyeballing headline numbers.

This shows up for anyone feeding CMIP6 / CORDEX output through xarray into PyPSA, so it seemed worth flagging here in case it raises the priority of the issue.

## Checklist

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,13 +6,17 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Release Notes
 
-<!--## Upcoming Release
+## Upcoming Release
 
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
-    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.-->
+    to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
+
+### Bug Fixes
+
+- `Network.set_snapshots` now preserves `xarray.CFTimeIndex` inputs (CF calendars such as `noleap` and `360_day`, standard for CMIP6 climate datasets) instead of collapsing them to a generic object index, and correctly computes `weightings_from_timedelta` for them. See [#1090](https://github.com/PyPSA/PyPSA/issues/1090).
 
 
 ## [**v1.2.0**](https://github.com/PyPSA/PyPSA/releases/tag/v1.2.0) <small>21st April 2026</small> { id="v1.2.0" }
--- a/pypsa/common.py
+++ b/pypsa/common.py
@@ -22,6 +22,7 @@ import pandas.testing as pd_testing
 from deprecation import deprecated
 from packaging import version
 from pandas.api.types import is_list_like
+from xarray import CFTimeIndex
 
 from pypsa._options import options
 from pypsa.definitions.structures import Dict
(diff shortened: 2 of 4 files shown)


lkstrp (commented):
The PR allows using `CFTimeIndex` in `n.set_snapshots`, but this also needs to be handled in IO, temporal clustering etc., and properly tested everywhere.

I'll close this for now.

fneum:
Just to stress: Thanks @darmis007 for your contribution! We'd highly welcome this functionality and will do our best to support you; closing the PR just means there's a little more work required so that `CFTimeIndex` can be used everywhere PyPSA handles time series.

darsh-mishr:
Thanks @fneum, that really means a lot. And thanks @lkstrp for taking the time to review and lay out clearly what's still missing; that gives me a much better picture of the scope.

I'd love to keep working on it. I'll be upfront that the IO and temporal-clustering internals are new ground for me, so any pointers from the team on where to start reading would be hugely appreciated, especially on the corners of `CFTimeIndex` that are likely to be trickier than they look. I'd rather take the time to understand the design than guess and create review burden.

A couple of things I'd value guidance on:

- One larger PR once it all works, or smaller incremental ones?
- Worth opening a tracking issue first to align on scope, or just iterate in PRs?

Thanks again. I'll take my time with it and lean on the team's guidance as I go.

PR #1665, opened 2026-04-28
refactor: move from string based component references to typed instances

lkstrp:
A big refactor split out from validation work. The diff is still large, but shouldn't be tooo horrible to review. I'll also start to use a `dev` branch as base here now. `dev` is an up to date rebase of `master` to allow firing patches on the master but still being able to merge features. Also this will have follow ups before `v1.3`. 

- Refactors component type dispatch to use Python's class hierarchy rather than string set membership
	- There are now mixins (MultiPort, Branch, OnePort, Controllable, Passive, StandardType) and internal code uses isinstance() checks. 
	- Also the signatures of functions can now be updated to represent the actual Components type/category. 
	- Some of the internal function signatures and most of the code is migrated. Internal functions should fully stop passing string representations of components.
- `c.is_branch`
	-  Instead of `c.name in n.branch_components` (same goes for `c.name == "GlobalConstraints"` etc.) there is now `c.is_branch` for user-facing code.
	- Internal use is `isinstance(c, Branch)`, since then we get a fully typed instance. I think this is more Pythonic and better fits the signatures. A type guard for those boolean helpers is a bit hacky and brings in tons of circular problems again. But not super sure yet. I would probably not even expose the categories to the user anyway, and keep them as an internal structure, as already done with `MultiPort`.
- `n.components.filter(..)`
	- The sets are deprecated and to get a group of components you can now use `n.components.filter(branch=True, ..)`, which also yields the component class and not a string.
- I tried to come up with a clean-up of singular PascalCase (`Generator`) vs plural snake_case (`generators`). It's not easy to bring that in backwards compatible.
	- PascalCase (`Generator`)
		- Wherever we refer to a single Component not the Components collection (e.g. the dataframe, or the class instance)
		- Where this makes sense
			- Any dataframe index (Statistics, `n.branches()`)
			- `n.add` etc. should be fine with both, but we wanna move those functions to be used from the class directly anyway (-> `n.generators.add`)
			- String representations/ error messages/ logs if referring to a single component
		- Where it does not and it's deprecated
			- `n.components['Generator']`
			- set properties (`n.branch_components`.. )
			- Data csvs
			- Passing weights for mga
		- Where it does not and it's kept, could be a breaking `2.0` change
			- Linopy constraints and variables: `Generator-p_nom`
			- Maybe even get rid of single use completly?
	
	- snake_case (`generators`)
		- All other use
- The network collection is not touched at all, which I had planned during the implementation. It now just duck types components classes via `__class__` which is not very pretty. Same goes for SubNetwork Components.

This PR is not breaking anything exposed to the user, and only changing things if they can be deprecated first. Internally the difference between both string representation doesn't matter too much since we are moving to the instance directly. If we deprecate this further and even need those two representations (three with the class), can be discussed when we are in `2.0` territory. See diff in tests of first commit to see which parts are breaking.

## Checklist


docs/examples/mga.ipynb
```diff
@@ -190,7 +190,7 @@
    "source": [
     "The function `n.optimize.optimize_mga` takes three main arguments:\n",
     "1. The `slack` for the allowed relative cost deviation from the cost-optimum (0.05 corresponds to 5%).\n",
-    "2. A dictionary of weights for defining the new objective function. The first level defines the component (e.g. \"Generator\"), the second level defines the optimisation variable (e.g. `p_nom` for investment), and the third level defines the component name (e.g. from `n.generators.index`).\n",
+    "2. A dictionary of weights for defining the new objective function. The first level defines the component (e.g. \"generators\"), the second level defines the optimisation variable (e.g. `p_nom` for investment), and the third level defines the component name (e.g. from `n.generators.index`).\n",
```
fneum, line 193: This is a breaking change that should only take effect with 2.0, right?

pypsa/clustering/spatial.py
```diff
@@ -206,7 +209,7 @@ def flatten_multiindex(m: pd.MultiIndex, join: str = " ") -> pd.Index:
 def aggregateoneport(
     n: Network,
     busmap: dict,
-    component: str,
+    component: ComponentsLike,
```
fneum, line 212: Is the difference between Components and ComponensLike that the latter can still include a string? Why not `str | Components`?

pypsa/clustering/spatial.py
```diff
@@ -220,8 +223,8 @@ def aggregateoneport(
         The network containing the generators.
     busmap : dict
         A dictionary mapping old bus IDs to new bus IDs.
-    component : str
-        The component to aggregate.
+    component : str | Components
+        The component to aggregate (name or Components instance).
```
fneum, line 227: One disadvantage is that user has to deal with more classes; before they could just pass a string here.

pypsa/clustering/spatial.py
```diff
@@ -636,7 +645,7 @@ def busmap_by_hac(
             buses_i = n.c.buses.static.index
 
         if branch_components is None:
-            branch_components = n.branch_components
+            branch_components = [c.list_name for c in n.components.filter(branch=True)]
```
fneum, line 648: This looks more complicated than before.

pypsa/components/_types/_patch.py
```diff
@@ -16,6 +16,37 @@
 _T = TypeVar("_T", bound=Components)
 
 
+def _format_str_dtype(value: Any) -> Any:
```
fneum, line 19: Is this really necessary to do by hand? Could imagine there's built in funcionality for that in pandas and alike?

pypsa/components/_types/loads.py
```diff
@@ -10,11 +10,12 @@
...
 
 from pypsa.components._types._patch import patch_add_docstring
-from pypsa.components.components import Components
+from pypsa.components.categories import Controllable, OnePort
+from pypsa.components.types import all_components as component_types
 
 
 @patch_add_docstring
-class Loads(Components):
+class Loads(Controllable, OnePort):
```
fneum, line 18:
```suggestion
class Loads(OnePort):
```
Is not controllable.

fneum, line 18:
```suggestion
class Loads(OnePort):
```
Is not controllable.

fneum, line 18:
```suggestion
class Loads(OnePort):
```
Is not controllable.

pypsa/components/categories.py
```diff
@@ -18,14 +21,57 @@
 logger = logging.getLogger(__name__)
 
 
-class _Multiport(Components):
-    """_Multiport components class.
+class Branch(Components):
```
fneum, line 24: Do these show up in documentation somewhere?

pypsa/components/categories.py
```diff
@@ -18,14 +21,57 @@
...
+    """Marker for passive components: lines, transformers, shunt_impedances.
+
+    <!-- md:badge-version v1.3.0 -->
+    """
+
+    is_passive: bool = True
+
+
+class Controllable(Components):
+    """Marker for controllable components: generators, loads, stores, storage_units, links, processes.
```
fneum, line 52:
```suggestion
    """Marker for controllable components: generators, stores, storage_units, links, processes.
```

pypsa/components/types.py
fneum, line 1: Not sure I understood what made this rewrite necessary. Doesn't this bear high risk of mistakes?

pypsa/data/component_attrs/buses.csv
```diff
@@ -1,20 +1,20 @@
...
-"control","string","n/a","PQ","P,Q,V control strategy for power flow, must be ""PQ"", ""PV"" or ""Slack"". Note that this attribute is an output inherited from the controls of the generators attached to the bus. Do not set by hand.","Output"
-"generator",string,"n/a","n/a","Name of slack generator attached to slack bus.","Output"
-"sub_network","string","n/a","n/a","Name of connected sub-network to which bus belongs. This attribute is set by `n.determine_network_topology()`. Do not set by hand.","Output"
-"p","series","MW",0.,"active power at bus (positive if net generation at bus)","Output"
-"q","series","MVar",0.,"reactive power (positive if net generation at bus)","Output"
-"v_mag_pu","series","per unit",1.,"Voltage magnitude, per unit of `v_nom`","Output"
-"v_ang","series","radians",0.,"Voltage angle","Output"
-"marginal_price","series","currency/MWh",0.,"Shadow price from energy balance constraint","Output"
+attribute,type,dtype,unit,default,description,status,dynamic,nullable
+name,string,str,,,Unique name,Input (required),False,False
```
fneum, line 2: Do I see it right that the new attributes `dtype` and `dynamic` are splitting info from `type` and are set to replace it in the long run? If yes, how is information "static *or* series" retained?

What does `nullable` mean in this context?

pypsa/data/component_attrs/networks.csv
fneum, line 1: This wasn't used anywhere, right?

pypsa/data/variables.csv
```diff
@@ -1,35 +1,35 @@
...
-Store,p,True,True,False,False,False,True
-Store,e_nom,False,False,False,False,True,False
-Store,n_mod,False,False,False,False,False,True
-StorageUnit,p_dispatch,True,True,False,False,False,False
-StorageUnit,p_store,False,False,False,False,False,False
-StorageUnit,state_of_charge,False,False,False,True,False,False
-StorageUnit,p_nom,False,False,False,False,True,False
-StorageUnit,n_mod,False,False,False,False,False,True
-StorageUnit,spill,False,False,True,False,False,True
+generators,p,True,True,False,False,False,False
```
fneum, line 2: Is this really backwards compatible?

pypsa/network/descriptors.py
```diff
@@ -176,7 +179,7 @@ def get_switchable_as_iter(
 
         Examples
         --------
-        >>> gen = n.get_switchable_as_iter('Generator', 'p_max_pu', n.snapshots[:2])
+        >>> gen = n.get_switchable_as_iter('generators', 'p_max_pu', n.snapshots[:2])
```
fneum, line 182: I still feel quite attached to "Generator" pascal case somehow. ;)

pypsa/network/graph.py
```diff
@@ -262,11 +270,17 @@ def incidence_matrix(
         from pypsa.networks import Network, SubNetwork  # noqa: PLC0415
 
         if branch_components is not None:
-            branch_components = set(branch_components)
+            branch_components = {
+                self.components._get(c).list_name for c in branch_components
+            }
```
fneum, line 275: More complicated than before?

pypsa/network/io.py
fneum, line 1: How will these changes affect the import ability of networks from older PyPSA versions in the long-run? Will it still be possible?

pypsa/network/power_flow.py
```diff
@@ -624,7 +625,7 @@ def aggregate_multi_graph(sub_network: SubNetwork) -> None:
 
             # remove all but first line
             for line in line_objs[1:]:
-                n.remove("Line", line[1])
+                n.remove("lines", line[1])
```
fneum, line 628: Why not change directly to `n.c.lines.remove(line[1])`?

pypsa/network/transform.py
```diff
@@ -479,17 +479,21 @@ def merge(
             Merged network, or None if inplace=True
 
         """
-        to_skip = {"Network", "SubNetwork", "LineType", "TransformerType"}
+        to_skip = {"sub_networks", "line_types", "transformer_types"}
```
fneum, line 482: why use strings here and not component type classes?

pypsa/optimization/constraints.py
```diff
@@ -1083,8 +1079,9 @@ def define_nodal_balance_constraints(
 
         #  drop non-existent multiport buses which are ''
         if (
-            c.name in n.controllable_branch_components
-            and isinstance(c, _Multiport)
+            isinstance(c, Branch)
+            and isinstance(c, Controllable)
```
fneum, line 1083: This is 1:1 implementation of early version, but I think just `isinstance(c, MultiPort) should be sufficient.

pypsa/optimization/constraints.py
```diff
@@ -1440,32 +1434,32 @@ def define_committability_variables_constraints_with_fixed_upper_limit(
 
     active = c.da.active.sel(snapshot=sns, name=inter_i) if n._multi_invest else None
 
-    status = m.variables[f"{component}-status"].loc[sns, inter_i]
+    status = m.variables[f"{c.name}-status"].loc[sns, inter_i]
```
fneum, line 1437: This will still be called "Generator-status" right? Otherwise, would be huge backwards-incompatible change.

pypsa/optimization/global_constraints.py
```diff
@@ -153,17 +153,16 @@ def define_nominal_constraints_per_bus_carrier(n: Network, sns: pd.Index) -> Non
 
         lhs = []
 
-        for c, attr in nominal_attrs.items():
-            var = f"{c}-{attr}"
-            static = n.c[c].static
+        for c in n.components[list(nominal_attrs)]:
+            attr = nominal_attrs[c.list_name]
+            var = f"{c.name}-{attr}"
+            static = c.static
```
fneum, line 159: More complicated than before?

pypsa/statistics/abstract.py
```diff
@@ -174,7 +175,17 @@ def _aggregate_components(
...
+                "Generator",
+                "Line",
+                "Link",
+                "Load",
+                "Process",
+                "ShuntImpedance",
+                "StorageUnit",
+                "Store",
+                "Transformer",
+            ]
```
fneum, line 188: Why hard-code and not use `n.components.filter(...)`?

test/test_pf_against_pypower.py
```diff
@@ -105,7 +105,7 @@ def test_pypower_case():
 
     # compare branch flows
     for c in n.components:
-        if c.name not in n.passive_branch_components:
+        if not (c.is_branch and c.is_passive):
```
fneum, line 108: When to use `c.is_branch` and when `isinstance(c, Branch)`?

pypsa/networks.py
```diff
@@ -2023,7 +2028,7 @@ def iterate_components(
 
         """
         if components is None:
-            components = self.n.all_components
+            components = [c.list_name for c in self.n.components.values()]
```
fneum, line 2031: More complicated than before?

pypsa/statistics/expressions.py
```diff
@@ -2120,14 +2127,14 @@ def transmission(  # noqa: D417
         n = self._n
 
         if components is None:
-            components = n.branch_components
+            components = ["Line", "Transformer", "Link", "Process"]
```
fneum, line 2130: Why not `n.components.filter(...)`?

fneum (changes requested):
Long one to review!

- I loved everything related to the component classes. The `isinstance(...)` checks are quite elegant and much better than handling strings.

- The `n.components.filter()` function is nice, but did not universally reduce code complexity. In fact, there were a few instances, where I thought the code got more complex than before.

- I didn't see much benefit from phasing out the Pascal case to be honest. What is the benefit? How convinced are we that the value of doing that outweighs the cost of change? Costs would be significant, since it shows up in all model variables/constraints.

- Load is not controllable component.

docs/examples/mga.ipynb
lkstrp: It's not, it just deprecates now and both versions work.

pypsa/clustering/spatial.py
lkstrp: Yeah, wasn't too sure here as well. For `Components | str` also argues that we can still do `Branches | str` and don't need `BranchesLike` or anything, which would be quite confusing. I'll migrate this to `Components | str`.

pypsa/clustering/spatial.py
lkstrp: I don't understand this one. You can now pass just both? String as before still works and now also a components instance?

pypsa/clustering/spatial.py
lkstrp: It does. But it only does so because I did not refactor further. The proper refactor would migrate from using `branch_components` as a list of strings here to `branch_components` as a list of instances. Then you would just do:
```
if branch_components is None:
    branch_components = n.components.filter(branch=True)
else:
    branch_components = n.components[branch_components]
```

And use that list of instances from there.

pypsa/components/categories.py
lkstrp: As just discussed, it will now be documented and we will remove `c.is_branch`.

pypsa/components/types.py
lkstrp: This is mainly removing code that was previously handling the option to add custom components, which we had in a recent version but fully deprecated without anyone noticing. I will also further simplify this in the next PR, since what we need here is just a translation of component csv files to components classes.

pypsa/data/component_attrs/buses.csv
lkstrp: Yes, both right, but I will revert again since this should go in the next PR. To be discussed there.

pypsa/data/component_attrs/networks.csv
lkstrp: No, we removed that with component classes in `0.33` I think already.

pypsa/data/variables.csv
lkstrp: If people curl to `raw.github.com` and the `variables.csv` file, no. In the optimization module where this is used, yes. If we don't change that here, you just get spammed with deprecation warnings.

pypsa/network/descriptors.py
lkstrp: In `n.add` it could still make sense, because you can add a single `Generator`. But here it is also just not correct because you get a list of `generators` back. In general, I strongly vote for getting rid of those two representations in the long term, and `generators` is just the better one.

pypsa/network/graph.py
lkstrp: See comment above

pypsa/network/io.py
lkstrp: No, and it should never. I wrote tests for that, I can move them into this PR already.

pypsa/network/transform.py
lkstrp: Would again need further refactoring, otherwise it's just patched and more complicated

pypsa/optimization/constraints.py
lkstrp: Yes, this is what I would put on the `v2.0` breaking list. See comment in the main PR text:

> Where it does not and it's kept, could be a breaking 2.0 change
>     Linopy constraints and variables: Generator-p_nom
>     Maybe even get rid of single use completly?

pypsa/networks.py
lkstrp: See comment above.

PR #1666, opened 2026-04-28
fix: clear stale dynamic attributes on n.add(overwrite=True)

### LINKED ISSUE #1628: `n.add(..., overwrite=True)` preserves `dynamic` attributes of old components

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/pypsa/releases) of PyPSA.

### Issue Description

Using `overwrite=True` to overwrite a component leave intact dynamic attributes that existed for the old component, that are not set.

Not sure if this is intended (there can be use cases for it), but I would expect `overwrite=True` to behave as `n.remove(...)` and then `n.add(...)` on this component.

Observed e.g. with dynamic `marginal_cost` on a component, that pre-existed after adding a new component with different attrs but *static* `marginal_cost`. However the *dynamic* attributes were kept, yielding to unexpected behaviour.

### Reproducible Example

```python
-
```

### Expected Behavior

-

### Installed Versions

<details>
Replace this line.
</details>

darsh-mishr:
## Summary

`n.add(..., overwrite=True)` was replacing a component's static row but silently leaving behind any pre-existing time-varying (dynamic) attributes attached to the same component name. Because PyPSA's optimization uses the dynamic value whenever both static and dynamic exist for an attribute, the stale dynamic data kept shadowing the new static value at solve time — with no warning.

This PR makes `overwrite=True` behave symmetrically with `n.remove(...)` followed by `n.add(...)`, which is the behavior the docstring and user mental model already imply.

Closes #1628.

## Why this matters

Whenever a workflow re-uses a network across configurations and reaches for `overwrite=True` to swap a parameter, the result is currently fictional:

- **Scenario sweeps** — flip a marginal cost from a time-varying curve to a flat scalar; the optimizer keeps using the curve.
- **Rolling horizon / redispatch** — overwriting components per window leaves prior-window dynamics in place.
- **Sensitivity analysis** — the parameter you "varied" never actually moves in the optimization.

Because nothing throws and `print(n.<comp>)` shows the new static value, the bug is invisible until someone instruments `get_switchable_as_dense()` mid-sweep.

### Minimal reproducer

```python
import pandas as pd, pypsa

n = pypsa.Network()
n.set_snapshots(pd.date_range("2025-01-01", periods=3, freq="h"))
n.add("Bus", "b")
n.add(
    "Generator", "g", bus="b", p_nom=100,
    marginal_cost=pd.Series([10.0, 20.0, 30.0], index=n.snapshots),
)
n.add(
    "Generator", "g", bus="b", p_nom=100,
    marginal_cost=5.0, overwrite=True,
)

print(n.generators.at["g", "marginal_cost"])              # 5.0   (looks correct)
print(n.generators_t.marginal_cost["g"].tolist())          # [10.0, 20.0, 30.0]   (BUG: stale)
print(n.get_switchable_as_dense("Generator", "marginal_cost")["g"].tolist())
# before this PR: [10.0, 20.0, 30.0]   (the optimizer sees these, not 5.0)
# after  this PR: [ 5.0,  5.0,  5.0]
```

## Fix

In `_import_components_from_df` (`pypsa/network/io.py`), the `overwrite=True` branch already drops the duplicated names from the static dataframe. Extend it to also drop those names from every dynamic dataframe of the component class:

```python
else:
    old_static = old_static.drop(duplicated_components)
    # Also clear stale dynamic columns for the names being overwritten,
    # so overwrite=True behaves as remove() + add(). See issue #1628.
    dynamic = self.c[cls_name].dynamic
    for k in dynamic:
        cols = dynamic[k].columns.intersection(duplicated_components)
        if len(cols) > 0:
            dynamic[k] = dynamic[k].drop(columns=cols)
```

The patch is local to the existing duplicate-handling branch — no new public API, no new keyword arguments. Lines 1843–1853 (which run later in the same function) reindex each dynamic key against the new static index and fill defaults, so dropped columns naturally come back as default-filled if the new `add()` call re-supplies them, and stay gone otherwise — matching `remove()` + `add()` semantics exactly.

The only other caller of `_import_components_from_df` (`pypsa/network/io.py:1398`, the NetCDF/CSV import path) passes `overwrite=False`, so it is unaffected.

## Test plan

- [x] New regression test `test_add_overwrite_clears_stale_dynamic` in `test/test_network.py` covering the exact scenario above (dynamic attribute pre-existing, overwriting call passes only a static value, asserts the dynamic column is dropped and `get_switchable_as_dense` returns the static value).
- [x] Existing `test_add_overwrite_static` and `test_add_overwrite_varying` still pass.
- [x] Full local runs of `test_network.py`, `test_components.py`, `test_components_array.py`, `test_network_transform.py` — 131/131 pass.

## Checklist

- [x] Code changes are sufficiently documented (inline comment references the issue).
- [x] Unit tests for the bug fix were added.

### DIFF
--- a/pypsa/network/io.py
+++ b/pypsa/network/io.py
@@ -1808,7 +1808,10 @@ class NetworkIOMixin(_NetworkABC):
                 )
                 new_static = new_static.drop(duplicated_components)
             else:
-                old_static = old_static.drop(duplicated_components)
+                # Use n.remove() so both static rows and dynamic columns are
+                # cleared, making overwrite=True behave like remove() + add().
+                # See issue #1628.
+                self.remove(cls_name, duplicated_components)
 
         # Concatenate to new dataframe
         if not old_static.empty:
--- a/pypsa/network/io.py
+++ b/pypsa/network/io.py
@@ -1808,9 +1808,6 @@ class NetworkIOMixin(_NetworkABC):
                 )
                 new_static = new_static.drop(duplicated_components)
             else:
-                # Use n.remove() so both static rows and dynamic columns are
-                # cleared, making overwrite=True behave like remove() + add().
-                # See issue #1628.
                 self.remove(cls_name, duplicated_components)
 
         # Concatenate to new dataframe
--- a/test/test_network.py
+++ b/test/test_network.py
@@ -306,6 +306,19 @@ def test_add_overwrite_varying(n_5bus_7sn, caplog):
     assert (n_5bus_7sn.c.buses.dynamic.p.loc[:, bus_names[:5]] == p).all().all()
 
 
+def test_add_overwrite_clears_stale_dynamic():
+    """Regression for #1628: overwrite=True must clear pre-existing dynamic
+    columns even if the new call does not re-supply them.
+    """
+    n = pypsa.Network(snapshots=range(3))
+    n.add("Bus", "b")
+    n.add("Generator", "g", bus="b", marginal_cost=[10.0, 20.0, 30.0])
+    assert "g" in n.c.generators.dynamic.marginal_cost.columns
+
+    n.add("Generator", "g", bus="b", marginal_cost=5.0, overwrite=True)
+    assert "g" not in n.c.generators.dynamic.marginal_cost.columns
+
+
 def test_add_stochastic():
     """Test adding components to stochastic networks."""
     n = pypsa.Network(snapshots=range(5))
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -18,6 +18,8 @@ SPDX-License-Identifier: CC-BY-4.0
 
 - Fix `n.statistics.transmission()` returning zero flows when `bus_carrier` is set. (<!-- md:pr 1662 -->)
 
+- Fix `n.add(..., overwrite=True)` leaving stale time-varying (dynamic) attributes from the previously existing component, which silently shadowed the new static values at solve time. `overwrite=True` now behaves consistently with `n.remove(...)` followed by `n.add(...)`. See [#1628](https://github.com/PyPSA/PyPSA/issues/1628).
+
 
 ## [**v1.2.0**](https://github.com/PyPSA/PyPSA/releases/tag/v1.2.0) <small>21st April 2026</small> { id="v1.2.0" }
 


pypsa/network/io.py
```diff
@@ -1809,6 +1809,13 @@ def _import_components_from_df(
                 new_static = new_static.drop(duplicated_components)
             else:
                 old_static = old_static.drop(duplicated_components)
```
lkstrp: Can we just straight up call `n.remove` here?

test/test_network.py
```diff
@@ -306,6 +306,46 @@ def test_add_overwrite_varying(n_5bus_7sn, caplog):
     assert (n_5bus_7sn.c.buses.dynamic.p.loc[:, bus_names[:5]] == p).all().all()
 
 
+def test_add_overwrite_clears_stale_dynamic():
```
lkstrp, line 309: Could you make that test more minimal?

lkstrp (changes requested):
Thanks @darmis007,

just some comments which would simplify this fix. Also note that we are currently changing a lot of the interface here, but it is still worth to patch this.

darsh-mishr:
Thanks @lkstrp, both addressed in `a88c1dee`:

1. **`pypsa/network/io.py:1811`** — replaced the manual static-drop + dynamic-column-clearing loop with a single `self.remove(cls_name, duplicated_components)`. Confirmed `n.remove` (`pypsa/network/transform.py:441-445`) clears both static rows and all dynamic columns via `inplace=True`, so the existing local `old_static` reference reflects the post-remove state without further bookkeeping. Net: 6 lines → 1.

2. **`test/test_network.py:309`** — trimmed `test_add_overwrite_clears_stale_dynamic` from 40 lines to 11. Dropped the `get_switchable_as_dense` re-check (the dynamic-column-absent assertion is sufficient, since the optimizer only reaches for the dynamic column when it exists), the verbose docstring, and the redundant intermediate assertions.

Verified locally — both `default` and `--new-components-api` modes pass the 131-test sweep across `test_network.py`, `test_components.py`, `test_components_array.py`, `test_network_transform.py`.

test/test_network.py
darsh-mishr: Replaced verbose setup, a four-line docstring, and two redundant assertions with one-line equivalents, kept only the assertion.

pypsa/network/io.py
```diff
@@ -1808,7 +1808,10 @@ def _import_components_from_df(
                 )
                 new_static = new_static.drop(duplicated_components)
             else:
-                old_static = old_static.drop(duplicated_components)
+                # Use n.remove() so both static rows and dynamic columns are
+                # cleared, making overwrite=True behave like remove() + add().
+                # See issue #1628.
+                self.remove(cls_name, duplicated_components)
```
lkstrp:
```suggestion
                self.remove(cls_name, duplicated_components)
```

lkstrp (approved):

PR #1667, opened 2026-04-30
warn about infinite resistance

lindnemi:

I noticed that sometimes the secant loss approximation would not be able to construct enough linear segments. This happened because r_pu may contain INF when s_nom is set 0. Basically, this should not happen in a well specified network, so i added some warnings

@fneum 

## Checklist

### DIFF
--- a/pypsa/network/power_flow.py
+++ b/pypsa/network/power_flow.py
@@ -15,7 +15,7 @@ from typing import TYPE_CHECKING, Any
 import networkx as nx
 import numpy as np
 import pandas as pd
-from numpy import ones, r_
+from numpy import ones, r_, isinf
 from numpy.linalg import norm
 from scipy.sparse import csc_matrix, csr_matrix, dok_matrix, issparse
 from scipy.sparse import hstack as shstack
@@ -761,6 +761,10 @@ class NetworkPowerFlowMixin(_NetworkABC):
         self.c.lines.static["r_pu"] = self.c.lines.static.r / (
             self.c.lines.static.v_nom**2
         )
+        if isinf(self.c.lines.static["r_pu"]).any():
+            logger.warning(
+                "Warning, some lines have infinite per unit resistance. This may cause issues with power flow convergence. Consider removing these lines."
+            )
         self.c.lines.static["b_pu"] = (
             self.c.lines.static.b * self.c.lines.static.v_nom**2
         )
--- a/pypsa/network/power_flow.py
+++ b/pypsa/network/power_flow.py
@@ -15,7 +15,7 @@ from typing import TYPE_CHECKING, Any
 import networkx as nx
 import numpy as np
 import pandas as pd
-from numpy import ones, r_, isinf
+from numpy import isinf, ones, r_
 from numpy.linalg import norm
 from scipy.sparse import csc_matrix, csr_matrix, dok_matrix, issparse
 from scipy.sparse import hstack as shstack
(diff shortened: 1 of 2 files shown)


pypsa/network/power_flow.py
```diff
@@ -761,6 +761,10 @@ def calculate_dependent_values(self) -> None:
         self.c.lines.static["r_pu"] = self.c.lines.static.r / (
             self.c.lines.static.v_nom**2
         )
+        if isinf(self.c.lines.static["r_pu"]).any():
+            logger.warning(
+                "Warning, some lines have infinite per unit resistance. This may cause issues with power flow convergence. Consider removing these lines."
+            )
```
fneum, line 771: Should be analogous for reactance `x_pu` and for transformers as well.

pypsa/optimization/constraints.py
```diff
@@ -2159,7 +2159,9 @@ def define_secant_loss_constraints(
         stepfactor_k = maximum(stepfactor_atol, stepfactor_rtol)
         breakpoint_factors_list.append(breakpoint_factors_list[-1] * stepfactor_k)
         if k >= max_segments:
-            msg = f"Secant loop hit max_segments; check atol/rtol or line parameters; current inputs would result in {2 * max_segments} additional constraints per line"
+            msg = f"Secant loop hit max_segments; check atol/rtol or line parameters; current inputs would result in {2 * max_segments} additional constraints per line."
+            if isinf(r_pu_eff).any():
+                msg += " Hint: Found Inf in r_pu_eff values, which may indicate a line with zero capacity, which should probably be removed from the network."
```
fneum:
```suggestion
                msg += " Identified infinite `r_pu_eff` in lines likely caused by a line with `s_nom=0`. Consider removing this line."
```

fneum (changes requested):
Looks good, should just extend to reactance and transformers as well while we are at it.

pypsa/optimization/constraints.py
```diff
@@ -2159,7 +2159,9 @@ def define_secant_loss_constraints(
         stepfactor_k = maximum(stepfactor_atol, stepfactor_rtol)
         breakpoint_factors_list.append(breakpoint_factors_list[-1] * stepfactor_k)
         if k >= max_segments:
-            msg = f"Secant loop hit max_segments; check atol/rtol or line parameters; current inputs would result in {2 * max_segments} additional constraints per line"
+            msg = f"Secant loop hit max_segments; check atol/rtol or line parameters; current inputs would result in {2 * max_segments} additional constraints per line."
+            if isinf(r_pu_eff).any():
+                msg += " Identified an infinite `r_pu_eff` in lines likely caused by a line with `s_nom=0`. Consider removing this line."
```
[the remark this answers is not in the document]
fneum: @lindnemi, this could actually be a valid issue. Could you add a test case where we run into this infinite `r_pu_eff` check? Then, ready to merge.

lkstrp (changes requested):
Could we not merge this yet and bring it straight into my new validation layer instead? If it makes sense, we can also add a strict check there instead of just a warning. But both will work.

I will migrate all of those checks everywhere, as well as the consistency check, to that layer anyway, so I would rather not merge this before then.

lindnemi:
Would be fine by me. What is the timeline for the validation layer?

And a  more general point would be: Do we want to support non_extendable lines with s_nom=0, or forbid that outright?

fneum:
Am I reading it right that @lkstrp wanted to close this? if not, reopen.

lkstrp:
I just keep it open an misuse it as an issue. Will close once implemented in alternative PR

PR #1671, opened 2026-05-06
Feat/piecewiese update

FabianHofmann:
Update code to use latest linopy master. Some refactoring and type improvements. Added tests 

## Checklist

### DIFF
--- a/pypsa/optimization/optimize.py
+++ b/pypsa/optimization/optimize.py
@@ -269,7 +269,7 @@ def define_objective(
             if var_name not in m.variables and cost_type == "spill_cost":
                 continue
 
-            # Piecewise marginal cost via tangent constraints.
+            # Piecewise marginal cost formulation.
             p = m[var_name].sel(snapshot=sns)
             extra_options = filter(
                 lambda p: p.component == c.name and p.attribute == cost_type,
@@ -283,7 +283,7 @@ def define_objective(
                 seg_attr=cost_type,
                 aux_var_name=f"{c.name}-{cost_type}_piecewise",
                 active_names=c.active_assets,
-                operator="<=",
+                operator=">=",
                 marginal_attr=True,
                 extra_options=extra_options,
             )
@@ -351,7 +351,10 @@ def define_objective(
 
         caps = m[f"{c.name}-{attr}"]
         y_attr = "capital_cost"
-        extra_options = piecewise_options.get(y_attr, {}) if piecewise_options else {}
+        extra_options = filter(
+            lambda p: p.component == c.name and p.attribute == y_attr,
+            piecewise_options,
+        )
         seg_cc_var = define_piecewise(
             m,
             c,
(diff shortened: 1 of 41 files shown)


pypsa/optimization/optimize.py
```diff
@@ -353,20 +351,23 @@ def define_objective(
...
+        seg_cc_var = define_piecewise(
             m,
             c,
             x_var=caps,
-            pw_attr=y_attr,
+            seg_attr=y_attr,
             aux_var_name=f"{c.name}-{y_attr}_piecewise",
             active_names=ext_i,
-            operator="<=",
+            operator=">=",
```
coroa: this one looks intrusive?

pypsa/optimization/piecewise.py
```diff
@@ -89,188 +91,200 @@ def define_piecewise(
...
-            operator,
-            pw_names,
-        )
+            names, aux = seg_names, aux_var_name
+            opt_method, opt_op = option.method, option.operator
+        if names.empty:
+            continue
+        if opt_method == "lp" and opt_op == "==":
+            msg = "method 'lp' requires PyPSA operator '<=' or '>='."
+            raise ValueError(msg)
```
brynpickering: Linopy will catch this 4 lines later so it seems redundant maintenance burden to set this here

pypsa/optimization/piecewise.py
```diff
@@ -89,188 +91,200 @@ def define_piecewise(
...
-    x_breakpoints_sel = x_breakpoints.sel(name=valid_names)
-    y_breakpoints_sel = y_breakpoints.sel(name=valid_names)
-    piecewise_func = piecewise(x_var_sel, x_breakpoints_sel, y_breakpoints_sel)
-    m.add_piecewise_constraints(
-        getattr(piecewise_func, OPS[operator])(y_var_sel),
-        name=aux_var_name,
-        method=method,
-    )
+    seg_names = seg_df.columns.unique("name").intersection(active_names)
+    valid_names = [name for name in seg_names if seg_df[name].notna().any().any()]
```
brynpickering: just dropna from `seg_df` when getting `seg_names` and then you don't need this extra step.

test/test_lopf_piecewise_costs.py
brynpickering, line 1: Most of this should move to a `test_optimization_piecewise.py` module and should be unit tests applied directly on the functions in `pypsa/optimization/piecewise.py` (several are tests of the `_normalize_segments` method so should be calling that method directly)

brynpickering (changes requested):
Thanks for the contribution. Likely needs updating _again_ following my most recent pushes. I think `_normalise_segments` could be much simpler by using the method as a groupby func on the grouper `seg_df.groupby(axis=1, level="name")` (or transposed since pandas is keen for people to not group on columns any more for some reason). I'll test something locally and then make a suggestion.

FabianHofmann:
@brynpickering thanks for the review. I addressed most of them (test suite still outstanding). about the normalization with groupby, I came up with something but it is not using groupby but a "column wise" (top level) validate and sorting followed by a `pd.concat`. don't know if groupby makes it leaner than that...

FabianHofmann:
@brynpickering the doctests are failing due to the new columns in the components. something we can fix later.

brynpickering (approved):
Looks to be in good shape. We can clean up the tests as necessary in the main PR. Thanks for the updates!

PR #1683, opened 2026-05-20
fix: capacity bounds for infinite non-extendables

FabianHofmann:
## Summary
- When a non-extendable component has `p_nom=inf` and `p_min_pu`/`p_max_pu=0`, the operational constraint bound was `inf*0 = NaN`. With linopy `>=0.7`, NaN bounds are no longer silently dropped and the solve fails.
- Fall back to `0` for the bound in that case.

## Test plan
- [x] Added `test_define_fixed_operational_constraints_infinite_p_nom` asserting the lower-bound rhs is zero and the solve dispatches correctly.
- [x] Existing `fixed_operational` tests still pass.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,13 +6,16 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Release Notes
 
-<!--## Upcoming Release
+## Upcoming Release
 
 !!! info "Upcoming Release"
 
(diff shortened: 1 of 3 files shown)


tgilon:
Thanks a lot for the speedy fix!

fneum (approved):

lkstrp (approved):

Issue #1585, opened 2026-03-02
ArrowStringArray from NetCDF loading breaks optimize() on pandas >= 2.0


MaykThewessen:
## Description

Networks loaded from `.nc` files with **pandas >= 2.0** have Arrow-backed string columns and indices (`ArrowStringArray`). When these are passed to `optimize()`, xarray/linopy operations fail or produce incorrect results because `ArrowStringArray` doesn't behave identically to object-dtype strings for index operations.

This silently affects all users who:
1. Save a network with `network.export_to_netcdf()`
2. Reload it with `pypsa.Network("network.nc")`
3. Call `network.optimize()` or operate on the reloaded network

## Reproduction

```python
import pypsa
import pandas as pd

# Create and solve a simple network
n = pypsa.Network()
n.set_snapshots(pd.date_range("2025-01-01", periods=24, freq="h"))
n.add("Bus", "bus0")
n.add("Generator", "gen0", bus="bus0", p_nom=100, marginal_cost=10)
n.add("Load", "load0", bus="bus0", p_set=50)

n.optimize(solver_name="highs")
n.export_to_netcdf("test_network.nc")

# Reload — indices are now ArrowStringArray
n2 = pypsa.Network("test_network.nc")
print(type(n2.generators.index.values))  # <class 'pandas.core.arrays.string_arrow.ArrowStringArray'>

# Operations on n2 may fail or behave unexpectedly
# e.g. network.copy() + bus remapping + optimize() in redispatch patterns
```

## Current workaround

We convert Arrow strings to object dtype manually after loading:

```python
def _convert_arrow_strings(network):
    """Convert ArrowStringArray columns/indices to regular object dtype."""
    for c in network.components:
        df = c.static
        if not df.empty:
            if isinstance(df.index.values, pd.arrays.ArrowStringArray):
                c.static.index = pd.Index(df.index.astype(object))
            for col in df.columns:
                if isinstance(df[col].values, pd.arrays.ArrowStringArray):
                    c.static[col] = df[col].astype(object)
        for key in c.dynamic:
            dyn_df = c.dynamic[key]
            if isinstance(dyn_df, pd.DataFrame) and not dyn_df.empty:
                if isinstance(dyn_df.columns.values, pd.arrays.ArrowStringArray):
                    c.dynamic[key].columns = pd.Index(dyn_df.columns.astype(object))
```

[...]

PR #1687, opened 2026-05-21
fix(io): coerce Arrow-backed strings on import (closes #1585)

MaykThewessen:
Closes #1585.

## Changes proposed in this Pull Request

Under pandas with PyArrow-backed strings (`future.infer_string=True` or pandas 3.x), reloading a network from NetCDF leaves `ArrowStringArray` indices/columns on the components. `optimize()` and `copy()`-based redispatch then misbehave silently because xarray/linopy expect object-dtype strings.

Fix: a small `_coerce_arrow_strings` helper in `pypsa/network/io.py`, applied at three places:

1. `_ImporterNetCDF.get_static` / `get_series` — coerce as soon as xarray returns Arrow.
2. End of `_import_components_from_df` — the existing `df.index.astype(str)` re-introduces Arrow under `infer_string`, so coerce again before storing in `Components.static`. This also covers CSV/HDF5/Excel for free, which all share the same chokepoint.
3. End of `_import_series_from_df` — same idea for dynamic frames.

New regression test `test_netcdf_io_no_arrow_string_arrays` exports a tiny network, reloads under `future.infer_string=True`, and asserts no static index, static column, or dynamic columns are `ArrowStringArray`.

The pre-existing `test_{csv,hdf5,netcdf}_io_multiindexed` cases compare a mutated fixture (dynamic columns kept as `StringDtype` because pandas re-infers on assignment) against the coerced reload (now object dtype). Relaxed those assertions with `check_index_type=False, check_column_type=False`.

Release note added.

<details>
<summary>AI assistance note</summary>

This PR was drafted with AI assistance. The fix location, test, and PR description were reviewed and edited by me. I take responsibility for the content under the project's contribution guidelines.
</details>

## Checklist

- [x] Code changes are sufficiently documented (helper has a docstring linking the issue).
- [x] Unit tests for new features were added.
- [x] A note for the release notes `docs/release-notes.md` of the upcoming release is included.
- [x] PR description is written by me. AI assistance marked above.
- [x] I consent to the release of this PR's code under the MIT license.

PR #1690, opened 2026-05-21
fix: normalize Arrow-string indexes before optimize (#1585, #1656)

### LINKED ISSUE #1656: rolling horizon ends with error KeyError "Key '0' not found in scenarios, collection, or buses."

### Issue Description
```
Running the code below shows the following (error) messages:

INFO:pypsa.network.io:Imported network 'AC-DC-Meshed' has buses, carriers, generators, global_constraints, lines, links, loads
INFO:pypsa.consistency:Sanitizing network...
INFO:pypsa.consistency:Network sanitization complete.
INFO:pypsa.optimization.abstract:Optimizing network for snapshot horizon [2015-01-01 00:00:00:2015-01-01 09:00:00] (1/1).
WARNING:pypsa.consistency:The following lines have zero x, which could break the linear load flow:
Index(['2', '3', '4'], dtype='str', name='name')
WARNING:pypsa.consistency:The following lines have zero r, which could break the linear load flow:
Index(['0', '1', '5', '6'], dtype='str', name='name')
INFO:linopy.model: Solve problem using Highs solver
INFO:linopy.model:Solver options:
 - log_to_console: False
INFO:linopy.io: Writing time: 0.25s
INFO:linopy.constants: Optimization successful: 
Status: ok
Termination condition: optimal
Solution: 187 primals, 468 duals
Objective: 1.84e+07
Solver model: available
Solver message: Optimal

INFO:pypsa.optimization.optimize:The shadow-prices of the constraints Generator-ext-p-lower, Generator-ext-p-upper, Line-ext-s-lower, Line-ext-s-upper, Link-ext-p-lower, Link-ext-p-upper, Kirchhoff-Voltage-Law were not assigned to the network.
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[26], line 7
      3 n = pypsa.examples.ac_dc_meshed()
      4 n.sanitize()
      5 n.snapshots
      6 horizon = int(24 * days / n.snapshot_weightings['stores'].mean())
----> 7 ret, status = n.optimize.optimize_with_rolling_horizon(horizon=horizon, overlap=1, include_objective_constant=False, progress=False, log_to_console=False)

File ~/PyPSA-X/.pixi/envs/default/lib/python3.12/site-packages/pypsa/networks.py:475, in Network.__getitem__(self, key)
    473     else:
    474         msg = f"Key '{key}' not found in scenarios, collection, or buses."
--> 475         raise KeyError(msg)
    476 else:
    477     # Check for deprecated tuple usage (buses, snapshots)
    478     if isinstance(key, tuple) and len(key) == 2:

KeyError: "Key '0' not found in scenarios, collection, or buses."
```
### Reproducible Example

```python
import pypsa
horizon = 24
ret, status = n.optimize.optimize_with_rolling_horizon(horizon=horizon, overlap=1, include_objective_constant=False, progress=False, log_to_console=False)
```

### Expected Behavior

Run through without error message.

### Installed Versions

<details>
Replace this line.
</details>

MaykThewessen:
## What

Fixes the `TypeError: Invalid array type: <class 'pandas.arrays.ArrowStringArray'>` that breaks `Network.optimize()` and `optimize_with_rolling_horizon()` under pandas ≥ 2.1 with the default `future.infer_string=True`.

Closes / supersedes #1585.
Closes #1656 (rolling-horizon failure on `ac_dc_meshed`; the originally-reported `KeyError '0'` path is gone on current master, but the example still fails — at the Arrow-string boundary now, reported in #1585).

## Why

CSV-loaded component data becomes `ArrowStringArray`-backed under the new pandas default:

```python
n.c.generators.static.index.dtype           # 'str' (Arrow-backed)
type(n.c.generators.static.index.values)    # ArrowStringArray
```

xarray rejects Arrow-backed pandas arrays as `.sel()` indexers. Dozens of call sites in `pypsa/optimization/{constraints,variables,optimize,global_constraints}.py` pass `c.active_assets`, `get_extendable_i()`, `get_committable_i()`, etc. into `.sel(name=...)`. The first one to hit it is:

```
File ".../pypsa/optimization/constraints.py", line 1080, in define_nodal_balance_constraints
    cbuses = cbuses.sel(name=c.active_assets)
TypeError: Invalid array type: <class 'pandas.arrays.ArrowStringArray'>
```

Reproducer:

```python
import pandas as pd, pypsa
pd.set_option("future.infer_string", True)
n.optimize.optimize_with_rolling_horizon(horizon=4, overlap=1)
```

## How

Single helper `_normalize_static_string_dtypes(n)` in `consistency.py` that casts string-typed indexes and columns to plain `object`. Wired into two entry points:

1. `Network.sanitize()` — manual sanitize path.
2. `Network.optimize.create_model()` — implicit fix when users call `optimize()` without `sanitize()`. `Network.optimize.__call__` already routes through `create_model`, so it inherits the fix.

This is the most upstream fix that doesn't require patching dozens of `.sel(name=...)` call sites or introducing per-helper dtype casts.

## Test

New regression test `test_rolling_horizon_ac_dc_meshed_arrow_string` in `test/test_lopf_rolling_horizon.py`:

- Forces `future.infer_string=True`.
- Loads `ac_dc_meshed`, runs `sanitize()`.
- Asserts post-sanitize invariant on every component static index and every string-typed column.
- Runs `optimize_with_rolling_horizon(horizon=4, overlap=1)` end-to-end.

Verified locally:

- `test/test_lopf_rolling_horizon.py` — 9/9 pass (previously 8/9 broken on master under current pandas).
- `test/test_lopf_unit_commitment.py`, `test/test_consistency_check.py`, `test/test_components.py` — green.
- Pre-existing 3 failures in `test/test_components_array.py` (`test_as_xarray_static_with_scenarios`, `test_as_xarray_scenario_dim_order`, `test_from_xarray[ac_dc_stochastic]`) reproduce on unmodified master — unrelated to this PR.

## Notes

- Helper is module-private (`_normalize_static_string_dtypes`) — happy to make it public if the maintainers prefer.
- The dtype check uses `str(dtype) in ("str", "string")` to catch both `pd.StringDtype()` and pandas 3.0's Arrow-backed string dtype. Pure-numpy `object` columns are left untouched.
- Tested against the project use case (PyPSA-based Dutch grid model with weekly-chunk rolling solves) and against the included examples.

### DIFF
--- a/pypsa/consistency.py
+++ b/pypsa/consistency.py
@@ -49,6 +49,28 @@ def _log_or_raise(strict: bool, message: str, *args: Any) -> None:
     logger.warning(message, *args)
 
 
+def _normalize_static_string_dtypes(n: NetworkType) -> None:
+    """Cast every component's `static` string-typed index and columns to object.
+
+    Under pandas >= 2.1 with `future.infer_string=True` (the default since
+    pandas 3.0), CSV/parquet-loaded component data becomes
+    ArrowStringArray-backed. xarray rejects Arrow-backed pandas arrays as
+    `.sel()` indexers, breaking dozens of call sites in the optimization
+    module (e.g. `cbuses.sel(name=c.active_assets)` in
+    `define_nodal_balance_constraints`, where `cbuses` is built from
+    bus-column values). See GH #1585 / #1656.
+
+    Called from `Network.sanitize()` and from `Network.optimize` before model
+    construction, so the invariant holds for both manual and implicit flows.
+    """
+    for c in n.components:
+        if c.static.index.dtype != object:
+            c.static.index = c.static.index.astype(object)
+        for col in c.static.columns:
+            if str(c.static[col].dtype) in ("str", "string"):
+                c.static[col] = c.static[col].astype(object)
+
+
 def check_for_unknown_buses(
     n: NetworkType, component: Components, strict: bool = False
 ) -> None:
@@ -1045,6 +1067,8 @@ class NetworkConsistencyMixin(_NetworkABC):
 
         self.c.carriers.add_missing_carriers()
         self.c.carriers.assign_colors()
+
+        _normalize_static_string_dtypes(self)
 
         logger.info("Network sanitization complete.")
 
--- a/pypsa/optimization/optimize.py
+++ b/pypsa/optimization/optimize.py
@@ -21,7 +21,11 @@ from pypsa._options import options
 from pypsa.common import UnexpectedError, as_index
 from pypsa.components.array import _from_xarray
 from pypsa.components.common import as_components
-from pypsa.consistency import check_big_m_exceeded, check_no_modular_committables
+from pypsa.consistency import (
+    _normalize_static_string_dtypes,
+    check_big_m_exceeded,
+    check_no_modular_committables,
+)
 from pypsa.descriptors import nominal_attrs
 from pypsa.guards import _assert_data_integrity
 from pypsa.optimization.abstract import OptimizationAbstractMixin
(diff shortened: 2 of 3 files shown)


FabianHofmann (changes requested):
@MaykThewessen thanks for the contribution. I would say, let's fix the data model right from the start on the IO level: move the helper function to the io module and run it after the read in

FabianHofmann:
> @MaykThewessen thanks for the contribution. I would say, let's fix the data model right from the start on the IO level: move the helper function to the io module and run it after the read in

seeing that #1687 does exactly that (also yours). please don't offload this kind of bookkeeping maintenance work to us.

PR #1692, opened 2026-05-21
fix(optimization): include static phase_shift in cycle KVL

### LINKED ISSUE #1220: Update needed on tutorial example "Transformer with non-trivial phase shift and tap ratio"

MaykThewessen:
Closes #1220.

Static `phase_shift` on `Transformer` components is now included in the cycle-based Kirchhoff Voltage Law constraint built by `n.optimize()`. Before this fix the term was silently dropped in LOPF, so an `n.optimize()` solution and a subsequent `n.lpf()` / `n.pf()` verification on the same network gave different flows whenever any transformer had a non-zero `phase_shift`. `n.lpf()` and `n.pf()` already handled the term correctly; this just brings LOPF into line with them.

The cycle KVL around each cycle `k` becomes

    sum_l C_{l,k} * (x_pu_l * P_l + phi_l_rad) = 0

where `phi_l` is zero for lines and for transformers without a phase shift. Static phase shifts contribute a constant on the RHS via the unweighted cycle matrix, so the existing weighted-cycle LHS code is untouched. No new attributes, no new variables.

## Tests

Four new cases in `test/test_phase_shift_static_kvl.py` on a 2-bus / parallel-path network with equal per-unit reactance:

- Zero phase shift → equal 50/50 split (regression guard for the unaffected path).
- +10° → flow split matches the analytical value `P_L - P_T = phi_rad / x_pu`.
- −10° → exact mirror of the +10° case (sign symmetry).
- LOPF flows match a subsequent `n.lpf()` on the same network (the bug this PR fixes).

## Checklist

- [x] Code changes are sufficiently documented; new code has docstring + inline comment explaining the cycle-direction sign.
- [x] Unit tests for new features were added.
- [x] PR description is written by me.

## Note on PR #1661

The variable / extendable phase-shift feature from PR #1661 (issue #456) is being reworked separately into Option B (`phase_shift_{min,max,set}` writing back into `phase_shift`, no `_extendable` flag) per @fneum's review. This bugfix is split out so it can land on its own merits.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,7 +14,11 @@ SPDX-License-Identifier: CC-BY-4.0
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
 
+### Bug Fixes
+
 - Fix operational constraints for non-extendable components producing `NaN` bounds when `p_nom` is infinite and `p_min_pu`/`p_max_pu` is zero. The bound now falls back to zero in this case. Relevant for linopy versions `>=0.7` where `NaN` bounds are not dropped explicitly.
+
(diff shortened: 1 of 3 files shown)


FabianHofmann:
to ease bookkeeping, let's focus on #1661 and bring that one on board

PR #1694, opened 2026-05-21
feat: add Lines.apply_seasonal_rating helper

### LINKED ISSUE #1693: Feature request: helper for seasonal line ratings
### Use case

Transmission line ampacity varies by season — most TSOs publish at least two thermal periods (e.g. TenneT's *zomer*-periode 1 Apr–30 Sep / *winter*-periode 1 Oct–31 Mar; RTE has four; National Grid has summer/winter; CAISO has four periods). For long-horizon DCOPF runs, this typically derates summer capacity 10–20% on conductor-limited lines and shows up as binding constraints in summer evenings.

PyPSA today exposes `lines_t.s_max_pu` (per-snapshot fraction of `s_nom`), so the building block exists, but everyone who needs seasonal derating ends up writing the same broadcast logic:

```python
n.lines.at[line, "s_nom"] = max(summer_mva, winter_mva)
factor = min(summer_mva, winter_mva) / max(summer_mva, winter_mva)
summer_mask = n.snapshots.month.isin([4, 5, 6, 7, 8, 9])
n.lines_t.s_max_pu[line] = existing * np.where(summer_mask, factor, 1.0)
```

### Proposed API

```python
def apply_seasonal_line_ratings(
    n: Network,
    ratings: pd.DataFrame,
    summer_months: Sequence[int] = (4, 5, 6, 7, 8, 9),
    *,
    compose: bool = True,
) -> None:
    """Apply seasonal s_max_pu derating to lines based on snapshot month.

    Parameters
    ----------
    n : Network
        Network with a DatetimeIndex on n.snapshots.
    ratings : pd.DataFrame
        Index: line names (subset of n.lines.index).
        Columns: 'summer', 'winter' — per-line MVA ratings.
    summer_months : Sequence[int], default (4..9)
        Snapshot months treated as summer. Northern-hemisphere default;
        flip for southern hemisphere (e.g. (10, 11, 12, 1, 2, 3)).
    compose : bool, default True
        If True, multiply into existing s_max_pu (preserves N-1 margins).
        If False, overwrite.

    Behaviour
    ---------
    For each line in `ratings`:
    - n.lines.at[line, 's_nom'] is set to max(summer, winter) (winter envelope).
    - n.lines_t.s_max_pu[line] gets the seasonal factor
      min(summer, winter) / max(summer, winter) on summer months, else 1.0.
    """
```

Hemisphere-agnostic: caller controls `summer_months`. Convention: summer = period with the lower rating (cooling-limited), matching IEEE 738 / IEC 61597.

### Suggested home

`pypsa/utils/thermal_ratings.py` with a re-export on `pypsa.Network.apply_seasonal_line_ratings` (mixin pattern, similar to `optimize`).

Alternative: pure function in `pypsa.utils`, no method binding.

### Why upstream rather than a recipe in docs

It's small (~30 LOC + tests) but the composition semantics with N-1 margins, the "winter envelope" choice, and the per-snapshot broadcast are the kind of details that everyone gets slightly wrong on their first try. A canonical implementation in core saves a lot of subtle bug-hunting downstream.

### Scope boundaries (what this is NOT)

- **Not** a JAO/CSV loader — `ratings` is a `DataFrame` the user assembles.
- **Not** an upgrade detector — caller decides which lines to feed in.
- **Not** dynamic line rating (weather-dependent) — that's a separate, larger PR I'm scoping (atlite/IEEE 738 with wind speed and ambient temp).

### Real-world usage

Currently maintained out-of-tree in [`psa_mee`](https://github.com/MaykThewessen/PyPSA-MEE) (Dutch HV grid model). The local module also handles JAO auto-discovery and upgrade-detection — the policy layer would stay local; only the per-snapshot broadcast mechanism is what I'd like to move upstream.

Happy to open a PR with implementation + tests + docs page if the scope and home look right. Pinging @FabianHofmann / @fneum for direction.

MaykThewessen:
Closes #1693 

## What

`n.c.lines.apply_seasonal_rating(ratings, summer_months=(4..9), *, compose=True)` — a method on the `Lines` component that broadcasts a per-line `(summer, winter)` MVA rating table onto `n.lines_t.s_max_pu` based on snapshot month. Implements the API proposed in #1693.

## Mechanism

Policy-light, mechanism-only.

- For each line in `ratings`, per-snapshot `s_max_pu = rating / s_nom`: the `summer` rating on `summer_months`, the `winter` rating on all other snapshots.
- **`s_nom` is left unchanged.** The seasonal limit is expressed purely as an `s_max_pu` scaling of the existing `s_nom`, so a rating above `s_nom` yields `s_max_pu > 1`.
- Each season uses its own column, so the helper makes no assumption about which period is lower: a winter-lower dataset is handled correctly. The published-TSO convention (summer = lower, cooling-limited) is documented but not enforced.
- `compose=True` (default) multiplies the seasonal scaling into any pre-existing `n.lines_t.s_max_pu` so N-1 margins / other dynamic limits survive intact; `compose=False` overwrites.
- `summer_months` is caller-controlled: northern-hemisphere default, pass `(10, 11, 12, 1, 2, 3)` for the southern hemisphere.
- `ratings` is a `DataFrame` the caller assembles: no CSV / JAO / TSO-specific loaders in PyPSA.

Out of scope (separate work): JAO/TSO data loaders, grid-upgrade detection (caller filters before calling), and weather-dependent Dynamic Line Rating.

## Location

Lives as `Lines.apply_seasonal_rating` in `pypsa/components/_types/lines.py`, invoked as `n.c.lines.apply_seasonal_rating(...)`. (Earlier revisions used a top-level `pypsa.apply_seasonal_line_ratings` function under a new `pypsa/utils/` namespace; relocated to the component method per review.)

## Validation

- `TypeError` if `n.snapshots` is not a `DatetimeIndex`.
- `KeyError` if `ratings` is missing the `summer` / `winter` columns, or references a line not in `n.lines.index`.
- `ValueError` if any rating is non-positive, `summer_months` is empty, or a rated line has a non-positive `s_nom`.

## Tests

`test/test_seasonal_line_ratings.py` — 16 cases across 8 fixture-based, parametrized tests:

- Seasonal scaling: summer-lower, winter-lower, equal ratings, rating-above-`s_nom`, SH override.
- Compose: multiply into a static N-1 margin, `compose=False` overwrite, element-wise compose with a pre-existing dynamic per-snapshot `s_max_pu`.
- Partial subset (unrelated lines untouched).
- Validation errors: missing column, unknown line, non-positive rating, empty `summer_months`, non-datetime snapshots, non-positive `s_nom`.
- End-to-end `optimize()` solves to `optimal`.

## Docs + release notes

- New user-guide page `docs/user-guide/seasonal-line-ratings.md` (registered in `mkdocs.yml` nav): motivation, minimal example, compose semantics with an N-1 margin, SH override, and the summer-is-lower convention.
- Features entry in `docs/release-notes.md`.

## Since the last review (@FabianHofmann)

- Relocated the helper to the `Lines.apply_seasonal_rating` method (was a top-level `pypsa/utils` function).
- `s_nom` no longer overwritten with the `max(summer, winter)` envelope: scales `s_max_pu` only.
- Test file slimmed, renamed to `test_seasonal_line_ratings.py`, and parametrized around a shared network fixture.
- Docs wording reworked; "Out of scope" section dropped from the page.
- Dropped the temporary `future.infer_string=False` workaround in the solve test: the Arrow-string fix landed upstream in v1.2.4 (#1687), so `optimize()` now runs on the pandas >= 3.0 default path.

## Usage

```python
import pandas as pd
import pypsa

n = pypsa.Network(...)
ratings = pd.DataFrame({"summer": [...], "winter": [...]}, index=line_names)

n.c.lines.apply_seasonal_rating(ratings)                                        # northern hemisphere
n.c.lines.apply_seasonal_rating(ratings, summer_months=(10, 11, 12, 1, 2, 3))   # southern hemisphere
n.optimize()
```


MaykThewessen:
Docs page added in 0e5f8aff; happy to relocate if a different section fits better.

FabianHofmann:
thanks for the pr @MaykThewessen ; first question is whether we want such opinionated functions (we have to think about it). if so, I would like to place them at the Line component, such that one would call `n.c.lines.apply_seasonal_rating`.  I will come back to this later next week, have to do some work in linopy first.

MaykThewessen:
Thanks @FabianHofmann. Pushed 01abad6 relocating the helper to a method on the `Lines` component:

```python
n.c.lines.apply_seasonal_rating(ratings)
```

Changes:
- New method `Lines.apply_seasonal_rating(ratings, summer_months=..., *, compose=True)` in `pypsa/components/_types/lines.py` (drops the explicit `n` parameter; uses `self.n_save` / `self.static` / `self.dynamic`).
- Removed `pypsa.utils.thermal_ratings` module and `pypsa.utils` package (sole export).
- Removed top-level `pypsa.apply_seasonal_line_ratings`.
- Tests + docs migrated to method form.

Happy to revisit the broader \"do we want this helper at all\" question whenever you've got bandwidth after the linopy work. The narrow API surface (one method, opinionated only about NH summer default + envelope semantics) should make the keep/drop call cheaper.

docs/user-guide/seasonal-line-ratings.md
```diff
@@ -0,0 +1,85 @@
+<!--
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Seasonal Line Ratings
+
+Transmission lines have different thermal capacities in summer and winter, because conductor ampacity is cooling-limited. TenneT, RTE, National Grid, and CAISO all publish per-line summer / winter ratings derived from IEEE 738 / IEC 61597 conductor models.
```
FabianHofmann:
```suggestion
Transmission lines have different thermal capacities in summer and winter, because conductor ampacity is cooling-limited. Grid operators usually publish per-line summer / winter ratings, often derived from IEEE conductor models.
```

docs/user-guide/seasonal-line-ratings.md
```diff
@@ -0,0 +1,85 @@
...
+
+The [`Lines.apply_seasonal_rating`][pypsa.components._types.lines.Lines.apply_seasonal_rating] method broadcasts a per-line `(summer, winter)` rating table onto `n.lines.s_nom` and `n.lines_t.s_max_pu` so the seasonal envelope is visible to the optimiser.
+
+## When to use it
+
+- A TSO publishes per-line summer / winter MVA ratings and you want them in the dispatch problem.
+- You already model an N-1 margin via `n.lines.s_max_pu` and want to compose a seasonal factor on top of it.
+- You want a small, dependency-free utility, not a JAO loader or DLR engine.
+
+## Minimal example
```
FabianHofmann:
```suggestion
As illustrated below, use the function in an already initialized network with date-time indexed snapshots. 
```

docs/user-guide/seasonal-line-ratings.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Out of scope
+
+The method is intentionally policy-light:
+
+- It does not load JAO / TenneT / RTE feeds; the caller assembles `ratings`.
+- It does not detect grid upgrades or replace ratings when a line is reinforced.
+- It does not implement dynamic line rating (DLR) from weather time series.
+
+These belong upstream of the method. Pass in whatever rating table your data pipeline produces.
```
FabianHofmann: remove that section

docs/user-guide/seasonal-line-ratings.md
```diff
@@ -0,0 +1,85 @@
...
+SPDX-FileCopyrightText: PyPSA Contributors
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Seasonal Line Ratings
+
+Transmission lines have different thermal capacities in summer and winter, because conductor ampacity is cooling-limited. TenneT, RTE, National Grid, and CAISO all publish per-line summer / winter ratings derived from IEEE 738 / IEC 61597 conductor models.
+
+The [`Lines.apply_seasonal_rating`][pypsa.components._types.lines.Lines.apply_seasonal_rating] method broadcasts a per-line `(summer, winter)` rating table onto `n.lines.s_nom` and `n.lines_t.s_max_pu` so the seasonal envelope is visible to the optimiser.
```
FabianHofmann:
```suggestion
The [`Lines.apply_seasonal_rating`][pypsa.components._types.lines.Lines.apply_seasonal_rating] method broadcasts a per-line `(summer, winter)` scaling to `n.lines_t.s_max_pu` to properly reflect on the two seasonal operation modes.
```

pypsa/components/_types/lines.py
```diff
@@ -94,6 +97,128 @@ def add(
...
+
+        if compose:
+            existing = self.dynamic.s_max_pu.reindex(
+                index=n.snapshots, columns=ratings.index
+            )
+            static = self.static.loc[ratings.index, "s_max_pu"].to_numpy()
+            existing = existing.fillna(pd.Series(static, index=ratings.index))
+            self.dynamic.s_max_pu[ratings.index] = (existing * seasonal).to_numpy()
+        else:
+            self.dynamic.s_max_pu[ratings.index] = seasonal.to_numpy()
```
FabianHofmann, line 230: adjust the code to not alter `s_nom` but only adjust the scaling `s_max_pu`

test/test_thermal_ratings.py
```diff
@@ -0,0 +1,190 @@
+# SPDX-FileCopyrightText: PyPSA Contributors
```
FabianHofmann: this test file is too heavy. initialize the network as a fixture and reuse that for all non-inplace changes test, use parametrization where possible. rename the file to test_seasonal_line_ratings

FabianHofmann (changes requested):
thanks @MaykThewessen, here are some more requests. let me know if I should take over of some. The important bit is that `s_nom` is not changed inplace by applying the seasonal ratings. they should only apply the scaling to `s_max_pu`.

FabianHofmann:
> thanks @MaykThewessen, here are some more requests. let me know if I should take over of some. The important bit is that `s_nom` is not changed inplace by applying the seasonal ratings. they should only apply the scaling to `s_max_pu`.
Also could you confirm that TSOs report their rating in absolute capacity terms and not per-unit max load factors?

MaykThewessen:
Thanks @FabianHofmann, all addressed:

**`s_nom` is no longer mutated.** The helper now writes only `n.lines_t.s_max_pu`, scaling the existing `s_nom` by `rating / s_nom` per season. Nothing on `n.lines` is changed in place.

**On absolute vs per-unit:** confirmed, TSOs publish ratings in absolute capacity (MVA), not per-unit load factors. TenneT, RTE and National Grid all publish per-line summer/winter MVA. So the input columns stay absolute MVA and the conversion to a scaling is `rating / s_nom` against the unchanged `s_nom`. A rating above `s_nom` therefore yields `s_max_pu > 1`, which is the correct representation when a published seasonal rating exceeds the nominal value; I added a guard that raises if a rated line has a non-positive `s_nom`.

A side effect of dropping the envelope/min-max logic: each season now takes its own column directly, so the helper no longer assumes summer is the lower rating (a winter-lower dataset is handled correctly).

Docs: intro and method description reworded, "When to use it" + "Minimal example" collapsed into the single sentence, "Out of scope" removed, `s_nom` wording corrected. Tests: renamed to `test_seasonal_line_ratings.py`, the network is now a fixture (fresh per mutating test, shared for the validation cases that raise before mutating), and the scaling + input-validation cases are parametrized.

Also merged current `master` in, so the branch is up to date; that pulled in the arrow-string fix (#1687), so I dropped the now-redundant `future.infer_string` workaround from the solve test.

Happy for you to take any of it if you'd prefer.

FabianHofmann:
great @MaykThewessen ; I plan to take another look until tomorrow afternoon

FabianHofmann:
@MaykThewessen btw as I know you love claude code and use it quite frequently, here a skill that I recently came up with that can ease our lives in the review process. it is called `compactify` and aims at making code changes leaner. store it under the `.claude/skills/compactify/SKILL.md`

[SKILL.md](https://github.com/user-attachments/files/29538889/SKILL.md)

beyond that, we will come up with a more distributable system on LLM usage and their conventions

FabianHofmann:
and can you update the pr summary please?

MaykThewessen:
@FabianHofmann summary updated to match the current code. There's a **Since the last review** section at the bottom capturing what moved since you last looked: helper relocated to the `Lines.apply_seasonal_rating` method, `s_nom` left untouched (scales `s_max_pu` only), test file slimmed/renamed to `test_seasonal_line_ratings.py` + parametrized around a shared fixture, docs reworded, and the temporary `future.infer_string=False` workaround dropped now that the arrow-string fix is upstream in v1.2.4 (#1687).

I also resolved the two threads that were already addressed but still showing open. Ready whenever you get to it tomorrow.

And thanks for the `compactify` skill, added it to my setup.

FabianHofmann (approved):
I think this is good to go. It is not *really* needed but nice-to-have, and finally we leverage the ability of defining component specific functions.

MaykThewessen:
> **Note:** AI-assisted (written with Claude).

Merged current `master` in again (7ec98f34) to clear the two remaining blockers:

- **Release-notes conflict** with #1909, which landed right after the last master merge and adds a line under the same `### Features` heading. Kept both entries.
- **Docs check**: the doctest `n.lines_t.s_max_pu['a-b'].iloc[3000]` printed `np.float64(0.8)` under numpy 2, so it now reads `float(...)` like the `s_nom` line above it.

Checked locally: the `apply_seasonal_rating` doctest runs 9/9 examples clean against the branch, and the branch is 0 commits behind `master`. No change to the method itself.

fneum (approved):
Fine.

PR #1699, opened 2026-05-22
rolling horizon: track e_sum_max constraints

fneum:
Rolling horizon optimisation now tracks Generator `e_sum_max` budgets across iterations by reducing the remaining budget in each window by the energy already committed in previous windows (excluding overlap that is re-optimised).

### DIFF
--- a/pypsa/optimization/abstract.py
+++ b/pypsa/optimization/abstract.py
@@ -498,6 +498,10 @@ class OptimizationAbstractMixin(OptimizationAbstractMGAMixin):
     ) -> Network:
         """Optimizes the network in a rolling horizon fashion.
 
+        Generator `e_sum_max` budgets are tracked across iterations: the remaining
+        budget in each window is reduced by the weighted energy already produced
+        in previously committed snapshots (excluding overlap that is re-optimised).
+
         Parameters
         ----------
         snapshots : list-like
@@ -519,6 +523,10 @@ class OptimizationAbstractMixin(OptimizationAbstractMGAMixin):
             raise ValueError(msg)
 
         starting_points = range(0, len(snapshots), horizon - overlap)
+        track_e_sum_max = (n.c.generators.static.e_sum_max < np.inf).any()
+        if track_e_sum_max:
+            e_sum_max_orig = n.c.generators.static.e_sum_max.copy()
+
         for i, start in enumerate(starting_points):
             end = min(len(snapshots), start + horizon)
             sns = snapshots[start:end]
@@ -541,6 +549,15 @@ class OptimizationAbstractMixin(OptimizationAbstractMGAMixin):
                             snapshots[start - 1]
                         ].values
                     )
+                if track_e_sum_max:
+                    committed = snapshots[:start]
+                    weights = n.snapshot_weightings.generators.loc[committed]
+                    generated = (
+                        n.c.generators.dynamic.p.loc[committed]
+                        .multiply(weights, axis=0)
+                        .sum()
+                    )
+                    n.c.generators.static["e_sum_max"] = e_sum_max_orig - generated
 
             status, condition = n.optimize(sns, **kwargs)
             if status != "ok":
@@ -549,6 +566,9 @@ class OptimizationAbstractMixin(OptimizationAbstractMGAMixin):
                     status,
                     condition,
                 )
+
+        if track_e_sum_max:
+            n.c.generators.static["e_sum_max"] = e_sum_max_orig
         return n
 
     def optimize_and_run_non_linear_powerflow(
--- a/test/test_lopf_rolling_horizon.py
+++ b/test/test_lopf_rolling_horizon.py
@@ -222,6 +222,46 @@ def test_rolling_horizon_committable_overlap_matches_full_run():
     assert (ramping >= -static.eval("ramp_limit_down * p_nom_opt")).all().all()
 
 
+def test_rolling_horizon_e_sum_max():
+    n = get_network(committable=False)
+
+    # Total unconstrained dispatch for coal in a single-shot run as reference
+    n_ref = n.copy()
+    n_ref.optimize()
+    full_coal = n_ref.c.generators.dynamic.p["coal"].sum()
+
+    # Cap coal generation below the unconstrained level
+    budget = 0.5 * full_coal
+    n.c.generators.static.loc["coal", "e_sum_max"] = budget
+
+    n.optimize.optimize_with_rolling_horizon(horizon=3)
+
+    weights = n.snapshot_weightings.generators
+    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
+    assert coal_total <= budget + 1e-6
+
+    # Static attribute is restored after the run
+    assert n.c.generators.static.loc["coal", "e_sum_max"] == budget
+
+
+def test_rolling_horizon_e_sum_max_with_overlap():
+    n = get_network(committable=False)
+
+    n_ref = n.copy()
+    n_ref.optimize()
+    full_coal = n_ref.c.generators.dynamic.p["coal"].sum()
+
+    budget = 0.5 * full_coal
+    n.c.generators.static.loc["coal", "e_sum_max"] = budget
+
+    n.optimize.optimize_with_rolling_horizon(horizon=3, overlap=1)
+
+    weights = n.snapshot_weightings.generators
+    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
+    assert coal_total <= budget + 1e-6
+    assert n.c.generators.static.loc["coal", "e_sum_max"] == budget
+
+
 def test_rolling_horizon_linearized_uc_with_ramp_limits():
     """
     Test rolling horizon with linearized UC and ramp limits on committables.
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,8 @@ SPDX-License-Identifier: CC-BY-4.0
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
 
+- [`optimize_with_rolling_horizon`][pypsa.optimization.OptimizationAccessor.optimize_with_rolling_horizon] now tracks Generator `e_sum_max` budgets across iterations by reducing the remaining budget in each window by the energy already committed in previous windows (excluding overlap that is re-optimised).
+
 - Fix operational constraints for non-extendable components producing `NaN` bounds when `p_nom` is infinite and `p_min_pu`/`p_max_pu` is zero. The bound now falls back to zero in this case. Relevant for linopy versions `>=0.7` where `NaN` bounds are not dropped explicitly.
 
 - Lift `xarray<2026.4` upper bound and bump `linopy>=0.7.0` floor. The xarray API change (rejecting `Dataset` as `data_vars` to the `Dataset()` constructor) was fixed in [linopy PR #647](https://github.com/PyPSA/linopy/pull/647), released in linopy 0.7.0. With the new floor, the xarray cap is no longer needed.


pypsa/optimization/abstract.py
```diff
@@ -541,6 +549,15 @@ def optimize_with_rolling_horizon(
...
                     )
+                if track_e_sum_max:
+                    committed = snapshots[:start]
+                    weights = n.snapshot_weightings.generators.loc[committed]
+                    generated = (
+                        n.c.generators.dynamic.p.loc[committed]
+                        .multiply(weights, axis=0)
+                        .sum()
+                    )
+                    n.c.generators.static["e_sum_max"] = e_sum_max_orig - generated
```
Irieo, line 560: How likely is situation that (with lower feasibility tolerance settings or in MILP with  `committable=True`) we face `e_sum_max_orig - generated` < 0 causing an issue here? `.clip(lower=0)` ?

pypsa/optimization/abstract.py
```diff
@@ -519,6 +523,10 @@ def optimize_with_rolling_horizon(
             raise ValueError(msg)
 
         starting_points = range(0, len(snapshots), horizon - overlap)
+        track_e_sum_max = (n.c.generators.static.e_sum_max < np.inf).any()
```
Irieo, line 526: What about `e_sum_min`? We treat it symmetrically in constraints.py.
I see with a quick test that e_sum_min (if ignored as now) applies to each window, which doesn't make sense to ignore if we start tracking e_sum_max already. 
I think you can add it in a similar minimalistic implementation as  `e_sum_min_new = max(0, e_sum_min_orig - generated)`

pypsa/optimization/abstract.py
```diff
@@ -519,6 +523,10 @@ def optimize_with_rolling_horizon(
...
+        if track_e_sum_max:
+            e_sum_max_orig = n.c.generators.static.e_sum_max.copy()
```
Irieo, line 528: Just to flag that I recall mypy gets happier if such parameters are initialized to None outside of if terms; but I see in this case it is defined once here and never mutated. @lkstrp will have a definite opinion on such things.

pypsa/optimization/abstract.py
```diff
@@ -549,6 +566,9 @@ def optimize_with_rolling_horizon(
                     status,
                     condition,
                 )
+
+        if track_e_sum_max:
+            n.c.generators.static["e_sum_max"] = e_sum_max_orig
```
Irieo, line 571: Note that we use attribute assignments (n.c.stores.static.e_initial) within same function few lines above.

test/test_lopf_rolling_horizon.py
```diff
@@ -222,6 +222,46 @@ def test_rolling_horizon_committable_overlap_matches_full_run():
     assert (ramping >= -static.eval("ramp_limit_down * p_nom_opt")).all().all()
 
 
+def test_rolling_horizon_e_sum_max():
```
Irieo, line 225: I can think of two cases that I'd add to these tests:

- More than one generator (say both coal and gas) with own budgets, to o make sure there's no issues with multiple e_sum_max arithmetic.
- weights != 1 for .multiply(weights, axis=0) in generated volume track parameter

Irieo (changes requested):
Such a joy to tinker in PRs with +62 -0 :)

pypsa/optimization/abstract.py
MaykThewessen, line 528: Combines all four `abstract.py` review threads (track `e_sum_min` symmetrically, init outside the `if` for mypy, attribute-style assignment is handled below):

```suggestion
        track_e_sum_max = (n.c.generators.static.e_sum_max < np.inf).any()
        track_e_sum_min = (n.c.generators.static.e_sum_min > -np.inf).any()
        e_sum_max_orig = n.c.generators.static.e_sum_max.copy()
        e_sum_min_orig = n.c.generators.static.e_sum_min.copy()
```

Rationale: always copying both Series is cheap (one float64 vector per generator) and lets mypy see the names unconditionally. Tracking `e_sum_min` here is required so it doesn't get re-applied per window.

pypsa/optimization/abstract.py
MaykThewessen, line 560: Addresses both the `.clip(lower=0)` numerical-safety thread and the `e_sum_min` symmetry thread in one block:

```suggestion
                if track_e_sum_max or track_e_sum_min:
                    committed = snapshots[:start]
                    weights = n.snapshot_weightings.generators.loc[committed]
                    generated = (
                        n.c.generators.dynamic.p.loc[committed]
                        .multiply(weights, axis=0)
                        .sum()
                    )
                    if track_e_sum_max:
                        n.c.generators.static.e_sum_max = (
                            e_sum_max_orig - generated
                        ).clip(lower=0)
                    if track_e_sum_min:
                        n.c.generators.static.e_sum_min = (
                            e_sum_min_orig - generated
                        ).clip(lower=0)
```

Why clip both:
- `e_sum_max`: tight feasibility tolerances or MILP rounding can leave `generated` marginally above the original budget by O(1e-6); without clipping, the next window receives a negative RHS and becomes infeasible.
- `e_sum_min`: clipping at zero is the natural semantics; once the minimum has been met by past windows, the remaining floor is zero, not a negative value the LP would treat as "you may produce up to -X units less than nothing".

For generators where the original limit is `±inf`, the corresponding entry in `e_sum_{max,min}_orig` is `±inf`, `(±inf - finite)` stays `±inf`, and clipping the `-inf` case to `0` is harmless because generator dispatch is `≥ 0` anyway.

pypsa/optimization/abstract.py
MaykThewessen, line 571: Switches to attribute-style assignment to match the convention used a few lines above for `n.c.stores.static.e_initial` / `n.c.storage_units.static.state_of_charge_initial`, and restores `e_sum_min` symmetrically:

```suggestion
        if track_e_sum_max:
            n.c.generators.static.e_sum_max = e_sum_max_orig
        if track_e_sum_min:
            n.c.generators.static.e_sum_min = e_sum_min_orig
```

test/test_lopf_rolling_horizon.py
```diff
@@ -222,6 +222,46 @@ def test_rolling_horizon_committable_overlap_matches_full_run():
...
+
+    budget = 0.5 * full_coal
+    n.c.generators.static.loc["coal", "e_sum_max"] = budget
+
+    n.optimize.optimize_with_rolling_horizon(horizon=3, overlap=1)
+
+    weights = n.snapshot_weightings.generators
+    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
+    assert coal_total <= budget + 1e-6
+    assert n.c.generators.static.loc["coal", "e_sum_max"] == budget
```
MaykThewessen, line 262: Two extra cases to cover @Irieo's requested scenarios: multiple generators with independent budgets, and non-unity `snapshot_weightings` so the weighted sum branch in `abstract.py` is actually exercised. Also adds a minimal `e_sum_min` test mirroring the symmetric handling proposed in `abstract.py`.

```suggestion
        assert n.c.generators.static.loc["coal", "e_sum_max"] == budget


def test_rolling_horizon_e_sum_max_multi_generator():
    """Both generators are independently budgeted and both budgets must hold."""
    n = get_network(committable=False)

    n_ref = n.copy()
    n_ref.optimize()
    full_coal = n_ref.c.generators.dynamic.p["coal"].sum()
    full_gas = n_ref.c.generators.dynamic.p["gas"].sum()

    coal_budget = 0.5 * full_coal
    gas_budget = 0.8 * full_gas
    n.c.generators.static.loc["coal", "e_sum_max"] = coal_budget
    n.c.generators.static.loc["gas", "e_sum_max"] = gas_budget

    n.optimize.optimize_with_rolling_horizon(horizon=3)

    weights = n.snapshot_weightings.generators
    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
    gas_total = (n.c.generators.dynamic.p["gas"] * weights).sum()
    assert coal_total <= coal_budget + 1e-6
    assert gas_total <= gas_budget + 1e-6
    assert n.c.generators.static.loc["coal", "e_sum_max"] == coal_budget
    assert n.c.generators.static.loc["gas", "e_sum_max"] == gas_budget


def test_rolling_horizon_e_sum_max_weighted_snapshots():
    """Snapshot weights != 1 must be reflected in the budget bookkeeping."""
    n = get_network(committable=False)
    n.snapshot_weightings.loc[:, :] = 2.0

    n_ref = n.copy()
    n_ref.optimize()
    weights = n_ref.snapshot_weightings.generators
    full_coal = (n_ref.c.generators.dynamic.p["coal"] * weights).sum()

    budget = 0.5 * full_coal
    n.c.generators.static.loc["coal", "e_sum_max"] = budget

    n.optimize.optimize_with_rolling_horizon(horizon=3)

    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
    assert coal_total <= budget + 1e-6
    assert n.c.generators.static.loc["coal", "e_sum_max"] == budget


def test_rolling_horizon_e_sum_min():
    """Symmetric handling: a floor on total energy must hold across windows."""
    n = get_network(committable=False)

    floor = 3000.0
    n.c.generators.static.loc["coal", "e_sum_min"] = floor

    n.optimize.optimize_with_rolling_horizon(horizon=3)

    weights = n.snapshot_weightings.generators
    coal_total = (n.c.generators.dynamic.p["coal"] * weights).sum()
    assert coal_total >= floor - 1e-6
    assert n.c.generators.static.loc["coal", "e_sum_min"] == floor
```

MaykThewessen (commented):
Drive-by suggestions consolidating @Irieo's five threads into one coherent diff:

- **`.clip(lower=0)`** on the per-window budget update (numerical safety for tight feasibility / MILP rounding).
- **Symmetric `e_sum_min` tracking**. Without it, the floor is silently re-applied per window, which is what @Irieo's quick test surfaced.
- **Init `e_sum_max_orig` / `e_sum_min_orig` unconditionally** at the top: costs nothing (one `Series.copy()` per side), keeps mypy happy, and lets the loop body branch on a single flag.
- **Attribute-style assignment** (`n.c.generators.static.e_sum_max = ...`) to match the convention used a few lines above for stores / storage units.
- **Tests**: two extra `e_sum_max` cases (multi-generator budgets, non-unity `snapshot_weightings`) plus a minimal `e_sum_min` test mirroring the symmetric handling.

Suggestions are individually applyable. Feel free to discard the symmetry / clip if you'd rather split `e_sum_min` into a follow-up PR; flagging here mainly so the two threads stay linked.

FabianHofmann:
The implementation makes sense and seems like a low hanging fruit. My question is whether this approach makes sense modelling wise. While the total energy limit is preserved with this approach, only late horizons will actually see it. Without  additional smoothing mechanism, I would expect the dispatch to run fully unconstrained in the first horizons until "depletion" and then be zero. When combining with additional penalty cost à la water value this can be avoided. But the latter would already do the job in the first place without the constraint.

fneum:
> I would expect the dispatch to run fully unconstrained in the first horizons until "depletion" and then be zero. 

Yes, that is clear. But the current state will allow the rolling horizon to break any `e_sum_max` limits silently. It makes sense to have the effect represented that the rolling horizon's myopia leads it to deplete resources sub-optimally too early and then having to work around it in later periods.

FabianHofmann:
> Yes, that is clear. But the current state will allow the rolling horizon to break any `e_sum_max` limits silently. It makes sense to have the effect represented that the rolling horizon's myopia leads it to deplete resources sub-optimally too early and then having to work around it in later periods.

but then a user warning at the beginning of the rolling horizon in case of non-inf/non-nan `e_sum_max` values would already do it. don't get me wrong, I wonder about the modelling value of this and whether we should instead guide users to other techniques of fulfilling these kind of global constraints in a rolling horizon in the first place. smoothing the effect with the constraints inplace is hard because as soon as it is binding, the depletion is reached and generation capped. the exercise is then ulitimatively targeting at making the constraint redundant and non-binding again to reach a sensible dispatch pattern. then you could also do it without.

fneum:
The modelling value is high. When running rolling horizon, you would expect the model to run into problems by early depletion of resources and investigate the effects of this. If users want to implement mechanisms for smoothing (e.g. through water values), they can still do it with `e_sum_max` in place (as the water values regulate away from early depletion, while still being limited to a maximum energy use). I would not implement any heuristic mechanism to deal with such global constraints (as we don't know what a "sensible" dispatch pattern is) in rolling horizon but stick to the problem definition (which is a naive upper energy usage limit).

FabianHofmann:
good, I don't want to be a blocker. my point is that this rather a "debugging" constraint for iterative modelling exercise. as long as it is binding, users cannot expect realistic market behavior and solely the order of horizons determines where generation is capped. but I shut up now, let's get this in. let me know if I should help with addressing some of the remainers

MaykThewessen:
Agree with @FabianHofmann that while the cap is binding this is essentially a diagnostic constraint: where generation gets capped is then an artifact of horizon ordering rather than economics, so it isn't a realistic dispatch pattern. The value, as @fneum says, is in surfacing that myopic early-depletion instead of letting it pass silently.

One thing I'd add to that framing: even once a user layers a water value on top to smooth dispatch, keeping `e_sum_max` in place still earns its keep. A water value is a soft incentive, and under myopic foresight a mis-calibrated one can still overshoot the annual limit, with the breach only visible in post-processing. So I'd treat the two as complementary (hard guarantee + soft shaping) rather than the cap becoming redundant. We run rolling-horizon DC-OPF on the Dutch grid and hit exactly this.

Robustness nit on the implementation:

```python
n.c.generators.static["e_sum_max"] = (e_sum_max_orig - generated).clip(lower=0)
```

Solver tolerance can let committed energy creep just above the original budget, pushing the remaining budget slightly negative, and a negative `e_sum_max` makes the next window infeasible (`sum(p·w) ≤ negative` with `p ≥ 0`). Clipping at 0 avoids it.

Happy to add the `e_sum_min` mirror as a follow-up, though that side has the extra wrinkle that a window can't always make up a deficit later, so it probably wants the warning @FabianHofmann suggested rather than silent tracking.

pypsa/optimization/abstract.py
MaykThewessen, line 560: Guard the remaining budget against going negative. Solver feasibility tolerance can let cumulative committed energy creep a hair above `e_sum_max_orig`, which makes the next window's `e_sum_max` slightly negative. A negative `e_sum_max` then makes that window infeasible: the constraint becomes `sum(p·w) ≤ (negative)`, and since `p ≥ 0` even `p = 0` violates it.

```suggestion
                    n.c.generators.static["e_sum_max"] = (e_sum_max_orig - generated).clip(lower=0)
```

PR #1703, opened 2026-05-30
fix: optimize() now honours static `p_set` on Generator/StorageUnit/Store/Link/Process

### LINKED ISSUE #1701: inconsistent use of dynamic / static p_set in lpf and opf

### Issue Description

Historically, n.generators_t.p_set and n.storage_units_t.p_set was only taken into account for lpf, not for optimization. This changed in PR #1154 I think and the docu was adjusted accordingly.
However, one thing is still inconsistent: According to the docu, p_set of generators and of storage units can be static or series and is taken into account for both pf and optimization. BUT actually it is NOT taken into account, if a static p_set is set for the optimization.

See the following example, where a static generator p_set has no effect on optimization outcomes, but the dynamic p_set has. The same is true for the storage unit. See the differences by commenting in lines 9-11.
Summary: The actual behavior is like this:
series p_set: taken into account in lpf AND optimization
static p_set: taken into account in lpf but NOT in optimization

I believe there are two options to handle this:
- adjust docu to make clear that generators p_set is only considered in optimal power flow if set as series
- include the static p_set values in optimization

### Reproducible Example

```python
import pypsa
 
network = pypsa.Network()
 
network.add('Bus', 'B1')
 
network.add('Generator', 'cheap', bus='B1', p_nom=200, marginal_cost=10)
network.add('Generator', 'expensive', bus='B1', p_nom=200, marginal_cost=20, p_set=50) # static p_set has effect on lpf, but not on optimization
# network.add("StorageUnit", "S1", bus="B1", p_nom=100, marginal_cost=20, p_set=50) # static p_set has effect on lpf, but not on optimization
# network.generators_t.p_set.loc['now', 'expensive'] = 50 # dynamic p_set has effect on both lpf and optimization
# network.storage_units_t.p_set.loc['now', 'S1'] = 50 # dynamic p_set has effect on both lpf and optimization
network.add('Load', 'L1', bus='B1', p_set=100)
 
network.lpf()

print("After lpf:")
print(network.generators_t.p)
print(network.storage_units_t.p)
print(network.loads_t.p)
print(network.buses_t.p)
 
network.optimize()
 
print("After optimization:")
print(network.generators_t.p)
print(network.storage_units_t.p)
print(network.loads_t.p)
print(network.buses_t.p)
```

### Expected Behavior

Static p_set values o generators and storage units should be taken into account for opf.

darsh-mishr:
Closes #1701.

`n.optimize()` silently ignored a static (scalar) `p_set` on Generator, StorageUnit, Store, Link, and Process components, while `n.lpf()` honoured it. This contradicts the `static or series` declaration in the component CSVs and the docs. The fix drops a stale `or c.dynamic[attr_set].empty` clause from the early-return guard in `define_fixed_operation_constraints`, so the merged `c.da[attr_set]` accessor on the next line (which already combines static and dynamic) is actually reached.

Regression test added next to the existing dynamic-form tests in `test/test_lopf_basic_constraints.py`.

<details>
<summary>Reproducer and background</summary>

```python
import pypsa

n = pypsa.Network()
n.add("Bus", "B1")
n.add("Generator", "cheap", bus="B1", p_nom=200, marginal_cost=10)
n.add("Generator", "expensive", bus="B1", p_nom=200, marginal_cost=20, p_set=50)
n.add("Load", "L1", bus="B1", p_set=100)
n.optimize()
print(n.c.generators.dynamic.p)
# before: cheap=100, expensive=0   (static p_set=50 ignored)
# after:  cheap=50,  expensive=50
```

The guard was introduced by #1154, which migrated the read from `n.dynamic(c)[attr_set]` to the merged `c.da[attr_set]` accessor but left the early-return still checking only `c.dynamic`, so the new merged read never fired for the static-only case.

Note: this surfaces a separate latent bug in PyPSA-Eur (`cluster_heat_buses` aggregates `p_set` via `sum`, turning NaN into 0.0), tracked in PyPSA/PyPSA-Eur#2182. The PyPSA-Eur (master) CI check stays red until that is fixed.
</details>

## Checklist

- [x] Code changes are sufficiently documented.
- [x] Unit tests for the bug fix were added.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -6,15 +6,20 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Release Notes
 
-## [**v1.2.2**](https://github.com/PyPSA/PyPSA/releases/tag/v1.2.2) <small>25th May 2026</small> { id="v1.2.2" }
+## Upcoming Release
 
-<!--
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
--->
+
+### Bug Fixes
+
+- Fix `n.optimize()` silently ignoring a static `p_set` on [Generator](./user-guide/components/generators.md), [StorageUnit](./user-guide/components/storage-units.md), [Store](./user-guide/components/stores.md), [Link](./user-guide/components/links.md), and [Process](./user-guide/components/processes.md) components. `lpf()` already honoured both static and dynamic `p_set`; the optimizer now matches the documented `static or series` semantics. See [#1701](https://github.com/PyPSA/PyPSA/issues/1701).
+
+
+## [**v1.2.2**](https://github.com/PyPSA/PyPSA/releases/tag/v1.2.2) <small>25th May 2026</small> { id="v1.2.2" }
 
 ### Bug Fixes
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -16,7 +16,7 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ### Bug Fixes
 
-- Fix `n.optimize()` silently ignoring a static `p_set` on [Generator](./user-guide/components/generators.md), [StorageUnit](./user-guide/components/storage-units.md), [Store](./user-guide/components/stores.md), [Link](./user-guide/components/links.md), and [Process](./user-guide/components/processes.md) components. `lpf()` already honoured both static and dynamic `p_set`; the optimizer now matches the documented `static or series` semantics. See [#1701](https://github.com/PyPSA/PyPSA/issues/1701).
+- Fix `n.optimize()` silently ignoring a static `p_set` attribute. (<!-- md:pr 1703 -->)
 
 
 ## [**v1.2.2**](https://github.com/PyPSA/PyPSA/releases/tag/v1.2.2) <small>25th May 2026</small> { id="v1.2.2" }
--- a/pypsa/optimization/constraints.py
+++ b/pypsa/optimization/constraints.py
@@ -1574,7 +1574,8 @@ def define_fixed_operation_constraints(
     c = as_components(n, component)
     attr_set = f"{attr}_set"
 
-    if attr_set not in c.dynamic.keys() or c.dynamic[attr_set].empty:
+    # Those internal guards should be removed and for Lines/ Transformers which are passive, the method should not even be called (clean up needed everywhere)
+    if attr_set not in c.dynamic.keys():
         return
 
     fix = c.da[attr_set].sel(snapshot=sns, name=c.active_assets)
--- a/test/test_lopf_basic_constraints.py
+++ b/test/test_lopf_basic_constraints.py
@@ -326,10 +326,11 @@ def test_define_generator_constraints():
     )
 
 
-@pytest.mark.parametrize("static", [False, True])
-def test_define_fixed_operational_constraints_positive(static):
+@pytest.mark.parametrize("mode", ["dynamic", "static", "mixed"])
+def test_define_fixed_operational_constraints_positive(mode):
     """
-    Test fixed operational constraints: fix to a positive value
+    Test fixed operational constraints: fix to a positive value, set statically,
+    dynamically, or both on the same snapshot
     """
     n = pypsa.Network()
     n.add("Bus", "bus0")
@@ -338,18 +339,26 @@ def test_define_fixed_operational_constraints_positive(static):
     n.add("Generator", "gen1", bus="bus0", p_nom=5, marginal_cost=5)
     n.add("Generator", "gen2", bus="bus0", p_nom=10, marginal_cost=9)
 
-    if static:
-        n.c.generators.static.loc["gen2", "p_set"] = 8
-    else:
+    if mode == "dynamic":
         n.c.generators.dynamic.p_set["gen2"] = 8
+    elif mode == "static":
+        n.c.generators.static.loc["gen2", "p_set"] = 8
+    elif mode == "mixed":
+        n.c.generators.static.loc["gen2", "p_set"] = 8
+        n.c.generators.dynamic.p_set["gen1"] = 1
 
     n.optimize()
 
     assert n.c.generators.dynamic.p["gen2"].eq(8).all()
-    assert n.c.generators.dynamic.p["gen0"].eq(2).all()
+    if mode == "mixed":
+        assert n.c.generators.dynamic.p["gen1"].eq(1).all()
+        assert n.c.generators.dynamic.p["gen0"].eq(1).all()
+    else:
+        assert n.c.generators.dynamic.p["gen0"].eq(2).all()
 
 
-def test_define_fixed_operational_constraints_zero():
+@pytest.mark.parametrize("static", [False, True])
+def test_define_fixed_operational_constraints_zero(static):
     """
     Test fixed operational constraints: fix to a zero value
     """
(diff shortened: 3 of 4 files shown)


darsh-mishr:
CI is fully green on all 22 unit-test matrices (default, `new_api`, `pandas3` on Linux/macOS/Windows × Python 3.11-3.14). The single failing check is **PyPSA-Eur (master)**, and after investigation it turns out to surface a latent PyPSA-Eur bug rather than a problem with this PR.

## Root cause of the PyPSA-Eur failure

PyPSA-Eur's `cluster_heat_buses` in `scripts/prepare_sector_network.py` aggregates the static dataframes of Generator, Link, and Store via:

```python
aggregate_dict = { ..., "p_set": "sum", ... }
df = df.groupby(level=0).agg(aggregate_dict, numeric_only=False)
```

`pd.Series.sum()` with default `skipna=True` returns `0.0` for an all-NaN input, not `NaN`. So every clustered Generator/Link/Store ends up with `p_set = 0.0` in `c.static`, even though no source data ever set a `p_set` on those components. Before this PR, those phantom `0.0`s were silently ignored by `optimize()` (that's the bug #1701 is about). After this PR, they correctly fire `Generator-p_set: p == 0` constraints across every clustered heat-bus component, the LP becomes infeasible at HiGHS presolve, and the integration test fails.

This is precisely the asymmetric divergence `@gugrimm` described in #1701: PyPSA-Eur's `lpf()` honoured the static `p_set = 0.0` from the day clustering was written, `optimize()` silently ignored it, and the downstream workflow drifted into relying on the optimize branch. This PR removes the divergence and surfaces the latent defect.

The PyPSA-Eur side is now filed as [PyPSA/PyPSA-Eur#2182](https://github.com/PyPSA/PyPSA-Eur/issues/2182). The fix is to drop `"p_set"` from `aggregate_dict` (it isn't a real input on Generators/Links/Stores in this code path anyway).

## What I'd like to confirm with you

The PyPSA-side fix here is correct as-is. The integration test will go green once PyPSA-Eur#2182 is fixed. Three reasonable sequencings:

1. Land this PR first, then fix PyPSA-Eur#2182. PyPSA-Eur's sector test stays red until both are in.
2. Wait for PyPSA-Eur#2182 to merge, then this PR's CI goes green naturally.
3. I open the PyPSA-Eur PR (one-line change) so both can land together.

@lkstrp @Irieo no preference from my side. Happy to do option 3 if it helps.

docs/release-notes.md
```diff
@@ -18,6 +18,8 @@ SPDX-License-Identifier: CC-BY-4.0
 
 ### Bug Fixes
 
+- Fix `n.optimize()` silently ignoring a static `p_set` on [Generator](./user-guide/components/generators.md), [StorageUnit](./user-guide/components/storage-units.md), [Store](./user-guide/components/stores.md), [Link](./user-guide/components/links.md), and [Process](./user-guide/components/processes.md) components. `lpf()` already honoured both static and dynamic `p_set`; the optimizer now matches the documented `static or series` semantics. See [#1701](https://github.com/PyPSA/PyPSA/issues/1701).
```
lkstrp: This needs to be moved up and the Upcoming Release admonition should be uncommented, since this is not part of v1.2.2

pypsa/optimization/constraints.py
```diff
@@ -1574,7 +1574,10 @@ def define_fixed_operation_constraints(
     c = as_components(n, component)
     attr_set = f"{attr}_set"
 
-    if attr_set not in c.dynamic.keys() or c.dynamic[attr_set].empty:
+    # Skip only when the attribute is not declared for this component. A static
+    # `p_set` is honoured below via `c.da[attr_set]`, which merges static and
+    # dynamic. See #1701.
```
lkstrp: Fully remove this check, currently an attribute is never not declared for a component.

EDIT: It's actually needed because of the wild nesting on how we call `define_fixed_operation_constraints`

Leave it for now, but remove the comment and instead add:
```
# Those internal guards should be removed and for Lines/ Transformers which are passive, the method should not even be called (clean up needed everywhere)
```

test/test_optimize_static_p_set.py
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: PyPSA Contributors
+#
+# SPDX-License-Identifier: MIT
+"""Regression test for #1701 — static `p_set` honoured by optimize().
+
```
lkstrp: Rewrite this test and integrate it into existing tests, don't just create a new file per test

lkstrp (changes requested):
Thanks for tackling this @darmis007. Please check the review comments.

Also please have a look on https://docs.pypsa.org/latest/contributing/contributing/#examples. `1.` is not followed and this PR description contains a lot of noise

docs/release-notes.md
darsh-mishr: Moved to a new `## Upcoming Release` section above v1.2.2 with the admonition uncommented.

pypsa/optimization/constraints.py
darsh-mishr: Kept the guard, replaced the comment with your suggested wording about the cleanup needed for passive components.

test/test_optimize_static_p_set.py
darsh-mishr: Removed the standalone file. Added `test_define_fixed_operational_constraints_static` to test/test_lopf_basic_constraints.py, next to the existing dynamic-form _positive/_zero/_extendable tests.

darsh-mishr:
All four points addressed: trimmed the description, moved the release note into a new Upcoming Release section, swapped the inline comment, and folded the test into test_lopf_basic_constraints.py. Rebased onto current master. The PyPSA-Eur (master) check will stay red until PyPSA/PyPSA-Eur#2182 is resolved, since this fix is what surfaces that latent bug.

docs/release-notes.md
Irieo: What about version that is more compact and keeps patch notes log easy to follow.
> Fix `n.optimize()` silently ignoring a static `p_set` attribute. (PR reference)

test/test_lopf_basic_constraints.py
```diff
@@ -326,7 +326,8 @@ def test_define_generator_constraints():
     )
 
 
-def test_define_fixed_operational_constraints_positive():
+@pytest.mark.parametrize("static", [False, True])
```
Irieo: If we are already at it, we can deal with missing static p_set test coverage.

- cover StorageUnit that has own p_dispatch - p_store = fix branch in this constraint
- cover static p_set=0  to trim cheap generator/link down.
- one component has a static p_set and another has a dynamic p_set on the same snapshot

Irieo (changes requested):

lkstrp:
I also don't think this is correct:
> The PyPSA-Eur (master) check will stay red until https://github.com/PyPSA/pypsa-eur/issues/2182 is resolved, since this fix is what surfaces that latent bug.
 
I will push a bit more stuff here @darmis007

darsh-mishr:
Thanks both, the parametrized test and the `_sum_keep_na` clustering fix make sense and I've built on top of them.

I dug into why the PyPSA-Eur check is still red after the clustering fix, and I think there's one more spot. `cluster_heat_buses` in PyPSA-Eur keeps its own local `aggregate_dict` with `"p_set": "sum"` (`prepare_sector_network.py:5712`), so it doesn't go through `DEFAULT_ONE_PORT_STRATEGIES` and the `_sum_keep_na` change doesn't reach it. Running just that clustering step plus `optimize()`:

```
'p_set': 'sum'        -> merged heat-pump link p_set = 0.0  -> infeasible
'p_set': min_count=1  -> merged heat-pump link p_set = NaN  -> optimal
```

The two `residential/services heat pump` links merge, `sum` turns their unset `p_set` into `0.0`, and this PR then pins `Link-p == 0` on the only supply to the heat load. So I think the same `min_count=1` treatment is needed there too, which is what PyPSA-Eur#2182 is for. Happy to put that one-liner up mirroring your approach.

On @Irieo's points: I've added the extra coverage (StorageUnit `p_dispatch - p_store` branch, static `p_set=0`, and a static+dynamic mix on the same snapshot) and shortened the release-note bullet. I have these ready locally but haven't pushed, since you're both working in the branch. Just let me know once you're done with your changes and I'll push mine on top so it lands cleanly.

PR #1704, opened 2026-05-31
docs: add line limit examples

### LINKED ISSUE #1481: More forms of `Line` limits
- enforced limits on voltage angle differences (directly translate to `s_nom = 1/x_pu_eff * angle_diff`).

- TATL, Temporary Admissible Transmission Loading, see https://arxiv.org/pdf/2112.06667 (temporarily higher line loading `delta_s_nom`, which has constraint energy over set period of time).

- example of dynamic line rating (`s_max_pu > 0`).

- example of approximation of N-1 (`s_max_pu = 0.7)`

william-xue:
## Summary

- Add a short example of line loading limits in the `Line` component docs.
- Mention static thermal ratings, time-varying `s_max_pu`, and a simple `N-1` margin example.

## Validation

- `git diff --check`
- `python3` content check for the inserted examples

Closes #1481


lkstrp (changes requested):
Thanks @william-xue ! We highly welcome new contributors who are tackling issues!

I think this is not the place for it though, and should instead be a proper additional example, similar to the current structure of the docs. It might should further be discussed over in #1481

PR #1720, opened 2026-06-10
use importlib.resources for package-internal files

fneum:
Took inspiration from #1603


FabianHofmann:
this is nice, I would propose to rename `PYPSA_DIR` to `PYPSA_DATA_DIR` and move it to `constants.py` and use it in components/types.py too. I would push that here, if you don't mind

fneum:
Yes, thanks! Never found the time to finish up. Happy if you can do it.

FabianHofmann (approved):
great, looks good. @lkstrp want to take another look?

docs/release-notes.md
```diff
@@ -6,15 +6,17 @@ SPDX-License-Identifier: CC-BY-4.0
...
 !!! info "Upcoming Release"
 
     The features listed below have not yet been released, but will be included in the
     next update! If you would like to use these features in the meantime, you will need
     to install the `master` branch, e.g. `pip install git+https://github.com/pypsa/pypsa`.
--->
+
+### Features
+
+- Improved package-internal file path resolution to use `importlib.resources.files()`. (<!-- md:pr 1720 -->)
```
lkstrp: With the next release I would probably remove that again since this is not user facing. And a full list of PRs you always have in the release notes. But fine to merge

lkstrp (approved):

docs/release-notes.md
FabianHofmann: done

PR #1721, opened 2026-06-10
fix: resolve statistics plotting issues

### LINKED ISSUE #1719: Statistics plotting accessor: collection of bugs and two enhancement requests
I encountered quite some issues with the statistic plots when preparing the [statistics training](https://github.com/open-energy-transition/pypsa-training-sessions). I used Claude to create MRE and collect them (individual issues would feel like spam)

> [!Note]
> Generated with AI

While running training sessions on the `n.statistics.*.plot.*` / `.iplot.*` accessors we hit a cluster of related issues. They all live in `pypsa/plot/statistics/` and several share a root cause (kwarg routing through the schema), so we're reporting them together rather than as separate issues. Each item below is a self-contained, verified MRE with the observed behaviour and a pointer to the relevant source line.

**Environment** (all MREs run against this):

```
pypsa       1.2.2.post1.dev4+g46aa7ed62
python      3.12.3
plotly      6.5.2
seaborn     0.13.2
matplotlib  3.10.8
pandas      2.3.3
```

All MREs use bundled example networks (`pypsa.examples.*`) and call `n.optimize()` only where results are required (HiGHS, no external solver needed). `carbon_management` is used where the data itself is the trigger.

---

### 1. `color=None` is silently overridden to `"carrier"`

Passing `color=None` to disable color grouping has no effect — the schema replaces `None` with the method default (`"carrier"`).

```python
import warnings; warnings.filterwarnings("ignore")
import pypsa

n.optimize()

# y='bus_carrier' != 'carrier', so the charts.py:652 guard (color==x/y) does not mask it
fig = n.statistics.energy_balance.iplot.bar(y="bus_carrier", color=None)
print([t.name for t in fig.data])
# Observed: ['gas', 'load', 'wind'] — 3 carrier-colored traces
# Expected: a single un-grouped trace
```

**Root cause:** `schema.py:306-307` — `if value is None: kwargs[param] = schema[...]["default"]` conflates "user passed `None`" with "not provided". A sentinel default would let `None` mean "disable".

### 2. `at_port` raises `TypeError` (forwarded to `px.bar`)

`at_port` is a valid statistics parameter (`supply`, `withdrawal`, `energy_balance`, …) but is not routed into the stats call; it leaks into `**kwargs` forwarded to Plotly Express.

```python
import warnings; warnings.filterwarnings("ignore")
import pypsa

n.optimize()

n.statistics.supply.iplot.bar(at_port="all")
# Observed: TypeError: bar() got an unexpected keyword argument 'at_port'
# Expected: supply bar chart aggregated across ports
```

**Root cause:** `plotter.py:776` — unrecognised `**kwargs` go straight to `plotter.iplot()` → `px.bar`, bypassing `stats_kwargs`. Routing kwargs that match the statistics function's signature into `stats_kwargs` would fix this whole class (see also #3, #4).

### 3. `carrier=` is silently dropped for `prices`

Filtering is ignored with no error or warning: the figure still shows all buses.

```python
import warnings; warnings.filterwarnings("ignore")
import pypsa

n.optimize()

buses_all = sum(len(t.x) for t in n.statistics.prices.iplot.bar().data)
buses_ac  = sum(len(t.x) for t in n.statistics.prices.iplot.bar(carrier="AC").data)
print(buses_all, buses_ac)
# Observed: 9 9  (carrier="AC" had no effect)
# Expected: 9 6  (or a clear error/warning that prices takes bus_carrier, not carrier)
```

**Root cause:** `schema.py:293-294` — `EXCLUDED_PARAMS["prices"]` strips `carrier` silently. `prices()` accepts only `bus_carrier`; the accessor should raise/warn rather than discard the filter.

### 4. Distribution plots (`box`/`violin`/`histogram`) collapse to one point per category

The statistics function is called with its scalar time aggregation (e.g. `capacity_factor` → `mean`), so each category becomes a single value and no distribution is drawn. `groupby_time` is reserved and cannot be overridden from the plot API.

```python
import matplotlib; matplotlib.use("Agg")
import pypsa

n.optimize()

cf    = n.statistics.capacity_factor(groupby=["carrier"])
cf_ts = n.statistics.capacity_factor(groupby=["carrier"], groupby_time=False)
print(cf.shape, cf_ts.shape)   # (5,) vs (5, 10) — distributions need the time series

try:
    n.statistics.capacity_factor.plot.violin(x="value", groupby_time=False)
except ValueError as e:
    print("no override:", e)
# Observed: violin/box receive 1 value per category → flat mark; groupby_time can't be passed
# Expected: distribution kinds default to groupby_time=False
```

**Root cause:** `charts.py:866` — the `if "snapshot" in args:` guard never fires for distribution plots (schema forces `x="value"`), so `groupby_time=False` is never set. Suggest defaulting `groupby_time=False` for `box`/`violin`/`histogram` in `derive_statistic_parameters`.

### 5. `prices.plot.box` height does not scale with bus count

`height` is fixed at `4` regardless of bus count, so on large networks the boxes are illegible, and there's no `groupby` escape hatch for prices.

```python
import matplotlib; matplotlib.use("Agg")
import pypsa

n.optimize()
fig, ax, g = n.statistics.prices.plot.box()
print(fig.get_figheight())   # 4.0 for 9 buses; identical for 200 buses
# Observed: fixed 4-inch height (~0.08 in/bus at 50 buses); no per-bus scaling, no aggregation
# Expected: height auto-scales with row count, or a bus_carrier/region grouper is allowed
```

**Root cause:** `schema.py:39` (`DEFAULTS["height"] = 4`, no auto-scaling in `charts.py:458-477` when `y="name"`) and `charts.py:858-859` (always pops `groupby` for prices).

### 6. `.plot.line` / `.plot.scatter` ignore carrier colors

The static line/scatter paths don't pass `palette` to seaborn, so `color="carrier"` falls back to the default color cycle (the bar/area paths apply it correctly).

```python
import matplotlib; matplotlib.use("Agg")
import matplotlib.colors as mcolors
import pypsa

fig, ax, g = n.statistics.installed_capacity.plot.line(carrier=["gas", "wind"], color="carrier")
actual = {l.get_label(): mcolors.to_hex(l.get_color()) for l in ax.lines if not l.get_label().startswith("_")}
print(actual)
# Observed: {'gas': '#1f77b4', 'wind': '#ff7f0e'}  — matplotlib defaults
# Expected: {'gas': '#ff0000', 'wind': '#0000ff'}  — n.c.carriers.static.color
```

**Root cause:** `charts.py:512-514` — `sns.lineplot`/`sns.scatterplot` (and `sns.histplot`) are called without `palette=palette`. Fix: pass `palette` like the bar/area path does.

### 7. LaTeX `$_2$` in carrier `nice_name` breaks Plotly 6 legends

`carbon_management` ships carrier `nice_name`s with mathtext (e.g. `H$_2$ Storage`). Plotly 6 auto-injects MathJax, which renders only the subscript glyph and drops the surrounding text, so legends read `"2"`, `"."` etc.

```python
import pypsa
n = pypsa.examples.carbon_management()   # slow

latex = n.c.carriers.static.nice_name[n.c.carriers.static.nice_name.str.contains(r"\$")]
print(len(latex), "affected; e.g.:", latex.iloc[0])   # 12 affected; e.g.: H$_2$ Storage
import plotly; print("plotly", plotly.__version__)     # 6.5.2

n.statistics.installed_capacity.iplot.bar(bus_carrier="AC")                    # legend shows "2"…
n.statistics.installed_capacity.iplot.bar(bus_carrier="AC", nice_names=False)  # workaround: clean
```

**Root cause:** primarily upstream data (LaTeX baked into the example `.nc`), but PyPSA could sanitize `$_X$` → Unicode subscript in the label path (`base.py:98-101`). Workarounds today: `nice_names=False`, or `n.carriers.nice_name.str.replace(r"\$_2\$", "₂", regex=True)`.

### 8. `facet_col` silently drops empty facets

Requesting facets for bus carriers that have no data yields fewer panels, with no warning that the request was partially dropped.

```python
import warnings
import pypsa
n = pypsa.examples.carbon_management()   # slow

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    fig = n.statistics.installed_capacity.iplot.bar(
        bus_carrier=["AC", "H2", "heat", "co2 stored"], facet_col="bus_carrier",
    )
print([str(x.message) for x in w])                          # []
print([a.text for a in fig.layout.annotations if a.text])   # ['bus_carrier=AC']
# Observed: 1 panel, no warning (H2/heat/co2 stored are empty: multi-port links report at bus0, Stores need storage=True)
# Expected: a warning, or empty-but-visible panels
```

**Root cause:** `charts.py:570` — requested `bus_carrier` values are never forwarded as `col_order`, so Plotly only renders facets with data. (`col_order=[...]` is a manual workaround.)

### 9. `sharex=False` + `facet_col_wrap` hides x-axis tick labels on upper rows

Unsharing the axes leaves the upper-row subplots without x tick labels.

```python
import pypsa
n = pypsa.examples.carbon_management()

fig = n.statistics.installed_capacity.iplot.bar(
    bus_carrier=["AC", "Biomass", "Gas"], facet_col="bus_carrier",
    facet_col_wrap=2, sharex=False,
)
for k in sorted(x for x in fig.layout.to_plotly_json() if x.startswith("xaxis")):
    print(k, fig.layout.to_plotly_json()[k].get("showticklabels", "inherit"))
# Observed: xaxis3/xaxis4 (upper row) showticklabels=False
# Expected: all subplots show tick labels when sharex=False
```

**Root cause:** `charts.py:822-823` — `fig.update_xaxes(matches=None)` unlinks scales but doesn't restore `showticklabels`. Fix: `fig.update_xaxes(matches=None, showticklabels=True)` (and symmetrically for `sharey`).

---

### 10. Control carrier order (stacking / legend / axis)

There is currently no working way to control carrier order on the static path. `hue_order` is accepted but ignored, and `iplot`'s `color_order` doesn't reorder the categorical axis for single-trace bars.

```python
import matplotlib; matplotlib.use("Agg")
import pypsa

fig, ax, g = n.statistics.installed_capacity.plot.bar(color="carrier", hue_order=["DC", "AC", "wind", "gas"])
print([t.get_text() for t in ax.get_yticklabels()])
# Observed: ['AC', 'DC', 'gas', 'wind'] — hue_order ignored
# Desired:  ['DC', 'AC', 'wind', 'gas'] — and ideally a default order from n.carriers
```

**Root cause / direction:** `charts.py:312` — `g.hue_names` is always `None` because `FacetGrid` (line 458) is built without `hue=color`, so the reorder block (336-340) is dead. A clean fix would build the grid with `hue=` and default carrier order to `n.c.carriers.static.index` (filtered to present carriers). Relates to #6.

### 11. Expose `col_wrap` (a.k.a. `ncol`) for static faceted plots

`FacetGrid` supports `col_wrap`, but passing it crashes instead of wrapping the facets.

```python
import matplotlib; matplotlib.use("Agg")
import pypsa

n.statistics.installed_capacity.plot.bar(facet_col="bus_carrier", col_wrap=2)
# Observed: AttributeError: Rectangle.set() got an unexpected keyword argument 'col_wrap'
# Desired:  facets wrapped into 2 columns
```

**Root cause:** `charts.py:458` — `col_wrap` isn't extracted before `**kwargs` is forwarded, so it reaches matplotlib's `Rectangle.set()`. Also note `facet_iter` (`charts.py:238-241`) indexes a 2-D axes grid and would need to handle the 1-D layout `col_wrap` produces.

FabianHofmann:
> [!NOTE]
> The description below was drafted by AI and reviewed by me.

Closes #1719.

Resolves a collection of bugs and enhancements in the `n.statistics.*.plot` / `.iplot` accessors:

- `color=None` now disables color grouping instead of falling back to `"carrier"`.
- Statistics-only arguments (e.g. `at_port`) are routed to the statistics call instead of leaking to the plotting backend.
- Passing an unsupported filter (e.g. `carrier` to `prices`) raises a clear error instead of being silently dropped.
- Distribution plots (`box`/`violin`/`histogram`) default to the full time series so they show real distributions.
- Box/violin figure height scales per facet with the number of categorical rows (capped at 30 inches).
- Static `line`/`scatter` plots honor carrier colors via the palette.
- LaTeX sub-/superscripts in carrier nice names (e.g. `H$_2$`) are converted to Unicode so Plotly legends render correctly.
- Requested but empty facets stay visible instead of being silently dropped.
- `sharex=False` / `sharey=False` keeps tick labels on all faceted subplots.
- Interactive plots with a categorical y-axis label every bar and scale the figure height with the number of rows (capped at 2000 pixels), instead of thinning tick labels.
- Static faceted plots accept `col_wrap` to wrap facets across rows.
- `n.statistics.prices` accepts any static bus attribute (e.g. `country`) in `groupby`, and its plot accessors support these as dimensions (e.g. `prices.plot.box(facet_row="country")`) instead of failing with a `KeyError`.

## Checklist


pypsa/plot/statistics/base.py
```diff
@@ -17,6 +18,39 @@
...
+
+    def __repr__(self) -> str:
+        return "UNSET"
+
+
+UNSET: Any = _Unset()
+
+
+_SUBSCRIPT = str.maketrans("0123456789+-=()aeoxhklmnpst", "₀₁₂₃₄₅₆₇₈₉₊₋₌₍₎ₐₑₒₓₕₖₗₘₙₚₛₜ")
+_SUPERSCRIPT = str.maketrans("0123456789+-=()ni", "⁰¹²³⁴⁵⁶⁷⁸⁹⁺⁻⁼⁽⁾ⁿⁱ")
```
fneum: Where does this collection of letters come from? Is it general enough?

pypsa/plot/statistics/charts.py
```diff
@@ -606,6 +645,12 @@ def iplot(
             ldata = ldata.query(query)
         ldata = self._validate(ldata)
 
+        # Replace LaTeX mathtext in labels with Unicode so Plotly legends render
+        # the full text (Plotly 6 otherwise drops text around MathJax glyphs).
+        ldata = ldata.apply(
+            lambda col: col.map(sanitize_mathtext) if col.dtype == "object" else col
+        )
```
fneum: Does it make sense to open an issue in Plotly? Sounds like a bug and would be better if we didn't have to handle that here.

fneum (changes requested):
In the future, let's split such collections into multiple PRs.

Some before / after comparisons of code and resulting plots would be helpful to understand the changes.

FabianHofmann:
> In the future, let's split such collections into multiple PRs.

agreed with @lkstrp do batch the changes (would have been 10 prs or so). The code is mostly complex due to the many combinations of parameters and function dependencies. it really best to test it out in a notebook as in the one attached
[statistics-plotting.ipynb](https://github.com/user-attachments/files/29249215/statistics-plotting.ipynb)

fneum:
Fair, didn't know that was agreed. I'm happy when there are some guides to quickly test and understand the changes (like the notebook you attached).

pypsa/plot/statistics/charts.py
```diff
@@ -34,6 +45,20 @@
...
+SEABORN_PLOT_FUNCS = {
+    "scatter": sns.scatterplot,
+    "line": sns.lineplot,
+    "box": sns.boxplot,
+    "violin": sns.violinplot,
+    "histogram": sns.histplot,
+}
+
+
+def _stats_function_params(network: Any, method_name: str) -> set[str]:
```
lkstrp, line 53: If we go with `inspect` here, which we decided against in the beginning. But I am not opposed against when I look on this chaos here and the downfalls for plotting calls are also not too large. But I think we need to touch our schema then as well. For sure `ALLOWED_PARAMS` is now redundant  but I guess there is more possible clean up

pypsa/plot/statistics/plotter.py
```diff
@@ -235,7 +260,7 @@ def chart(  # noqa: D417
         chart_type: str,
         x: str | None = None,
         y: str | None = None,
-        color: str | None = None,
+        color: str | None = UNSET,
```
lkstrp, line 263: `UNSET` is now only used for colors. I would expect it to be used consistent. So wherever None just leads to a derived default, `UNSET` should be used instead. `None` then means the user explicitly passed it.

lkstrp (changes requested):
Lovely list of squashed bugs! Some comments and I made some edits in #1749

> Some before / after comparisons of code and resulting plots would be helpful to understand the changes.

We had tests on top of the actual plot figure at some point. Maybe we should bring them back. They haven't been very stable, which was annoying, but there is definitely a solution for this

> Fair, didn't know that was agreed. I'm happy when there are some guides to quickly test and understand the changes (like the notebook you attached).

That is generally a useful approach, which @FabianHofmann started do to, just a little notebook to explain what is going on for the reviewers, which can then be removed again before merging. So I am also happy with just dumping them in the PR. They just need to be removed before the merge. In attachments, it is always a bit fiddly

PR #1725, opened 2026-06-12
fix: address review findings for piecewise constraints

FabianHofmann:
Iterating with Claude on bugs (may still include false positives). All combined here

> [!Note]
> AI generated summary and code

Targets #1603. Addresses the must-fix findings from a structured review of the branch diff, plus three follow-up design decisions.

## Crashes fixed
- **≥2 `primary_energy` constraints + piecewise efficiency**: `define_piecewise` was called per constraint/scenario with a fixed aux name → duplicate-variable error. The primary-energy relation is now defined once and each constraint selects from it.
- **Piecewise port coefficients + delays**: duplicate aux var per delay group, and the delayed branch passed a `LinearExpression` as `x_var` (`_term` dim crash). Nodal balance now defines each port's piecewise relation once on the *undelayed* dispatch variable and time-shifts the aux variable for delayed groups — consistent with what `assign_solution` already assumed.
- Latent `NameError` in `_Multiport._piecewise_attrs` (`extra_rows` rebound/unbound).

## Silent wrong results fixed
- **StorageUnit piecewise `marginal_cost`** was accepted but silently dropped from the objective (looked up nonexistent `StorageUnit-p`); now wired to `p_dispatch` end-to-end.
- **Store piecewise `marginal_cost`** applied to the storage level `e`; now applies to dispatch `p`, consistent with linear `marginal_cost` (x per unit `e_nom`).
- **Spurious `p{i}_piecewise_opt` outputs** (efficiency ratios mislabeled as power) for Link ports ≥2 and all Process ports — `assign_solution` skip condition now uses `_get_base_coeff`.
- **`remove()` left stale piecewise curves**; a re-added component silently inherited the old curve.
- **`overnight_cost` + piecewise `capital_cost`** silently skipped annuitization → now a `ValueError` (the curve must already be periodized).
- **`TODO-1603` supply/withdrawal sign resolved**: piecewise ports are classified like linear ones by curve sign — positive → supply, negative → withdrawal (reported positive), mixed-sign → `NotImplementedError`.

## Fail fast instead of crash
- **Scenarios + piecewise** crashed in both orders (add-then-`set_scenarios` and reverse); both now raise `NotImplementedError` until stochastic support lands.
- **Dead/inverted MultiIndex column validation** in `n.add` (`Index.equals(list)` is always `False`) fixed to a set comparison.

## API
- `piecewise_options` is no longer a required positional parameter of the public `define_nodal_balance_constraints` (was inserted mid-signature), `define_objective`, `define_primary_energy_limit` — now `| None = None` appended at the end (backward compatible).
- `PiecewiseOptions.name` annotation matches accepted inputs (`str | Iterable[str] | None`).
- Removed module-level `warnings.filterwarnings` mutating global state on `import pypsa` (scoped suppression retained); removed write-only `piecewise_y_vars`; fixed stale `_import_piecewise_from_df` docstring and the broken docs CSV include path.

## Tests
- The four piecewise IO round-trip tests asserted nothing: `Components.equals` now compares `piecewise` and the tests assert the result.
- `test_piecewise_efficiency_gen` asserted an arbitrary vertex of a cost-degenerate optimum; now asserts load balance, objective, and CO2 feasibility.
- Multi-invest + piecewise capex verified consistent with the linear path and pinned by test (constant-slope curve ≡ linear equivalent).
- New regression tests: two primary-energy constraints, piecewise+delay with mixed groups, no leaked `p{i}_piecewise_opt`, StorageUnit/Store dispatch pricing, supply/withdrawal for positive and negative curves, remove-then-re-add, scenario fail-fasts, malformed MultiIndex input, overnight-cost conflict.

## Deferred (nice-to-haves, not in this PR)
Duplicated capex/opex aux-var lookup block; byte-identical Link/Process branches in `port_efficiency`; dead committable `cost_piecewise_opt` term; dead Excel sheet-name mappings (long `*_piecewise_opt` names still unmapped); sign-only statistics assertions; linopy-repr string assertions; `ComponentsData.piecewise` required field; `_Importer.get_piecewise` not declared on the ABC; unvalidated `sign`/`method` at the dict boundary (docs example uses `"sign": "=="`).

## Checklist
- [x] Unit tests for new features were added.
- [ ] Release notes — covered by #1603 itself.

### DIFF
--- a/pypsa/network/transform.py
+++ b/pypsa/network/transform.py
@@ -257,11 +257,6 @@ class NetworkTransformMixin(_NetworkABC):
             return s + suffix
 
         for k, v in kwargs.items():
-            # If index/ columns are passed (pd.DataFrame or pd.Series)
-            # - cast names index to string and add suffix
-            # - check if passed index/ columns align
-            msg = "{} has an index which does not align with the passed {}."
-
             # Enable default lookup for attributes with port suffixes, e.g. efficiency1, rate2.
             search_attr = c._get_base_coeff(k) if isinstance(c, _Multiport) else k
 
@@ -280,17 +275,19 @@ class NetworkTransformMixin(_NetworkABC):
                 if isinstance(v, dict):
                     v = pd.DataFrame({x_attr: list(v.keys()), k: list(v.values())})
                 if isinstance(v, pd.DataFrame):
+                    pw_msg = "Piecewise {k} Dataframe has {lvl} column level which does not align with the passed {lvl}s: {diff}."
                     if isinstance(v.columns, pd.MultiIndex):
                         v = v.rename(columns=add_suffix, level=0)
-                        if not v.columns.unique(0).equals(names):
-                            raise ValueError(msg.format(f"Dataframe {k}", names_str))
-                        if set(v.columns.unique(1)) != {x_attr, k}:
-                            msg_seg = (
-                                f"Piecewise Dataframe {k} must have column labels "
-                                f"consisting of {x_attr} and {k}"
+                        lvls = v.columns.levels
+                        if not (diff := lvls[0].difference(names)).empty:
+                            raise ValueError(pw_msg.format(k=k, lvl="name", diff=diff))
+                        if not (
+                            diff := lvls[1].symmetric_difference([x_attr, k])
+                        ).empty:
+                            raise ValueError(
+                                pw_msg.format(k=k, lvl="attribute", diff=diff)
                             )
-                            raise ValueError(msg_seg)
-                        piecewise[k] = v
+                        piecewise[k] = v.reindex(columns=[x_attr, k], level=1)
                         continue
                     elif set(v.columns) == {x_attr, k}:
                         # Build a MultiIndex-columned DataFrame: broadcast the curve to all names
@@ -310,6 +307,10 @@ class NetworkTransformMixin(_NetworkABC):
                 )
                 raise NotImplementedError(msg)
 
+            # If index/ columns are passed (pd.DataFrame or pd.Series)
+            # - cast names index to string and add suffix
+            # - check if passed index/ columns align
+            msg = "{} has an index which does not align with the passed {}."
             if isinstance(v, pd.Series) and single_component:
                 if not v.index.equals(self.snapshots):
                     raise ValueError(msg.format(f"Series {k}", "network snapshots"))
--- a/pypsa/network/transform.py
+++ b/pypsa/network/transform.py
@@ -284,7 +284,7 @@ class NetworkTransformMixin(_NetworkABC):
                         v = v.rename(columns=add_suffix, level=0)
                         if not v.columns.unique(0).equals(names):
                             raise ValueError(msg.format(f"Dataframe {k}", names_str))
-                        if v.columns.unique(1).equals([x_attr, k]):
+                        if set(v.columns.unique(1)) != {x_attr, k}:
                             msg_seg = (
                                 f"Piecewise Dataframe {k} must have column labels "
                                 f"consisting of {x_attr} and {k}"
@@ -393,6 +393,12 @@ class NetworkTransformMixin(_NetworkABC):
         # Broadcast to scenarios if network has scenario structure
         # Skip SubNetwork components, they are handled internally in determine_topology
         if self.has_scenarios and c.name != "SubNetwork":
+            if piecewise:
+                msg = (
+                    "Piecewise attribute data is not yet supported for stochastic "
+                    "networks (scenarios)."
+                )
+                raise NotImplementedError(msg)
             static_df = pd.concat(
                 dict.fromkeys(self.scenarios, static_df), names=["scenario"]
             )
@@ -401,12 +407,6 @@ class NetworkTransformMixin(_NetworkABC):
                     dict.fromkeys(self.scenarios, v), names=["scenario"], axis=1
                 )
                 for k, v in series.items()
-            }
-            piecewise = {
-                k: pd.concat(
-                    dict.fromkeys(self.scenarios, v), names=["scenario"], axis=1
-                )
-                for k, v in piecewise.items()
             }
 
         # Load piecewise breakpoint data
@@ -500,6 +500,11 @@ class NetworkTransformMixin(_NetworkABC):
         # Drop from time-varying components
         for df in c.dynamic.values():
             df.drop(df.columns.intersection(names), axis=1, inplace=True)
+
+        # Drop piecewise breakpoint data
+        for df in c.piecewise.values():
+            stale = df.columns[df.columns.get_level_values("name").isin(names)]
+            df.drop(columns=stale, inplace=True)
 
     def merge(
         self,
--- a/pypsa/optimization/constraints.py
+++ b/pypsa/optimization/constraints.py
@@ -927,10 +927,10 @@ def _get_delay_config(
     return config
 
 
-def _iter_balance_args(
+def _iter_balance_coeffs(
     c: _Multiport, sns: Sequence
-) -> Iterator[tuple[str, Any, pd.Index, int, bool]]:
-    """Iterate over all balance arguments, separating immediate and delayed.
+) -> Iterator[tuple[str, Any, pd.Index, str]]:
+    """Iterate over all balance arguments to get coefficient values per port.
 
     Parameters
     ----------
@@ -947,43 +947,65 @@ def _iter_balance_args(
         Coefficient values for the component
     names : pd.Index
         Component names with this delay configuration
-    delay : int
-        Delay in time units (0 for immediate, >0 for delayed)
-    is_cyclic : bool
-        Whether delay wraps around the horizon
 
     """
     if c.empty:
         return
 
     active = c.active_assets
-    delay_config = _get_delay_config(c)
 
     for port in c._output_ports:
         suffix = c._port_suffix(port)
         coeff = (
             c.da[f"{c._coefficient_attr}{suffix}"].where(c.da.active).sel(snapshot=sns)
         )
-        delays, cyclics = delay_config[suffix]
+        if not active.empty:
+            yield (f"bus{port}", coeff.sel(name=active), active, suffix)
 
-        for (d, cyc), group in c.static.assign(_delay=delays, _cyclic=cyclics).groupby(
-            ["_delay", "_cyclic"]
-        ):
-            delay_int = int(d)
 
-            names = group.index
-            if isinstance(names, pd.MultiIndex):
-                names = names.get_level_values("name").unique()
-            names = names.intersection(active)
+def _iter_balance_delay(
+    c: _Multiport, active: pd.Index, suffix: str
+) -> Iterator[tuple[pd.Index, int, bool]]:
+    """Iterate over all balance arguments, separating immediate and delayed.
 
-            if not names.empty:
-                yield (
-                    f"bus{port}",
-                    coeff.sel(name=names),
-                    names,
-                    delay_int,
-                    bool(cyc),
-                )
+    Parameters
+    ----------
+    c : _Multiport
+        _Multiport component (Link or Process).
+    active : pd.Index
+        Active component names
+    suffix : str
+        Port suffix (e.g., "", "1", "2") for which to get delay configuration
+
+    Yields
+    ------
+    names : pd.Index
+        Component names with this delay configuration
+    delay : int
+        Delay in time units (0 for immediate, >0 for delayed)
+    is_cyclic : bool
+        Whether delay wraps around the horizon
+
+    """
+    delay_config = _get_delay_config(c)
+    delays, cyclics = delay_config[suffix]
+
+    for (d, cyc), group in c.static.assign(_delay=delays, _cyclic=cyclics).groupby(
+        ["_delay", "_cyclic"]
+    ):
+        delay_int = int(d)
+
+        names = group.index
+        if isinstance(names, pd.MultiIndex):
+            names = names.get_level_values("name").unique()
+        names = names.intersection(active)
+
+        if not names.empty:
+            yield (
+                names,
+                delay_int,
+                bool(cyc),
+            )
 
 
 def _apply_delay_shift(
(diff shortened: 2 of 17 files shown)


pypsa/network/transform.py
```diff
@@ -284,7 +284,7 @@ def add_suffix(s: Any) -> str:
                         v = v.rename(columns=add_suffix, level=0)
                         if not v.columns.unique(0).equals(names):
                             raise ValueError(msg.format(f"Dataframe {k}", names_str))
-                        if v.columns.unique(1).equals([x_attr, k]):
+                        if set(v.columns.unique(1)) != {x_attr, k}:
```
brynpickering: Reject: my approach compares elements _and_ order. We care about both.

pypsa/optimization/constraints.py
```diff
@@ -1042,6 +1042,8 @@ def define_nodal_balance_constraints(
 
     """
     m = n.model
+    if piecewise_options is None:
+        piecewise_options = []
     if buses is None:
         buses = n.c.buses.static.index.unique("name")
```
brynpickering, line 1120: Reject: we're always passing this list from the calling function so it makes sense to have it be required, not optional.

pypsa/optimization/constraints.py
```diff
@@ -1139,29 +1139,46 @@ def define_nodal_balance_constraints(
...
+                if not pw_names.empty:
+                    p_piecewise = piecewise_var.sel(name=pw_names)
+                    if delay > 0:
+                        p_piecewise = (
+                            p_piecewise.isel(snapshot=src_snapshot_pos).assign_coords(
+                                sns_coords
+                            )
+                            * valid_mask
+                        )
+                    names = names.difference(pw_names)
```
brynpickering: Reject: unless there's something I'm missing here, this is adding unnecessary bloat. The x variable has already accounted for delay in the preceding lines (`var = shifted_p * valid_mask`); the process of handling delay isn't needed again.

pypsa/optimization/expressions.py
```diff
@@ -563,11 +563,40 @@ def func(n: Network, c: str, port: str) -> pd.Series:
...
+                    is_pos = ((curve >= 0) | curve.isna()).all()
+                    is_neg = ((curve <= 0) | curve.isna()).all()
+                    if not (is_pos | is_neg).all():
+                        bad = curve.columns[~(is_pos | is_neg)].tolist()
+                        msg = (
+                            f"Piecewise curves of {c} components {bad} at port "
+                            f"{port} mix positive and negative values; cannot "
+                            "split into supply and withdrawal."
+                        )
+                        raise NotImplementedError(msg)
```
brynpickering: It's probably something worth checking well in advance of the expressions (i.e., a consistency check on adding a component that doesn't allow efficiency to be a mix of pos & neg values)

pypsa/optimization/expressions.py
```diff
@@ -563,11 +563,40 @@ def func(n: Network, c: str, port: str) -> pd.Series:
             elif direction != "both":
                 msg = f"Got unexpected argument direction={direction}. Must be 'supply', 'withdrawal' or 'both'."
                 raise ValueError(msg)
-            # TODO-1603: work out sign of pw_var depending on direction
             piecewise_efficiency = port_efficiency(n, c, port=port, piecewise=True)
             if isinstance(piecewise_efficiency, pd.DataFrame):
                 pw_var = n.model.variables[f"{c}-p{port}_piecewise"]
                 coeffs.loc[{"name": pw_var.coords["name"]}] = 0
+                if direction != "both":
```
brynpickering: I'm unclear if this is actually doing what's intended. It doesn't account for `sign` in the same way as the static / dynamic coeffs are defined above (should it?). Also, this is applying the same logic as the static/dynamic case for assigning a sign to the coeff based on the idea that it will be multiplied by the `p` variable. However, the piecewise variables are already the resolved port `p` values (`p0`, `p1, `p2`, etc.) so I would expect them to already embed the sign of the efficiency in their values. My `todo` was about trying to untangle this. I _think_ all that is needed is that if we request `supply` or `withdrawal` then `p` of ports >=1 are made positive / negative respectively (and vice-versa for `p0` on processes). It doesn't need all this extra content.

pypsa/optimization/global_constraints.py
```diff
@@ -285,11 +285,13 @@ def define_primary_energy_limit(
...
     sns : list-like
         Set of snapshots to which the constraint should be applied.
-    piecewise_options : list[PiecewiseOptions]
+    piecewise_options : list[PiecewiseOptions], optional
         Options to override defaults in piecewise constraint formulation.
         List is of the form ``[PiecewiseOptions(...), ...]``.
 
     """
+    if piecewise_options is None:
+        piecewise_options = []
```
brynpickering: Reject: as above, don't make something optional that we are always explicitly passing from the parent function calling this funciton.

pypsa/optimization/global_constraints.py
```diff
@@ -350,36 +377,20 @@ def define_primary_energy_limit(
...
-                        linear_names = linear_names.difference(
-                            piecewise_var.indexes["name"]
+                    if not pw_names.empty:
+                        pw_var = primary_energy_pw_var.sel(
+                            name=pw_names, snapshot=sns[sns_sel]
                         )
+                        if n.has_scenarios:
+                            pw_var = pw_var.sel(scenario=scenario, drop=True)
+                        to_sum.append(pw_var)
+                        linear_names = linear_names.difference(pw_names)
```
brynpickering, line 386: I would opt for rejecting this.

It doesn't correctly slice on snapshots of periods by applying the piecewise constraint outside the nested loop and conditional. I'm not such a fan of the nesting either, but it would require a bigger refactoring to extract things. As it is, the piecewise implementation is safer inside the nesting.

pypsa/optimization/optimize.py
```diff
@@ -184,6 +184,8 @@ def define_objective(
     - For a stochastic problem, scenario probabilities are applied as weightings to all cost (includes *both* investment terms).
 
     """
+    if piecewise_options is None:
+        piecewise_options = []
```
brynpickering: See previous comments on this. If the method is exposed publicly, I can understand wanting to leave some things as optionally empty. If they're not, it creates ambiguity in the function signatures that makes type hinting and general maintenance more difficult.

pypsa/optimization/optimize.py
```diff
@@ -385,7 +387,17 @@ def define_objective(
...
+                overnight = c.static["overnight_cost"].loc[pw_names]
+                if overnight.notna().any():
+                    bad = overnight[overnight.notna()].index.tolist()
+                    msg = (
+                        f"Components {bad} of type {c.name} define both a piecewise "
+                        "'capital_cost' curve and 'overnight_cost'. The piecewise "
+                        "curve must already be periodized; remove 'overnight_cost'."
+                    )
+                    raise ValueError(msg)
+                ext_i = ext_i.difference(pw_names)
```
brynpickering, line 398: probably fine to have in but we should have a todo to convert piecewise overnight cost to a comparable capital cost (as we do with the static attr) in future.

pypsa/optimization/piecewise.py
```diff
@@ -37,7 +35,7 @@ class PiecewiseOptions:
     """Component attribute for which piecewise data is defined."""
     sign: SIGNS_T
     """Sign for the piecewise constraint, interpreted as y <sign> f(x)."""
-    name: tuple[str, ...] = field(default_factory=tuple)
+    name: str | Iterable[str] | None = field(default_factory=tuple)
```
brynpickering: Reject: the post_init function will force this to a tuple and so there's no need to worry about handling the different possible versions of `name` in the main method.

pypsa/optimization/piecewise.py
```diff
@@ -138,7 +136,9 @@ def define_piecewise(
     if y_var is None:
         y_var = _create_y_var(m, x_var, pw_names, aux_var_name)
     y_var_sel = y_var.sel(name=pw_names)
-    sorted_options = sorted(extra_options, key=lambda x: x.name, reverse=True)
+    sorted_options = sorted(
+        extra_options, key=lambda x: tuple(x.name or ()), reverse=True
+    )
```
brynpickering: Reject, see above.

pypsa/optimization/piecewise.py
```diff
@@ -148,10 +148,7 @@ def define_piecewise(
                 sign,
             )
         elif option.name:
-            names = pd.Index(
-                [option.name] if isinstance(option.name, str) else option.name,
-                name="name",
-            ).intersection(pw_names)
+            names = pd.Index(option.name, name="name").intersection(pw_names)
```
brynpickering, line 149: Interestingly, this is a correct change assuming option.name has always been transformed into a tuple. So it only works if my above rejections are applied.

test/test_lopf_piecewise_costs.py
```diff
@@ -231,3 +231,98 @@ def test_piecewise_capital_cost(self) -> None:
...
+    n.add("Bus", "bus0")
+    n.add("Generator", "gen0", bus="bus0", p_nom=100, marginal_cost=50)
+    n.add(
+        "StorageUnit",
+        "su0",
+        bus="bus0",
+        p_nom=100,
+        max_hours=1,
+        state_of_charge_initial=100,
+        marginal_cost={0.0: 10.0, 1.0: 10.0},
```
brynpickering: This is effectively the same as setting a single value for marginal cost (since the zero value will be ignored, so you'll just get marginal_cost=10). Would be better to set a different value at some intermediate point and set its state_of_charge_initial at some part-load value. Then you can check that the marginal cost matches the point along the part-load curve.

test/test_lopf_piecewise_costs.py
```diff
@@ -231,3 +231,98 @@ def test_piecewise_capital_cost(self) -> None:
         assert n.c.generators.static.capital_cost_piecewise_opt.equals(
             pd.Series({"gen0": 0.0, "gen1": 1.4})
         )
+
+
+def test_storage_unit_piecewise_marginal_cost() -> None:
```
brynpickering: Move these tests into the `TestPiecewiseCostsResults` class

test/test_lopf_piecewise_costs.py
```diff
@@ -231,3 +231,98 @@ def test_piecewise_capital_cost(self) -> None:
...
+    n.set_snapshots(range(2))
+    n.add("Bus", "bus0")
+    n.add("Generator", "gen0", bus="bus0", p_nom=100, marginal_cost=50)
+    n.add(
+        "Store",
+        "store0",
+        bus="bus0",
+        e_nom=100,
+        e_initial=100,
+        marginal_cost={0.0: 10.0, 1.0: 10.0},
```
brynpickering: As above

test/test_lopf_piecewise_costs.py
```diff
@@ -231,3 +231,98 @@ def test_piecewise_capital_cost(self) -> None:
...
+
+    n_pw = build(piecewise=True)
+    n_lin = build(piecewise=False)
+    assert n_pw.objective == pytest.approx(n_lin.objective)
+    assert n_pw.c.generators.static.p_nom_opt.item() == pytest.approx(
+        n_lin.c.generators.static.p_nom_opt.item()
+    )
+
+
+def test_piecewise_capital_cost_with_overnight_cost_raises() -> None:
```
brynpickering: Move this into a `TestPiecewiseCostsExceptions` class

test/test_lopf_piecewise_efficiency.py
```diff
@@ -74,8 +75,25 @@ def test_piecewise_efficiency_co2_constraint_with_only_segmented_gens(
...
+        constant=170,
+    )
+    status, _ = n.optimize()
+    assert status == "ok"
+    assert n.generators_t.p["gen"].iloc[0] == pytest.approx(50.0, rel=1e-3)
+
+
 def test_piecewise_efficiency_gen() -> None:
-    """We get a somewhat arbitrary solution as the optimisation problem aims to hit the CO2 limit exactly."""
+    """The dispatch split is cost-degenerate, so only assert properties unique to the optimum."""
```
brynpickering: Lots of jargon has been put in here. I don't think it will make sense to most maintainers in future.

What is weird about the result is that the objective doesn't care about the generator efficiency (because the split is "cost-degenerate", or because the objective isn't influenced by efficiency or the emissions target at all). The least emitting result would be `gen0` @10MW and `gen1` @70MW. However, the emissions constraint creates an edge in the problem space that, all else being equal for the objective, the algorithm aims to sit on. So we get `gen0` @30MW (60tCO2) and `gen1` @50MW (100tCO2, ish).

test/test_lopf_piecewise_efficiency.py
```diff
@@ -106,8 +124,13 @@ def test_piecewise_efficiency_gen() -> None:
...
     n.optimize()
-    assert n.generators_t.p["gen1"].item() == pytest.approx(50.0, rel=1e-3)
-    assert n.generators_t.p["gen0"].item() == pytest.approx(30.0, rel=1e-3)
+    p = n.generators_t.p.iloc[0]
+    assert p.sum() == pytest.approx(80.0, rel=1e-6)
+    assert n.objective == pytest.approx(20 * 80, rel=1e-6)
+    x_points = np.array([0.1, 0.5, 1.0]) * 70
+    fuel_points = x_points / np.array([0.3, 0.4, 0.6])
+    fuel = p["gen0"] / 0.5 + np.interp(p["gen1"], x_points, fuel_points)
+    assert fuel <= 160 * (1 + 1e-6)
```
brynpickering, line 133: This is overkill. We don't need to assert the objective or p.sum() contents to know it's doing something expected. We can move `x_points` and `efficiency_points` to being defined earlier and pass them to the generator addition as well as using them here (`efficiency: {"p_pu": x_points, "efficiency": efficiency_points}`

brynpickering (changes requested):

pypsa/network/transform.py
brynpickering: EDIT: I now understand that it was making a wise choice here and that the previous check wasn't doing what it should have done. Will push a slight change to it.

pypsa/optimization/constraints.py
brynpickering: I _was_ missing something but also Claude didn't identify the main issue (that "_term" dim name was making its way through to the generation of `y_var` in a piecewise module method). Fix incoming.

pypsa/optimization/expressions.py
brynpickering: moved

pypsa/optimization/expressions.py
brynpickering: Updated

pypsa/optimization/global_constraints.py
brynpickering: OK, now I've read the PR description, I understand why this is needed. Will keep with minor cleanup

Issue #1475, opened 2025-12-10
Examples for handling subsidies, penalties, CfDs, PPAs


fneum:
Work out what already works with existing functionality and provide an example notebook.

If not possible, extend features to model these effects.

> subsidy/penalty: "costs to (dis)incentivise investment above/below given thresholds, representing soft constraints (e.g. rebate for providing firm capacity, penalty for exceeding land use limits)"

Irieo:
- CAPEX subsidy (some form of an investment support) is more straightforward and can be modelled either as removing a fraction of annuity or by lowering WACC for a specific project(s) getting subsidy. Depends on a form of subsidy projects get IRL.
- OPEX / production subsidy (eg a feed-in premium) I think can be modeled via marginal_cost_effective (incl. fixed subsidy per €/MWh).
- CfDs are trickier, because they couple market price and revenue. If we want to model it explicitly, say CfD_t = strike_price - λ_t, than we need an MCP model with prices present in the problem formulation. (I have not seen in literature this one but I think) there's a workaround, where one can approximate CfD as effective guaranteed price. Solve a standard LOPF without CfDs -> extract prices -> include CfD as an exogenous revenue stream linked to generation as second point suggests.
- wrt modelling (physical) PPAs, there's plentiful examples with PyPSA code, open GitHub repos and documentation of math & study design within https://irieo.github.io/247cfe.github.io/
- virtual PPAs can be modelled as in example above by relaxing deliverability (geographical matching) pillar

pz-max:
About the PPA example.

[...]

PR #1727, opened 2026-06-12
docs: add CfD/PPA settlement example notebook

MaykThewessen:
Follows up on #1475. As discussed there, this takes the notebook-first route rather than touching `statistics/`.

The notebook (`docs/examples/cfd-ppa-settlement.ipynb`) shows how to settle support schemes as a post-processing step on a solved network, reading only `n.generators_t.p` and `n.buses_t.marginal_price`. The dispatch is unchanged, so no custom constraints or extra variables are needed. Each scheme is one line of arithmetic on a shared per-snapshot top-up rate; the variants differ only in how that rate is defined.

Variants covered:
- Two-sided CfD (symmetric top-up / clawback around a strike)
- One-sided feed-in premium with a price floor (SDE++ style)
- Cap-and-floor revenue stabilisation
- Virtual PPA vs baseload PPA (showing the buyer's shape risk)
- The UK negative-price rule, as an example of a scheme-specific tweak plugged in as a boolean mask

A small toy market (one zone, one week) is engineered to produce both a sustained negative-price spell and a few scarcity hours, so every scheme has something to settle.

Registered in the examples index, the mkdocs nav, and the release notes. The notebook executes cleanly under `mkdocs-jupyter` (`allow_errors: false`) and is ruff-clean with outputs stripped.

Happy to fold the deferred variants (multi-year strike indexation, vintage stacking) into a follow-up, or adjust the scope/placement if you'd prefer.


MaykThewessen:
Added a second notebook, `cfd-multi-year-settlement.ipynb`, covering the multi-year variants I'd flagged as follow-ups: CPI strike indexation over a contract lifetime and vintage stacking of a multi-cohort support portfolio. Same post-processing recipe, just evaluated per year and per vintage; it solves one base year and scales the price path exogenously (noted as a caveat in the notebook).

Happy to split it into its own PR if you'd rather keep this one to the single-year example, or to drop it entirely if it's more than you want here.

MaykThewessen:
Both notebooks run end-to-end under the docs build (`mkdocs-jupyter`, `allow_errors: false`), are ruff-clean, and have their outputs stripped. They only depend on numpy/pandas/matplotlib plus a HiGHS solve, so nothing new for the docs environment. Ready for review whenever you have a moment.

Irieo (changes requested):
Thanks @MaykThewessen for sketching this example! Also, for referencing the #1475 discussion at the top, as all of these examples rest on the assumption that the CfD scheme simply dictate how an exogenous cash flow on fixed prices are calculated. So we abstract from finding the strike-price endogenously, which would require KKTs or some form of bilevel formulations.

It was fun to review, since there is no concrete "end goal" of what exactly we want to achieve in this example. So I stick to your story, and just list few thoughts below on how would I improve it. Feel free to drop any comment if you don't think it strictly improves the example.

### On an illustrative story

- The intro says "a few scarcity hours where the peaking plant sets a price well above the strike", but `ocgt_peak` never runs actually. The peak load in this example is 440 MW while the non-peakers sum to 200+180+150 = 530 MW. Max price is 110 set by `ccgt_high`. To make it all work, I believe you can just shrink the merit order (drop `ccgt_mid` or adjust capacities)?

- Related: the cap in scheme 3  is parametrised to 120 EUR/MWh, but it never ever binds (since highest price in this example is below that) and the results show zero hours when clawback mechanism works. Is it how you intended the example to be? I guess fixing the merit order above will fix it too.

- I went through results of scheme 4 several times, and it feels like what is calculated is fine but what the text says is common sense in PPA terminology inversed (maybe I am missing what you've intended!). How I think about it: the text says the *baseload* buyer carries shape risk "because the block does not follow the wind". Typical convention afaik is the opposite: a pay-as-produced PPA puts profile risk on the *buyer* (they receive wind-shaped volume); a baseload PPA shifts it to the *seller*, who must firm the profile. Does it make sense?

- My little knowledge in cap-and-floor regimesuggest they typically apply to cumulative revenue over some assessment periods (rather than the hourly product, e.g. [here](https://www.ofgem.gov.uk/publications/nemo-link-electricity-interconnector-first-end-period-assessment-epa) or [here](https://assets.publishing.service.gov.uk/media/67d0159ead017430364eeecd/ldes-cf-cap-floor-design.pdf)). Do you have experience with those? I guess you can either word it somehow in example why would it be hourly scheme, or adjust example to net revenue calculation.

- The `storm` period is, how my PhD supervisor would say, [a Chekhov's gun](
 https://en.wikipedia.org/wiki/Chekhov%27s_gun) 

### Other comments

- Move import cell below. The examples typically start with a title and short easy-to-read summary what the illustrative example achieves. 

- for some reason mkdocs '!!! note' admonition is not rendered correctly when I build docs locally from this PR.

- Would it make sense to add a quick dispatch/price dashboard right after the solve, before the settlement sections. I had to plot the results myself to make sense of the numbers later.

- You can use statistics module for some metrics you hard code in an example. I checked that statistics yields same results. I agree that hard-written stuff gives better untuition, but there is an illustrative value to use statistics too. You might present both, and either one or another in a comment, e.g.
  ```python
  market_revenue = (price * energy).sum()  # == n.statistics.revenue().loc["Generator", "wind"]
  capture_price = market_revenue / total_mwh  # == n.statistics.market_value().loc["Generator", "wind"]
  ```
- `ccgt_base/ccgt_mid/ccgt_high` - maybe just `base/mid/peak`?

- Do we need `weights` in revenue function given that snapshot weights are all 1 here? I see illustrative purpose, but we can also trim down the code kicking it out; either is fine.

- `n.add("Carrier", "AC")` to get rid of the consistency warning?

- I would drop the second notebook presenting the same idea of CfD mechanism but with the multi-year revenue accumulation example

isavelli:
Hello @MaykThewessen  , nice examples! 
I agree with the points raised by @Irieo , and as he suggested, I would consider removing the multi-year example (or adding a note explaining that indexing the cashflows requires accounting for inflation in OPEX as well, and discounting at nominal rates, as opposed to keeping everything fixed and using a real rate — but to avoid all this extra explanation, I would suggest dropping the multi-year example).

FabianHofmann:
@MaykThewessen should I take over the reviewer's points? happy to give this a final push

MaykThewessen:
@FabianHofmann  Yes please!

FabianHofmann (approved):
I addressed @Irieo's request and went one more time freshly through the notebook and I really like it! good material, not really complicated but I still learned quite something.

Irieo (approved):

MaykThewessen:
Great work guys! Happy

fneum:
This example turned out very nice!

PR #1739, opened 2026-06-18
fix: piecewise supply/withdrawal direction in energy balance expressions

FabianHofmann:
> [!Note]
> AI written summary and partly code

Follow-up to #1603. Fixes the supply/withdrawal decomposition for piecewise multiport ports and tidies the surrounding data model.

## Problem
`supply() - withdrawal()` did not equal `energy_balance()` for piecewise ports: the piecewise auxiliary variable was added in every direction unfiltered, while the linear coefficients were direction-clipped. A port's piecewise contribution leaked into both `supply()` and `withdrawal()`.

## Fix
- Restrict each port's piecewise contribution to the direction implied by the sign of its breakpoints (`_direct_piecewise`), mirroring the clipping applied to linear coefficients. Mixed-sign curves are already rejected at `add` time, so each port is unambiguously supply or withdrawal.
- Introduce a clear schema-vs-data model: `Component.is_piecewise(attr)` / `Component._piecewise_aux_var(attr)` and `_Multiport._port_coefficient_attr(port)`, replacing ad-hoc name templates (`f"{c}-p{port}_piecewise"`) and `aux_var in m.variables` membership checks in `capex`/`opex`/`energy_balance`.
- Drop the `piecewise` flag from `port_efficiency`; it now returns static/dynamic efficiency only (single responsibility).

## Helper reuse & deduplication
The helpers introduced above are now leveraged across the optimization module, removing the remaining ad-hoc patterns:
- `_port_coefficient_attr(port)` replaces the raw `f"{c._coefficient_attr}{c._port_suffix(port)}"` construction in `optimize.py`, `constraints.py`, and `statistics/expressions.py`. The helper body is now the single source of that template.
- `is_piecewise(attr)` replaces the `not c.piecewise.get(attr, pd.DataFrame()).empty` data-check in the solution-assignment loop in `optimize.py`.
- New `Component._piecewise_schema(attr)` centralizes the recurring `_piecewise_attrs.query("y == @attr").squeeze()` schema lookup. `_piecewise_aux_var` is routed through it, and four scattered call-sites (`optimize.py`, `global_constraints.py`, `constraints.py`, `piecewise.py`) now delegate to it (dropping a `local_dict` workaround in the process).

Schema-vs-data semantics are kept distinct: `is_piecewise` (has data) is only used for genuine data-checks; `_piecewise_schema(...).empty` (schema gate) is left in place where the gate is schema-based. The only remaining `query("y == ...")` call is the deliberate multi-row `.iterrows()` case for costs in `optimize.py`, which is not the single-row squeeze pattern.

## Tests
- `test_supply_minus_withdrawal_equals_energy_balance` — numerical identity regression (the strongest guard).
- Updated `port_efficiency` and piecewise helper tests for the new contract.
- `SimpleNamespace` fakes in `test_optimization_piecewise.py` updated to faithfully expose `_piecewise_schema`.

Verified: ruff + mypy clean; affected suites green.

### DIFF
--- a/pypsa/optimization/expressions.py
+++ b/pypsa/optimization/expressions.py
@@ -382,22 +382,21 @@ class StatisticExpressionsAccessor(AbstractStatisticsAccessor):
 
         @pass_none_if_keyerror
         def func(n: Network, c: str, port: str) -> pd.Series | None:
-            attr = lookup.query("not nominal and marginal_cost").loc[c].index.item()
+            attr = "marginal_cost"
+            attr = lookup.query(f"not nominal and {attr}").loc[c].index.item()
             if attr is None:
                 return None
             var = n.model.variables[f"{c}-{attr}"]
             sns = var.indexes["snapshot"]
 
             c_obj = n.c[c]
-            if c_obj.has_piecewise("marginal_cost"):
-                add_opex = n.model.variables[c_obj._piecewise_aux_var("marginal_cost")]
+            if c_obj.has_piecewise(attr):
+                add_opex = n.model.variables[c_obj._piecewise_aux_var(attr)]
                 var = var.drop_sel(name=add_opex.coords["name"])
             else:
                 add_opex = 0
 
-            opex = (
-                var * n.get_switchable_as_dense(c, "marginal_cost").loc[sns] + add_opex
-            )
+            opex = var * n.get_switchable_as_dense(c, attr).loc[sns] + add_opex
 
             weights = n.snapshot_weightings.objective.loc[sns]
             return self._aggregate_timeseries(opex, weights, agg=groupby_time)
--- a/pypsa/optimization/expressions.py
+++ b/pypsa/optimization/expressions.py
@@ -18,6 +18,7 @@ from packaging import version
 from xarray import DataArray
 
 from pypsa.common import deprecated_kwargs, pass_none_if_keyerror
+from pypsa.components._types.mixin.multiports import _Multiport
 from pypsa.statistics import (
     get_transmission_branches,
     port_efficiency,
@@ -44,6 +45,25 @@ def check_if_empty(expr: LinearExpression) -> bool:
     if USE_EMPTY_PROPERTY:
         return expr.empty
     return expr.empty()
+
+
+def _direct_piecewise(
+    c: Any, y_attr: str, sign: Any, pw: Variable, names: Any, direction: str
+) -> Variable | LinearExpression:
+    """Restrict a piecewise contribution to the requested supply/withdrawal direction.
+
+    A port's direction follows the sign of its breakpoints (``sign * y_attr``),
+    mirroring the clipping applied to linear coefficients. Mixed-sign curves are
+    rejected at ``add`` time, so each port is unambiguously supply or withdrawal.
+    """
+    if direction == "both":
+        return pw
+    y = c.piecewise[y_attr].xs(y_attr, level="attribute", axis=1)[names]
+    withdraws = (sign * y < 0).any()
+    if direction == "withdrawal":
+        return -1 * pw.sel(name=names[withdraws])
+    else:
+        return pw.sel(name=names[~withdraws])
 
 
 class StatisticExpressionsAccessor(AbstractStatisticsAccessor):
@@ -222,16 +242,8 @@ class StatisticExpressionsAccessor(AbstractStatisticsAccessor):
 
             costs = c.static[cost_attribute][capacity.indexes["name"]]
 
-            piecewise_attr = (
-                n.c[component]._piecewise_attrs.query("y == @cost_attribute").squeeze()
-            )
-            aux_var = (
-                f"{component}-{piecewise_attr.aux_variable}"
-                if not piecewise_attr.empty
-                else None
-            )
-            if aux_var in m.variables:
-                add_capex = m.variables[aux_var]
+            if c.is_piecewise(cost_attribute):
+                add_capex = m.variables[c._piecewise_aux_var(cost_attribute)]
                 capacity = capacity.drop_sel(name=add_capex.coords["name"])
             else:
                 add_capex = 0
@@ -376,16 +388,9 @@ class StatisticExpressionsAccessor(AbstractStatisticsAccessor):
             var = n.model.variables[f"{c}-{attr}"]
             sns = var.indexes["snapshot"]
 
-            piecewise_attr = (
-                n.c[c]._piecewise_attrs.query("y == 'marginal_cost'").squeeze()
-            )
-            aux_var = (
-                f"{c}-{piecewise_attr.aux_variable}"
-                if not piecewise_attr.empty
-                else None
-            )
-            if aux_var in n.model.variables:
-                add_opex = n.model.variables[aux_var]
+            c_obj = n.c[c]
+            if c_obj.is_piecewise("marginal_cost"):
+                add_opex = n.model.variables[c_obj._piecewise_aux_var("marginal_cost")]
                 var = var.drop_sel(name=add_opex.coords["name"])
             else:
                 add_opex = 0
@@ -543,14 +548,15 @@ class StatisticExpressionsAccessor(AbstractStatisticsAccessor):
             )
 
         @pass_none_if_keyerror
-        def func(n: Network, c: str, port: str) -> pd.Series:
-            var = self._get_operational_variable(c)
+        def func(n: Network, component: str, port: str) -> pd.Series:
+            c = n.c[component]
+            var = self._get_operational_variable(component)
             sns = var.indexes["snapshot"]
             # negative branch contributions are considered by the efficiency
-            dynamic_efficiency = port_efficiency(n, c, port=port, dynamic=True)
+            dynamic_efficiency = port_efficiency(n, component, port=port, dynamic=True)
             if isinstance(dynamic_efficiency, pd.DataFrame):
                 dynamic_efficiency = dynamic_efficiency.loc[sns]
-            sign = n.c[c].static.get("sign", 1.0)
+            sign = n.c[component].static.get("sign", 1.0)
             weights = n.snapshot_weightings.generators.loc[sns]
             coeffs = DataArray(dynamic_efficiency * sign)
             if direction == "supply":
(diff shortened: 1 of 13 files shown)


pypsa/optimization/expressions.py
```diff
@@ -222,16 +242,8 @@ def func(n: Network, component: str, port: str) -> pd.Series | None:
...
-                n.c[component]._piecewise_attrs.query("y == @cost_attribute").squeeze()
-            )
-            aux_var = (
-                f"{component}-{piecewise_attr.aux_variable}"
-                if not piecewise_attr.empty
-                else None
-            )
-            if aux_var in m.variables:
-                add_capex = m.variables[aux_var]
+            if c.is_piecewise(cost_attribute):
```
brynpickering: Perhaps this should be `has_piecewise`, noting that for a given component class there can be simultaneously piecewise, static, and dynamic versions of an attribute, each assigned to different components

pypsa/optimization/expressions.py
```diff
@@ -563,13 +569,15 @@ def func(n: Network, c: str, port: str) -> pd.Series:
...
-                pw_var = 0
+
+            pw_var = 0
+            if isinstance(c, _Multiport) and c.is_piecewise(
+                y_attr := c._port_coefficient_attr(port)
+            ):
+                pw = sign * n.model.variables[c._piecewise_aux_var(y_attr)]
+                names = pw.coords["name"].values
+                coeffs.loc[{"name": names}] = 0
+                pw_var = _direct_piecewise(c, y_attr, sign, pw, names, direction)
```
brynpickering, line 579: I don't think this actually solves the problem that I initially raised.

Here's an example network definition that showcases the issue:

<details>

<summary>Network definition</summary>

```py
import pypsa
import pandas as pd
import numpy as np

idx = pd.Index([0, 1, 2])

# 3. Part-load efficiency
eta_rated = 0.7  # efficiency at full load
eta_min = 0.4  # efficiency at zero load
P_frac_min = 0  # minimum load fraction (η is finite at zero load)
P_max = 1000


def get_part_load_eff_curve(P_frac):
    return (eta_rated - eta_min) * np.sqrt(P_frac) + eta_min


def get_load_in_curve(P_out):
    eff = get_part_load_eff_curve(P_out / P_max)
    return P_out / eff


p_pu_breakpoints = np.array([0, 0.25, 0.5, 0.75, 1.0])
efficiencies = get_part_load_eff_curve(p_pu_breakpoints)
eff_dict = dict(zip(p_pu_breakpoints, efficiencies))

n = pypsa.Network()
n.set_snapshots(idx)
n.add("Bus", "electricity", carrier="electricity")
n.add("Bus", "hydrogen", carrier="hydrogen")
n.add("Carrier", ["electricity", "hydrogen", "generator", "load", "h2-lin", "h2-pw"])
n.add(
    "Generator", 
    "gen", 
    carrier="generator", 
    bus="electricity", 
    p_nom_max=30, 
    capital_cost={0: 10, 10: 20, 30: 50}, 
    p_nom_extendable=True
)
n.add(
    "Load", 
    "demand", 
    carrier="load", 
    bus="hydrogen", 
    p_set=pd.Series([1, 2, 3], index=idx)
)
n.add(
    "Link", 
    "electrolysis1", 
    carrier="h2-lin", 
    bus0="electricity", 
    bus1="hydrogen", 
    p_nom=10, 
    p_nom_extendable=False, 
    efficiency=0.5
)
n.add(
    "Link", 
    "electrolysis", 
    carrier="h2-pw", 
    bus0="electricity", 
    bus1="hydrogen", 
    p_nom=10, 
    p_nom_extendable=False, 
    marginal_cost=eff_dict, 
    efficiency=eff_dict, 
    capital_cost=1, 
    delay=1
)

n.optimize.create_model()
```

</details>

Now, if I call `n.optimize.expressions.energy_balance()`

I get:

```py
LinearExpression [group (component, carrier, bus_carrier): 4]:
--------------------------------------------------------------
[(Generator, generator, electricity)]: +1 Generator-p[0, gen] + 1 Generator-p[1, gen] + 1 Generator-p[2, gen]
[(Link, h2-pw, electricity)]: -1 Link-p[0, electrolysis-pw] - 1 Link-p[1, electrolysis-pw] - 1 Link-p[2, electrolysis-pw] + 0.5 Link-p[0, electrolysis-lin] + 0.5 Link-p[1, electrolysis-lin] + 0.5 Link-p[2, electrolysis-lin]
[(Link, h2-lin, electricity)]: -1 Link-p[0, electrolysis-lin] - 1 Link-p[1, electrolysis-lin] - 1 Link-p[2, electrolysis-lin] + 1 Link-p1_piecewise[0, electrolysis-pw] + 1 Link-p1_piecewise[1, electrolysis-pw] + 1 Link-p1_piecewise[2, electrolysis-pw]
[(Load, load, hydrogen)]: -6
```

But `Link-p1_piecewise` should be at the `hydrogen` bus, _not_ the `electricity` bus. At the moment, there is no "energy balance" at the `hydrogen` bus since it's just withdrawing 6 units and not having anything supplied. This differs from the `stats` energy balance, which does recognise the outflow at the `hydrogen` bus:

```py
n.stats.energy_balance()

component  carrier    bus_carrier
Generator  generator  electricity    10.64111
Link       h2-lin     electricity    -1.95814
           h2-pw      electricity    -8.68297
           h2-lin     hydrogen        0.97907
           h2-pw      hydrogen        5.02093
Load       load       hydrogen       -6.00000
dtype: float64
```

It's made more obvious if we aggregate a bit more:

```py
n.optimize.expressions.energy_balance(groupby=["bus"], aggregate_across_components=True)

LinearExpression [bus: 2]:
--------------------------
[electricity]: +1 Generator-p[0, gen] - 1 Link-p[0, electrolysis-lin] + 1 Generator-p[1, gen] ... +1 Link-p1_piecewise[0, electrolysis-pw] + 1 Link-p1_piecewise[1, electrolysis-pw] + 1 Link-p1_piecewise[2, electrolysis-pw]
[hydrogen]: -6
```

Perhaps this is intended `pypsa` convention, that the remote bus of a link is never explicitly recognised in the linear expressions, but it means that if you limit your focus to just the `hydrogen` bus, you don't see the possible withdrawals via the links in the other direction

For `supply`, we get:

```py
LinearExpression [group (component, carrier, bus_carrier): 3]:
--------------------------------------------------------------
[(Generator, generator, electricity)]: +1 Generator-p[0, gen] + 1 Generator-p[1, gen] + 1 Generator-p[2, gen]
[(Link, h2-pw, electricity)]: +0.5 Link-p[0, electrolysis-lin] + 0.5 Link-p[1, electrolysis-lin] + 0.5 Link-p[2, electrolysis-lin]
[(Link, h2-lin, electricity)]: +1 Link-p1_piecewise[0, electrolysis-pw] + 1 Link-p1_piecewise[1, electrolysis-pw] + 1 Link-p1_piecewise[2, electrolysis-pw]
```

To me, this doesn't make much sense since the `supply` being defined for the links is defined on the wrong bus, right? It is suggesting that you can get `0.5 Link-p` of `electrolysis-lin` supplied at the `electricity` bus. But this is a uni-directional link pointing _away_ from the `electricity` bus, so you can only withdraw from it, not supply. Indeed, that's supported by the `stats` module equivalent of this expression:

```py
n.stats.supply(groupby=["carrier", "bus_carrier"])

component  carrier    bus_carrier
Generator  generator  electricity    10.64111
Link       h2-lin     hydrogen        0.97907
           h2-pw      hydrogen        5.02093
dtype: float64
```

brynpickering (commented):
The added helper functions clean things up a lot - thanks!

However, my initial concern w.r.t. supply/withdrawal and the energy balance in the expressions module more generally hasn't been resolved. In fact, I think it _might_ point to a more general issue with the expressions module that has always existed. See my more detailed review comment for more.

pypsa/optimization/expressions.py
FabianHofmann: you are exactly right! the problem isn't with the data anymore - that seems correct. it has to do with the allocation to the bus from the bus carrier (`_get_grouping`). I am getting closer and sure I can resolve it.

FabianHofmann: @brynpickering it was deep down in the alignment on the linopy level. where linopy now requires explicit `join` parameters in the merge function. The convention changed there without a clear warning which is not ideal... but it is also super hard to make the transition fully smooth. pushing the change. the result now looks like 

```
In [11]: n.optimize.expressions.energy_balance()
Out[11]:
LinearExpression [group (component, carrier, bus_carrier): 6]:
--------------------------------------------------------------
[(Generator, generator, electricity)]: +1 Generator-p[0, gen] + 1 Generator-p[1, gen] + 1 Generator-p[2, gen]
[(Link, h2-lin, electricity)]: -1 Link-p[0, electrolysis1] - 1 Link-p[1, electrolysis1] - 1 Link-p[2, electrolysis1]
[(Link, h2-lin, hydrogen)]: +0.5 Link-p[0, electrolysis1] + 0.5 Link-p[1, electrolysis1] + 0.5 Link-p[2, electrolysis1]
[(Link, h2-pw, electricity)]: -1 Link-p[0, electrolysis] - 1 Link-p[1, electrolysis] - 1 Link-p[2, electrolysis]
[(Link, h2-pw, hydrogen)]: +1 Link-p1_piecewise[0, electrolysis] + 1 Link-p1_piecewise[1, electrolysis] + 1 Link-p1_piecewise[2, electrolysis]
[(Load, load, hydrogen)]: -6
```
looks right to me. can you confirm?

FabianHofmann: can confirm this matches now with statistics expressions: 

image

same for supply and withdrawal

brynpickering: Thanks @FabianHofmann - that does the trick!

brynpickering (approved):

brynpickering (commented):
Just looking at the failing docs build, it seems that something has changed with how efficiencies are grabbed (perhaps the change to the `port_efficiency` method?). What we get now in the chained hydro notebook is:

```
Encountered nan's in static data for columns ['efficiency2'] of component 'Link'.
```

FabianHofmann:
> Just looking at the failing docs build, it seems that something has changed with how efficiencies are grabbed (perhaps the change to the `port_efficiency` method?). What we get now in the chained hydro notebook is:
> 
> ```
> Encountered nan's in static data for columns ['efficiency2'] of component 'Link'.
> ```

@brynpickering I think I found it. the delay section is reassigning a variable

```
            for d_names, delay, is_cyclic in _iter_balance_delay(c, names, port_suffix):
                if delay > 0:
                    var = _apply_delay_shift(n, sns, c, var, d_names, delay, is_cyclic)

            expr = var * coeff.sel(name=names) + p_piecewise
```

pushing a fix here and case-distinct multiply with coeffs based on type of variable

brynpickering:
Thanks for the various micro fixes @FabianHofmann. I'm going to merge this into the main PR so we get the full test suite running on it.

Issue #1735, opened 2026-06-16
Make Cournot strategic behaviour an official attribute


FabianHofmann:
The wonderful example (`docs/examples/imperfect-competition.ipynb` - #1570) showcases a full Cournot–Nash equilibrium. I follows the ideas of manually adding the markup to the objective like 

```python
m = n.optimize.create_model()
q = m["Generator-p"].sel(name="gas_producer_1")
m.objective = m.objective.expression + (b * cv / 2) * (q * q).sum()  
n.optimize.solve_model()
```

I propose to fully integrate this functionality into the core via a new attribute `cournot_cv`. Then beginners can directly use it, advanced user don't have to worry about more complex like applications like in stochastic optimization. Also one would also not need to rely on adding the cost term explicitly to marginals which would distort statistics.

fneum:
I would be a bit hesitant because the conditions under which this example can be used is quite narrow (it cannot be mixed with perfectly inelastic demands as set with `n.loads_t.p_set`). @Irieo

Irieo:
@FabianHofmann Right, I was thinking about this idea when working on the notebook example. What stopped me from expanding this to PR is that the markup term for the fictitious objective annex depends **[and requires](https://www.jstor.org/stable/2950588)** the affine (linear) elastic demand. We can model it perfectly with combination of marginal_cost/marginal_cost_quadratic (like the notebook shows). However, for the implementation of `cournot_cv` as a new attribute, we'd need both documentation of this limitation and elaborate consistency checks. More, we'd need to think through how link/line losses may propagate into the the markup term (and elaborate the math and/or for endogenous  line losses possibly guard with "not implemented"), and so on. It is doable, but as all things requires care and time. Happy to chat about it.

PR #1751, opened 2026-06-24
Text correction in imperfect competition example

Irieo:
Closes # (if applicable).

I wanted to reply in #1735 and thus dived again into research papers on the fictitious objective topic (seminal work by [Slade 1994](https://www.jstor.org/stable/2950588) and [Hobbs 2001](https://hobbsgroup.johnshopkins.edu/docs/papers/IEEE_Hobbs_2001_LCP00918286.pdf)); and I realized that my reference to Hobbs paper in #1570 wasn't fully correct. This PR fixes wrong line in one markdown cell of the notebook example.

Summary: 
- *wrong take*: one of conditions for the existence of a fictitious objective is that there is no binding network constraints that create asymmetric shadow prices across strategic players. -> It is incorrect since the three-bus example in Section V of the paper actually solves the network case where the congestion is binding. That equilibrium is still the solution to the single QP.  So the binding network constraints with nonzero shadow prices do not destroy the fictitious objective.
- *correct take*: one of conditions for the existence of a fictitious objective is that players treat transmission prices as fixed (the player's perceived derivative of the transmission shadow price with respect to its own output is zero), so congestion enters every player's problem as a linear term. 

Why this matters: since congested networks don't break the fictitious objective (but endogenous congestion-price conjectures do), the fictitious objective formulation can scale to large transmission constrained systems. 

## Checklist


FabianHofmann (approved):
makes sense

PR #1755, opened 2026-06-25
Add functionality to handle binary `status` of committable components

brynpickering:
Adds functionality to handle the committable `status` of components in piecewise.

@FabianHofmann I had to split the piecewise definition in two whenever `status` is used (one pw definition with `active` used, one without) as otherwise it creates unexpected results in linopy. Namely, it seems to force the dispatch of components without a `status` variable to zero (via the `_active` pw constraint). I don't have time to investigate this further right now so thought I'd check with you in case it is my misinterpretation or an actual issue with the linopy implementation. Either way, the current implementation is a workaround that passes the tests.

I'm opening this in a separate PR as we might want to leave it out of #1603 and only come to consider this when it is rebased onto `master` post-merge. I don't mind either way. If it highlights issues that should already be solved in #1603 it might be worthwhile to already merge it in there.

## Checklist

### DIFF
--- a/pypsa/optimization/piecewise.py
+++ b/pypsa/optimization/piecewise.py
@@ -67,6 +67,7 @@ def define_piecewise(
     sign: SIGNS_T,
     cumulative_attr: bool,
     extra_options: Iterable[PiecewiseOptions],
+    status: Variable | None = None,
     invert_attr: bool = False,
     y_var: Any | None = None,
     method: PWL_METHOD = "auto",
@@ -99,6 +100,8 @@ def define_piecewise(
         Whether the y-axis breakpoints represent marginal values.
         If True, the integral of the piecewise curve will be used to define y breakpoints.
         If False, the nominal values at each x breakpoint will be used to define y breakpoints.
+    status : linopy.Variable or None, optional
+        The optimisation variable for the binary status of the component (e.g. generator on/off).
     y_var : linopy.Variable or None, optional
         The optimisation variable for the y-axis of the piecewise constraint.
         If None, a new auxiliary variable will be created and used instead.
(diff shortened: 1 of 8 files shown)


pypsa/optimization/piecewise.py
```diff
@@ -158,14 +161,49 @@ def define_piecewise(
         if names.empty:
             continue
 
+        if status is not None:
+            status_no_null = status.sel(name=~status.isnull().groupby("name").all(...))
```
brynpickering: messy workaround for masked variables/exprs that require us to only keep names of cases where they are unmasked. Otherwise we hit on the issue of setting an `active` constraint on non-committable components which will effectively force their operation to zero.

pypsa/optimization/piecewise.py
```diff
@@ -158,14 +161,49 @@ def define_piecewise(
...
+                        opt_sign,
+                    ),
+                    (
+                        x_var.sel(name=status_names),
+                        x_breakpoints.sel(name=status_names),
+                    ),
+                    method=opt_method,
+                    active=opt_status,
+                    name=aux + "_status",
+                )
```
brynpickering: I feel like this should be solved instead on the linopy side, to have two different pw implementations when `active` is supplied, one for cases where `~active.isnull()` and the other for cases where `active.isnull()`. Would make things cleaner here and would allow for partial LP convex bound implementations when a user provides `active` arrays with only partial intersection with the `x` and `y` var dims.

FabianHofmann: very good point!

FabianHofmann:
thanks @brynpickering for tackling the last item on the long todo list. I would propose to merge this in and add a tracking issue. after linopy integrates the feature of a partial `active` flag we update the code in pypsa too.

FabianHofmann:
@brynpickering while we have been tackling the partial `active` gate in linopy already, I have adjusted the code here too to skip the double constraint assignment, pulling this into the main pr

PR #1766, opened 2026-06-29
Fix: properly assign status to non-committable assets in mixed asset set

FabianHofmann:
Closes #1644 

When a network has both committable and non-committable assets with ramp limits, rolling-horizon optimization with unit commitment can lead to false-positive infeasibility. This is caused as `status` default 0, is propagated to non-commitables and not correctly set to 1 - meaning they were treated as turned off in each horizon. The fix treats non-committable components as always "on".

## Checklist


luciana-marques (approved):
Nice catch! We did have (a lot of) infeasibilities related to ramping in our workflow, but I have not thought it could be a matter of having committable and non-committable units in the rolling horizon approach. To test this we would need to change the version of pypsa (we are thinking about doing it, but it all depends on the time we have). I will keep you posted. Thanks!

pypsa/optimization/constraints.py
```diff
@@ -827,7 +827,7 @@ def define_ramp_limit_constraints(
     if is_rolling_horizon:
         start_i = n.snapshots.get_loc(sns[0]) - 1
         p_init = c.da[hist_attr][start_i]
-        s_init = c.da.status[start_i].fillna(1)
+        s_init = c.da.status[start_i].where(c.da.committable, 1).fillna(1)
```
lkstrp, line 830: Further down we have the same bug isnt it?
```
            s_init_ext = c.da.status[start_i].sel(name=ext_main_names).fillna(1)
```
Where we have the same for non committable and extendable, but I am not very sure

lkstrp (dismissed):

pypsa/optimization/constraints.py
FabianHofmann: good catch! done

fneum (approved):

PR #1774, opened 2026-07-02
Add stochastic optimization by decomposition via mpi-sppy

DLWoodruff:

Adds an alternative to PyPSA's EF solve for two-stage
stochastic problems (built with `n.set_scenarios`): solve them **by
decomposition** with [mpi-sppy](https://github.com/Pyomo/mpi-sppy)

PyPSA writes one optimisation file per scenario (`.lp`/`.mps`) plus small metadata files (first-stage nonant names + probability, per-nonant PH rho), and mpi-sppy's file-based scenario loader reads them. The new module imports neither mpi-sppy nor Pyomo.

New `n.optimize` accessor methods:

- `solve_stochastic_mpisppy` — inline driver: write files, then run the mpi-sppy solve as a blocking subprocess, then read the first-stage solution back onto the network.
- `write_stochastic_problem_mpisppy` / `read_stochastic_solution_mpisppy` — the
  same workflow decoupled, so the solve can run as its own scheduler job (e.g.
  SLURM) between the two calls.

First-stage capacities are read back as `*_nom_opt`. Second-stage dispatch and
scenario-conditional marginal prices are recovered with `dispatch="resolve"`.

Dependencies: needs `mpi-sppy` itself, plus a new optional `pypsa[mpisppy]` extra
(`mpi4py`, `mip`). The extra is intentionally PyPI-clean and does **not** list
`mpi-sppy`, which is installed separately
(`pip install "git+https://github.com/Pyomo/mpi-sppy.git"`) until a PyPI release
ships the MPS-file-loader support — see the note in `pyproject.toml`.

Docs: new user guide
[Stochastic Optimization by Decomposition](docs/user-guide/optimization/stochastic-decomposition.md)
(install, options, inline + decoupled/SLURM worked examples, dispatch recovery),
API reference entries, and a release note.

Validation: PH converges to the same objective as the native EF on
the test cases with unit tests added.

## Checklist

- [x] PR description is edit and written by me. Any potentially verbose AI-generated content has been deleted (see [Contributing guidelines](https://docs.pypsa.org/latest/contributing/contributing/)).


DLWoodruff:
These test failures seem to be unrelated to this PR.

FabianHofmann:
thanks @DLWoodruff as discussed online if it still up for discussion whether the communication with the solver should not live in linopy directly. I would strongly vote for the latter to not introduce parallel communication streams to solvers. @coroa will also likely spend some thoughts on this when he is back

DLWoodruff:
The thing that has been implemented already is called "Loose Coupling" in 
https://mpi-sppy.readthedocs.io/en/latest/agnostic.html 
and the thing you are talking about is "Tight Coupling" which takes up most of the document. Even though there is a lot of documentation there, it is probably not enough. Since linopy is a python-based AML, I recommend following the Pyomo agnostic example in terms of structure as opposed to the AMPL example. Tight coupling is not being actively used by anyone so far as I know, so please let me know what goes wrong and I will try to fix it quickly.

PR #1794, opened 2026-07-09
ci: commit and use `uv.lock` file

lkstrp:
- Commit `uv.lock` file, which is autoupdated once a week via dependabot and with a cooldown of 7 days
- PR tests and doc builds run on lock file. This should make PR tests also more stable
- Daily tests and master still run on updated env, so we catch breaking upstream dependencies


FabianHofmann:
`requirements.txt` was deleted on purpose ? not in the pr summary, hence I am asking

PR #1795, opened 2026-07-09
ci: run reduced test matrix on PRs

lkstrp:
Needs #1794 to be merged first (for clean diff)

PRs and master don't run full test matrix anymore, just min/max Python on Ubuntu. I don't remember a case where this would not have catched any regression. Full matrix is still running daily and before a release. This should speed up our CI runs


FabianHofmann:
@lkstrp I adjusted the GH setting for required tests for this to pass. note the current setup still requires manual updates of the GH rule set as soon as a new python version is released. I would say that is fine but of course would prefer a fully automated system. one approach for this could be to define one collection test which gates all tests and this use this as the required tests by GH.

lkstrp:
@FabianHofmann 

You mean this manually requires to update the GH settings with every release? I also think this is fine. We manually update new Python versions as well in https://github.com/PyPSA/PyPSA/blob/dfc3bc3ae5991b13c88afe6d648dd471e92b458e/pyproject.toml#L17-L20

FabianHofmann:
> @FabianHofmann 
> 
> 
> 
> You mean this manually requires to update the GH settings with every release? I also think this is fine. We manually update new Python versions as well in https://github.com/PyPSA/PyPSA/blob/dfc3bc3ae5991b13c88afe6d648dd471e92b458e/pyproject.toml#L17-L20

Yes, this is what I meant. Makes sense!

PR #1822, opened 2026-07-20
ci: run dependebot PRs only on direct deps

lkstrp:
This should remove the mass of dependabot PRs recently created. Those deps are now included in the weekly one

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,3 +29,6 @@ updates:
     uv:
       patterns:
       - '*'
+      update-types:
+      - minor
+      - patch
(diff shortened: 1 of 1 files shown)


FabianHofmann:
just to understand. this would mean bumps of major versions would be combined in the collective pr and not be explicit?

lkstrp:
> just to understand. this would mean bumps of major versions would be combined in the collective pr and not be explicit?

That is happening already. This only changes that we don't get PR's for the indirect deps (deps of deps, stuff not listed in `pyproject.toml`). They go in with the group PR. 

We could add a extra PR per major release of deps if you want. 

EDIT: Let's do it, I just checked the deps. This would have been 5 major releases this year, which is not overwhelming

FabianHofmann (approved):

PR #1827, opened 2026-07-21
[Do Not Merge] linopy 0.9 backwards compat

FabianHofmann:
DO NOT MERGE THIS PR. 

This pr is solely used for tracking the backwards compatibility of linopy on new semantics https://github.com/PyPSA/linopy/pull/717 in legacy mode. It is not meant to be merged at any time. Its purpose is to ease tracking the backwards compatibility with large upcoming changes in linopy v0.9. The latter would raise a `LinopySemanticsWarning` wherever the semantics change when opting-in to "v1" mode as well as linopy `v1.0`.

### DIFF
--- a/.github/workflows/test-models.yml
+++ b/.github/workflows/test-models.yml
@@ -123,7 +123,10 @@ jobs:
         # Allow upgrading internal packages without exclude-newer cap
         sed -i '/^exclude-newer\s*=/d' pixi.toml
         pixi remove ${{ env.PACKAGE_NAME }}
-        pixi upgrade linopy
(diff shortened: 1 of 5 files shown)


FBumann:
@FabianHofmann some tests fail. Are the tests just flaky here, or is there sth to fix?

FabianHofmann:
> @FabianHofmann some tests fail. Are the tests just flaky here, or is there sth to fix?

there is one more small issue in the docs creation related to the dim order of grouped objects. in linopy v0.8 the order is actually not the same as in xarray equivalent results. but with https://github.com/PyPSA/linopy/pull/860 it now is. so I would adjust the docs issue here and stick to the current implementation in linopy

Issue #1785, opened 2026-07-06
Have all data tables as views on a single, long-form source of truth table


brynpickering:
Instead of lots of disparate concrete data tables for components and properties, we would benefit from storing the data in a single location. To avoid completely breaking the current API, all our static and dynamic tables could then become views on the long-form data. What this breaks is the ability to directly edit those tables to make changes to the model definition. What we might then need is an explicit `update` method. The benefit of this is we can track updates explicitly and then make it clear to the user when results are up-to-date with updates to the input data.

There is the option to have static data to be concrete and only dynamic as views, since it is more common to want to edit entries in static using `pandas` functionality, which people are most comfortable with.

Issue #1474, opened 2025-12-09
Properties


FabianHofmann:
## Problem Statement
PyPSA currently treats non-physical entities (carriers, shapes, line types, etc.) as regular components alongside physical ones (generators, lines, buses). This mixing makes it difficult to:
- Distinguish between physical and metadata components == `property`s programmatically
- Establish hierarchical relationships between metadata layers
- Perform efficient grouped operations in statistics and plotting
- Validate cross-component references

## Idea
`property`s in PyPSA are data components representing non-physical entities and abstract attributes in dense format such as `n.carrier`, `n.country`, `n.shapes`. 
They are mostly used in internal functionalities for statistics, plotting and wherever mapping to meta data is needed. 

`Property`s are layers for dense information which can be referenced by (physical) components (Generator, Line, etc) or other properties through columns in the static dataframe. The latter is referred to as a nested API approach. To define and use it the user typically would 
1. initialize the meta component `n.property.add("x", attr1="a")`
2. reference the property in a column of a component, `n.generators["x"]`
3. refer to it in `statistics` (or similar modules) `n.stats.optimal_capacity(groupby="x")` or `n.stats.optimal_capacity(groupby=("x", "attr1"))` (we still have to decide on the API here!)

### 1. Property Class
Create a distinct `Property` base class that:
- Inherits from the existing `Components` base class
- Adds metadata-specific functionality (hierarchical relationships, validation)
- Clearly identifies non-physical entities

### 2. Nested API & Hierarchical Relationships
Enable parent-child relationships between components:
```python
# Define property hierarchy
n.properties.countries.add("DE", name="Germany")
n.properties.bidding_zones.add("DE_50Hz", country="DE", name="Germany 50Hz")
n.properties.nodes.add("node_1", bidding_zone="DE_50Hz")

# Reference in physical components
n.generators["node"] = "node_1"  # Automatically inherits bidding_zone and country

# Use in statistics with automatic hierarchy traversal
n.statistics.capacity(groupby="country")  # Aggregates through bidding_zone → country
n.statistics.capacity(groupby="bidding_zone.tso")  # Access nested attributes
```

## Example Use Case - Geographical Maps with different resolution

[...]

PR #1837, opened 2026-07-24
feat: data records/ experimental self-describing long-form parquet format

### LINKED ISSUE #1792: Tabular Parquet storage format for networks
The current `.nc` files are just the standard IO type used by PyPSA. For this app they have tons of disadvantages and are just not made for it. PyPSA can't lazy load at all currently, for a small plot the full `.nc` file needs to be loaded into memory. The schema is very PyPSA specific and isn't cleanly structured and can't be easily expanded. ...

Proposed new data record/ schema:
- Schema derived by PyPSA, owned by App so we can adapt as needed.
- A data record is just an individual representation with data of that schema.
- Processing multiple records (queries for analytics, but could be for anything) can just simply work by pointing to N records, when the query and processing generalises over multiple dimensions. I should be able to get the energy balance of a single data record as well as multiple data records combined via the same defined query. 
- Multiple scenarios are not represented by a single record. But the same paros logic can work on multiple records, which allows processing of multiple scenarios.

```
DataRecordA/
  ├── manifest.json          # version, attribute catalog — immutable, user may not change anything here
  ├── snapshots.parquet
  ├── periods.parquet
  ├── components.parquet
  ├── scenarios.parquet      # for stochastic optimization, not workflow scenarios
  ├── data/
  │   └── <attr>.parquet     # ComponentType | component | snapshot | scenario | period | value
  └── results/
      └── <attr>.parquet     # ComponentType | component | snapshot | scenario | period | value
```

To be discussed/ unclear:
- [ ] When should a Data Record be immutable? Should it? For example snapshotting it before or after a job in a workflow
- [ ] Should a single data record store data of multiple results? Or would you rather create one data record per result and combine them for shared analytics?

Todos:
- [ ] Implement first version
- [ ] Draft Collections of multiple data records. Similar to statistics module in PyPSA which works on a single Network, as well as a collection of Networks.

lkstrp:
Closes #1792 
Ref #1785 

- Adds a long form tabular IO store and data record based on Parquet. Find the full specifications in the docs/ and below
- The feature is marked as experimental. I will have a couple of follow up PRs for this. And I expect that we will further learn stuff and make more changes
- So far, it only affects IO. A new `c.long` accessor provides the long form format as a pandas representation. But this is only a view. In memory, the data still sits in `c.static` and `c.dynamic`. To investigate how we can migrate these will be the fun part in the future
- The main benefit is simply that the storage format uses the long form and is converging toward something self describing. This is useful and necessary for other tools, including GUIs (PyPSA App, Manta, etc.), but PyPSA itself does not yet benefit from it. For that more steps are needed (see below)
- This is also shown in the benchmarks. The melt and stacking needed to get to the long form currently eats up all benefits of the format. I will follow up with more infos here, also interested on what your @coroa findings are

Next steps (PRs, not in here)
1. Per table lazy loading: Load a df (`c.static` and all dfs in `c.dynamic`) only when accessing.
2. IO for network collections (add arbitray amount of dimensions to one compact long form version)
3. Migrate all attribute definitions of current `.csv`s in manifest. Including correct dim definitions.
4. Add validation query language to `manifest.json` which can validate straight on parquet (DuckDB) as well as in memory representation (pandas/ numpy) 

<details>
<summary>Specification</summary>

# Parquet Store

A parquet store is a directory of parquet files with a manifest, designed to
be self describing. The files plus one decode rule are enough to read the data
without PyPSA. Parquet is columnar and compressed, small and fast to scan, and
readable from many engines (pandas, polars, DuckDB, Arrow, R, Spark...), also
directly from cloud object storage. Install the optional dependencies via
`pip install pypsa[parquet]`.

To **export** a network run [`n.export_to_parquet()`][pypsa.network.io.NetworkIOMixin.export_to_parquet].
To **import** run [`n.import_from_parquet()`][pypsa.network.io.NetworkIOMixin.import_from_parquet]
or simply provide the path in the [`pypsa.Network`][] constructor.

``` py
n.export_to_parquet("foo/bar")
n_import = pypsa.Network("foo/bar")
```

## Store format

**Layout.** A store is a directory containing a manifest and parquet files:

```
my_network/
    dims/
        components/<Type>.parquet    members of a component type with static inputs
        snapshots.parquet            snapshot axis and weightings
        periods.parquet              period axis and weightings, only if multiperiod
        scenarios.parquet            scenario axis and weights, only if stochastic
    inputs/<attr>.parquet            one varying input per file
    outputs/<attr>.parquet           one result per file
    manifest.json                    see Manifest below
```

Solving only adds files under `outputs/`. Each `dims/components/<Type>.parquet`
is a subset of `c.static`: only the columns for attributes that cannot vary
across snapshots, scenarios or periods. Varying attributes live under `inputs/`
instead.

**Long schema.** Every file under `inputs/` and `outputs/` has the columns

```
component_type | name | snapshot | scenario | period | attribute | value
```

Each file holds exactly one attribute, so `value` carries that attribute's
dtype.

**Decode rule.** A row's `value` applies to every combination of its null
dimension columns, taken from the axis tables under `dims/`. Rows never
overlap. Anything not covered by a row, including an attribute with no file,
takes the attribute's `default` from the manifest. The writer mirrors the
in-memory representation, so a round-trip restores every value to the
container it came from.

**Manifest.** `manifest.json` holds the non-tabular network data and the
self-description of the store.

```json
{
    "format": "pypsa-parquet",
    "format_version": 1,
    "attributes": {"name": "my_network", ...},
    "meta": {...},
    "crs": "...",
    "attribute_catalog": {
        "Generator": {
            "p_max_pu": {
                "dims": ["name", "snapshot"],
                "default": 1.0
            },
            ...
        },
        ...
    }
}
```

The catalog is network independent, identical for every network with the
same custom attributes. Per attribute it records the allowed `dims` and
the `default` used wherever the files are silent. An attribute's actual
axes compose from its `dims` and the axis files under `dims/`. Reading
never depends on the catalog, just writing does, since it defines which
shapes are valid input.

!!! warning

    The catalog will be expanded in a future version to become the single
    source of truth for attribute metadata.

</details>

Also pinging @FBumann and @FabianHofmann since just discussed

## Checklist


FabianHofmann:
I would say 'c.table' would be a more intuitive accessor name.

pypsa/components/array.py
```diff
@@ -355,3 +416,99 @@ def _as_xarray(self, attr: str) -> xr.DataArray:
...
+    def _stack_dynamic(self, wide: pd.DataFrame) -> pd.DataFrame:
+        """Melt a wide `snapshots x components` frame into long rows.
+
+        Examples
+        --------
+        >>> c._stack_dynamic(c.dynamic['p_max_pu']).head(3)
+            snapshot             name     value  scenario  period
+        0 2015-01-01  Manchester Wind  0.930020       NaN     NaN
+        1 2015-01-01   Frankfurt Wind  0.559078       NaN     NaN
+        2 2015-01-01      Norway Wind  0.974583       NaN     NaN
```
brynpickering, line 429: A tidy dataframe would ideally have `value` always as the final column of data

pypsa/components/array.py
```diff
@@ -355,3 +416,99 @@ def _as_xarray(self, attr: str) -> xr.DataArray:
...
+        wide = wide.rename_axis(columns=["scenario", "name"] if stochastic else "name")
+
+        stacked = wide.stack(level=list(range(wide.columns.nlevels)), future_stack=True)
+        long = stacked.rename("value").reset_index()
+        if isinstance(self.snapshots, pd.MultiIndex):
+            long = long.rename(columns={"timestep": "snapshot"})
+        if "scenario" not in long.columns:
+            long["scenario"] = np.nan
+        if "period" not in long.columns:
+            long["period"] = np.nan
```
brynpickering, line 442: Rather than hardcode this, I would prefer it comes from the entries in the `dims` directory. If an entry exists there, it must also exist as a column in the dynamic entries, even if it is empty.

brynpickering (commented):
Nice!

I've had a quick look through. 

- How does it handle the equivalent of `get_switchable_as_dense`? Does it just have the entry in the long table with a `NaN` snapshot column entry? 
- How do we intend on making the decision on what can / cannot be dynamic? In a way, I can see limiting the component-level dim data to just groupers (e.g `bus`/`carrier`) (see #1474). In such a case, even the `snapshot` dim data should move to a long-form attr table, rather than being in its `dim` table. This would feel like a clearer distinction. Or, even more extreme (in case we imagine groupers _could_ change by some other dimension, e.g. `scenario`), the dimension tables are single column entries _only_, defining the dimension members and _nothing_ else.
- How will we handle piecewise data in this? Just have a `breakpoints` dimension and otherwise carry on as usual?
- Linked to the above - there should be a way to differentiate between data that _can_ be broadcast over a given dimension and one that _cannot_. When we have a breakpoints column, we need a way to say "I am a static efficiency and I should not be broadcastable over the breakpoints dimension but I can be broadcast over the snapshots dimension".

These implementation details will matter when we expand the freedom a user has for defining custom math, and storing that math alongside their data. You don't want to provide too many constraints at the data layer but rather at the math layer. This enables potentially innovative math definitions, with arbitrary new dimensions and parameters usually considered as static actually becoming dynamic.

We should also consider having `buses` and `carriers` as their own dimension entries.

lkstrp:
@brynpickering 

> How does it handle the equivalent of get_switchable_as_dense? Does it just have the entry in the long table with a NaN snapshot column entry?

Yes, this goes for all dimensions. And `manifest.json` will tell you (not yet) which dimensions apply/expand. See below. 

> How do we intend on making the decision on what can / cannot be dynamic? In a way, I can see limiting the component-level dim data to just groupers (e.g bus/carrier) (see https://github.com/PyPSA/PyPSA/issues/1474). In such a case, even the snapshot dim data should move to a long-form attr table, rather than being in its dim table. This would feel like a clearer distinction. Or, even more extreme (in case we imagine groupers could change by some other dimension, e.g. scenario), the dimension tables are single column entries only, defining the dimension members and nothing else.

PyPSA has this currently already decided. If the `type` col [[here](https://github.com/PyPSA/PyPSA/blob/master/pypsa/data/component_attrs/generators.csv)](https://github.com/PyPSA/PyPSA/blob/master/pypsa/data/component_attrs/generators.csv) is `static or series` it can be dynamic, if the attribute can vary across `scenario` dimensions is currently defined in consistency checks. I would move all of this scattered information in `manifest.json`. 

You already need to look into the schema to know on which dimensions you can expand your value if its nan. Again, see answer below. This is necessary and gives full flexibility for the long form. The only difference between dim table and long form is can this attribute be expanded to more than just a single dimension? For the dim tables you get that information without the manifest. If dim tables are single column entries only, and everything lives in long form you will always need to look into the manifest.

If you now just wanna load a simple static list of all generators (subset of `c.static`, but this is not really bound to our current data model), with the approach here you just (1) load a single dim table. With typed columns, defaults written out, every member is there and no manifest is needed. If we move all static attributes to long form, you (1) load the dim table for the members, (2) load the manifest to know which attributes make up the view, since the null rows in the files mix static attributes with the scalar values of varying ones, (3) query those attribute files, (4) pivot and cast them into wide form. 

Query engines like DuckDB do parts of that in one statement, but some things even those can't handle: 1. `value` has a different type per file once all static attributes move over, 2. nan rows need the manifest to know which dims they expand over, 3. components sitting fully at default write no rows at all, so you need the manifest for the defaults and a join back to the dim table to even see them. And all of this for the most common operation. And you not even get any of the long form storage benefits, since it's a couple of values spread over many small files.

If we think mid term, I see that we can kick out `c.dynamic` and just make it a view. `n.statistics` could then just hold lazy queries straight on parquet (I already have a prototype for this) and in the distant future I already see that the whole model definition is just parquet + your text based math language (which @FBumann already prototyped) and all the overhead of PyPSA `c.da` view + linopys complicated user interface which needs tons of checks is skipped/ removed. As you say. But what I guess you always wanna have is a quick access to `c.static` in PyPSA (most likely editable), this is what you also wanna have in any of those GUIs (PyPSA-App, Manta, whatever) and what is needed for many operations. And for all of that, you don't wanna do the 4 steps above.

So I don't see that the benefit of having an even cleaner data model where any value could expand to anything, outweighs all those cons. 

> This enables potentially innovative math definitions, with arbitrary new dimensions and parameters usually considered as static actually becoming dynamic.

Works with the stuff above. All this needs to sit in manifest.json anyway. And the costly read I describe above, you just don't do by default and only if custom math wants to expand a value which currently just sits in a dim table. Or you just do a format version update.

> How will we handle piecewise data in this? Just have a breakpoints dimension and otherwise carry on as usual?

I would have thought so. This also goes nicely with long form, since it's very sparse

> Linked to the above - there should be a way to differentiate between data that can be broadcast over a given dimension and one that cannot. When we have a breakpoints column, we need a way to say "I am a static efficiency and I should not be broadcastable over the breakpoints dimension but I can be broadcast over the snapshots dimension".

This should all go in `manifest.json`. See above, there is already

``` json
{
    "format": "pypsa-parquet",
    "format_version": 1,
    "attributes": {"name": "my_network", ...},
    "meta": {...},
    "crs": "...",
    "attribute_catalog": {
        "Generator": {
            "p_max_pu": {
                "dims": ["name", "snapshot"],
                "default": 1.0
            },
            ...
        },
        ...
    }
}
```

and `dims` tells you exactly that. Nan and in dims -> broadcast, Nan and not in dims -> don't. Currently in PyPSA we are quite strict there and it's coupled to the network attributes. But with that schema any attribute could technically be broadcasted on any subset of dimensions. It's just not correct and only descriptive currently, since I would move all the metadata we currently have in `components.csv` to the manifest, but wanted to do that in another PR and not stack in here. This also includes all the checks on the scenarios dimension. The ground truth just sits fully in the manifest then.

pypsa/components/array.py
lkstrp: This is the in memory accessor. There is no `dims` yet if the network wasn't written

PR #1858, opened 2026-08-03
feat: support pandas str dtype, require pandas 3 and pyarrow

lkstrp:
Closes #1840

Require `pandas>=3`. Pandas 3 is already supported and introduces [new string data type](https://pandas.pydata.org/docs/user_guide/migration-3-strings.html), which we currently don't use and cast everything back to numpy `object`. That also led to some confusion when using `pyarrow` on top (see https://github.com/PyPSA/PyPSA/pull/1687). This PR now allows to use the new string data type, backed by `pyarrow` which is a new dependency now. Next to keeping the old one by default. 

```
options._add_option(
    "api.legacy_string_dtype",
    None,
    "Whether to convert string data to numpy `object` dtype on import. pandas 3\n\t"
    "reads strings as `str` dtype, which PyPSA undoes by default to stay\n\t"
    "compatible with earlier versions. Set to False to keep the `str` dtype.\n\t"
    "PyPSA v2 will always keep it and drop this option.",
)
```

The new string type can be backed by PyArrow or numpy. If PyArrow is installed, it automatically is used by pandas. It brings a couple of benefits, see [PyArrow Functionality](https://pandas.pydata.org/docs/user_guide/pyarrow.html) in pandas docs. We also need it when using `.parquet` (see https://github.com/PyPSA/PyPSA/pull/1837) and brings in better interoperability with `polars` and similar engines. But currently it just matters for string dtypes. But until `v2`, nothing will change anyway.


pypsa/components/array.py
```diff
@@ -332,9 +347,9 @@ def _as_xarray(self, attr: str) -> xr.DataArray:
         if attr == "active":
             res = xr.DataArray(self.get_activity_mask())
         elif attr in self.dynamic.keys():
-            res = xr.DataArray(self._as_dynamic(attr))
+            res = xr.DataArray(_strings_to_object(self._as_dynamic(attr)))
```
FabianHofmann, line 350: ah that is unfortunate we need this... but I guess it is rare cases anyway

FabianHofmann (approved):

lkstrp:
https://github.com/PyPSA/pypsa-eur/pull/2260 is needed for PyPSA-Eur to run through

PR #1870, opened 2026-08-10
fix(optimization): inactive committable components fix

lkstrp:
Closes #1850

PR #1873, opened 2026-08-10
fix: support ramp limits under set_scenarios

lkstrp:
Ramp limits for stochastic networks currently break.

`c.da` dims under scenarios:
```
plain:      ('snapshot', 'name')
scenarios:  ('scenario', 'name', 'snapshot')
```
Fixes:
1. dim 0 was snapshot, becomes scenario -> isel./.loc select to right axis
2. `idx.isin(...)` is raw numpy, so xarray falls back to numpy broadcasting, which right aligns dims. `name` was last, with scenarios `snapshot` is. Which means the mask lands on the wrong axis. Wrapping in `DataArray(..., coords=[idx])` aligns by name label instead.

### DIFF
--- a/test/test_stochastic.py
+++ b/test/test_stochastic.py
@@ -1727,3 +1727,26 @@ def test_ramp_limit_stochastic_optimization_bug():
     for scenario in n.scenarios:
         diff = p[scenario].diff().dropna()
         assert (diff.abs() <= ramp * p_nom + tol).all()
+
+
+def test_active_inactive_assets_per_scenario():
+    """Per-scenario activeness keeps active/inactive assets mutually exclusive."""
+    n = pypsa.Network(snapshots=range(3))
+    n.add("Bus", "bus")
+    n.add("Generator", "g", bus="bus", p_nom_extendable=True)
+    n.add("Generator", "h", bus="bus", p_nom_extendable=True)
+    n.set_scenarios({"a": 0.5, "b": 0.5})
+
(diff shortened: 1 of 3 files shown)


test/test_stochastic.py
```diff
@@ -1655,6 +1655,66 @@ def test_transmission_volume_expansion_limit_constraint_stochastic():
...
+    n.set_scenarios({"a": 0.5, "b": 0.5})
+    status, condition = n.optimize(solver_name="highs")
+    assert status == "ok"
+    assert condition == "optimal"
+
+    p = n.c["Generator"].dynamic["p"].xs("g", axis=1, level="name")
+    tol = 1e-6
+    for scenario in n.scenarios:
+        diff = p[scenario].diff().dropna()
+        assert (diff.abs() <= ramp * p_nom + tol).all()
```
FabianHofmann:
```suggestion
    diff = p.diff().dropna()
    assert (diff.abs() <= ramp * p_nom + tol).all().all()
```
this should also work right?

FabianHofmann (commented):
sorry I fail to understand the bug and the fix here. I see the conversion to dataarray for the idx which does not seem to be related. is the fix beyond #1870 explicit selection of the first snapshot with `loc`/`isel`? 

(I know you dislike verbosity but a little more context would be good :) )

FabianHofmann (approved):
okay, reading again, now I understand, makes sense. dtype-geeier :egg: still needed it seems

PR #1874, opened 2026-08-10
docs: render pre-executed notebooks instead of executing on every build

lkstrp:
The readthedocs build takes ~40 mins, is triggered on every commit on all branches, and still times out from time to time. Most of those ~40 mins is the execution of notebooks.

This PR:
- Takes the burden from readthedocs. readthedocs builds still happen on every commit, but no longer execute notebooks.
- Notebooks are executed on demand and every night, and their output is stored in `docs-notebooks`. The `master` branch stays clean. Every readthedocs build simply picks up the notebook docs build from docs-notebooks.

New notebooks are now not checked with execution in the PR, but only every night on master. This is similar to the other general test matrix changes we introduced recently and I think it's a trade off which is worth it

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,58 @@
+# SPDX-FileCopyrightText: PyPSA Contributors
+#
+# SPDX-License-Identifier: MIT
+
+name: Build docs (execute notebooks)
+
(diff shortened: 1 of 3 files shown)


FabianHofmann:
nice one!

PR #1875, opened 2026-08-10
Add BODF threshold to optimize_security_constrained

### LINKED ISSUE #544: Sclopf - needs a thresholds I believe
Hi,

Sclopf (both the n.sclopf as well as the new optimize version) currently work like this:
- BODF is calculated which is reasonably fast
- Contingencies can be selected
- For every combination (contingency, element) the added flow is calculated and two new constraints for both flow directions are created

This naturally makes sclopf very expensive. As an approximation it might be good to
- filter out entries in the BODF with little added flow, which would be a first threshold
- allow the user to specify branches for which to formulate the N-1 constraints

If the user has a good idea about which branches are N-1 problematic that may be sufficient as an approximation.
At least the n.sclopf variant fails at the moment for single bus subnetworks - e.g. offshore wind - not sure where exactly.
If systems are split into async subnetworks due to the contingency... not sure what should be done there, that is probably complicated.

Maybe also worth mentioning is that the user needs to watch out for n_parallel > 1, that would currently not automatically be covered. Which is ok, but possibly should raise a warning.

FabianHofmann:
Closes #544 

Only formulate security constraints for outage/branch pairs whose absolute BODF reaches the given threshold. This makes the optimization model leaner and approximates the N-1 security efficiently if the threshold is well chosen. Additionally and unrelated to the original issue, warn when branch outages have num_parallel > 1.

## Checklist


fneum (changes requested):
Implementation looks great!

There's one element of this approximation that is not optimal. The threshold drops pairs where an outage shifts less than the threshold of the failed line's flow to the monitored line (relative threshold). But the lines could have very different capacities, e.g. a small monitored line being substantially affected by a large failed line even if the BODF is small.

If you look at Weinhold & Mieth (2020) (https://arxiv.org/abs/1910.09034 / https://ieeexplore.ieee.org/document/9094021), they propose to weight the BODF by the ratio of the failed line's capacity to the monitored line's capacity, so the screening bounds the actual worst-case flow change relative to the monitored line's own limit rather than the failed line's.

I wrote the math from their Section IV up to be added to `docs/user-guide/optimization/contingencies.md` as well. Should be quite quick to do this weighting and would be cool to add this reference as well there.

```md
!!! tip "Tip: Screening low-impact contingencies before optimisation"

The number of $N-1$ constraints can be reduced beforehand by discarding
contingency pairs whose worst-case impact is negligible, following Weinhold and
Mieth (2020)[^5]. Since the pre-outage flow $p_c$ is bounded by the branch
capacity $P_c$, the flow change on branch $b$ after the outage of branch $c$ is
bounded by:

$$|p_b^{(c)} - p_b| = |BODF_{bc}\, p_c| \leq |BODF_{bc}\, P_c|$$

Introducing a screening margin $\eta$, the contingency constraint for the pair
$(b, c)$ can be omitted whenever

$$\frac{|BODF_{bc}\, P_c|}{|P_b|} < \eta$$

i.e. when the outage of $c$ cannot shift more than a fraction $\eta$ of branch
$b$'s capacity onto it. Typical values are $\eta \in [0.01, 0.05]$, trading a
small increase in objective value for a large reduction in the number of
constraints. Because $\eta$ acts as a hard pre-filter that may alter the optimum,
it is advisable to re-check the reduced problem against a lower $\eta$ (or the full
$N-1$ set) and re-add any binding constraints.
```

As an aside, the paper contains a lot more options to speed up the N-1 contingency calculations for another time :)

PR #1829, opened 2026-07-21
Prepare optimization module for linopy v1 semantics

FabianHofmann:
Manual description to come later (even though the text below contains everything important). This PR is needed for ensuring compatibility with linopy v0.9 and the upcoming v0.10 where we introduce the v1 convention as opt-in. The main change is to make dimension alignment explicit with join parameters and interchange to use of multiindex by using xarray auxiliary coords instead.

> [!NOTE]
> This PR description was drafted with AI assistance.

Under linopy's v1 arithmetic convention ([PyPSA/linopy#717](https://github.com/PyPSA/linopy/pull/717), merged to linopy `master` and shipping in linopy `0.10`) a `pd.MultiIndex` may no longer sit on a dimension coordinate, and operands align by exact label instead of falling back to positional alignment. Multi-period models build their variables over a `MultiIndex` `snapshot`, so they stop building. This PR adapts `pypsa/optimization/` to that. Results are unchanged, and released linopy `0.9.x` stays supported.

The `snapshot` coordinate of a multi-period model:

```
# legacy — unchanged
* snapshot  (snapshot) object   MultiIndex
* period    (snapshot) int64    2030 2030 2040 2040
* timestep  (snapshot) datetime64[ns] 2013-01-01 ...

# v1
* snapshot  (snapshot) object   (2030, Timestamp('2013-01-01 00:00:00')) ...
  period    (snapshot) int64    2030 2030 2040 2040
  timestep  (snapshot) datetime64[ns] 2013-01-01 ...
```

Same labels either way, but under v1 the tuples are a plain object index and `period`/`timestep` are auxiliary coordinates. The round-trip back is `pd.MultiIndex.from_tuples`.

## Changes proposed in this Pull Request

- Expression combinations align their operands explicitly instead of relying on linopy's implicit alignment.
- The representation follows linopy's active semantics — `MultiIndex` under legacy, flat under v1 — and [`pypsa.options.optimization.model_snapshot_index`](https://github.com/PyPSA/PyPSA/blob/feat/explicit-expression-joins/pypsa/_options.py) overrides it. It is resolved once in `create_model()` and fixed for the model's lifetime.
[...]

PR #1883, opened 2026-08-17
refactor of #1829 (linopy v1 semantics)

lkstrp:
Base is #1829 and diff is shown to it

Next to some refactoring, I removed exposing `n.da` (removed it completely) and the whole window machinery. This is now internal only


FabianHofmann:
great stuff @lkstrp, just fixed one issue around the normalization. I love the trimming and found one issue and two minor things. the normalization of sns weightings did not work fully, and the cache of the flat labels seems to be not used well. will push the latter now and then we merge

Issue #1622, opened 2026-03-18
Add config attribute to store history of settings applied during network optimisation (and build?)


brynpickering:
A solved network loaded from file has missing information pertaining, in particular, to settings used at the optimisation stage. This means it isn't entirely reproducible.

I would propose a `Network.config` attribute that stores this information. It could be generalised so that it has a `optimize` sub-key which stores the most recent set of args passed to the optimisation. This would leave open the option to extend it to storing applied settings for other network methods.

This config could be:

1. applied by default when running the `optimize` step, such that a user automatically runs with the same settings. They would then call something like `n.config.clear()` to reset it. 
2. passed to the optimise function explicitly by the user, to achieve reproduction: `n.optimize(**n.config.optimize)`. 
 
This could be extended to other steps in the process of building and processing a network object, so that applied settings can be easily checked and/or can used as new baseline for defaults applied at different stages (if following the process defined in option (1) above).

PR #1898, opened 2026-08-28
feat: add selectable cycle-basis methods

WeiAi-Energy:
## Summary

Adds `Network.cycle_basis_method` with these options:

- `bfs-refined` (default)
- `bfs`
- `paton`
- `mcb`

Previously, PyPSA used NetworkX's Paton cycle basis by default. This PR changes the default to `bfs-refined`.

`bfs` constructs fundamental cycle bases from several high-degree BFS roots and selects the basis with the smallest maximum cycle length, breaking ties by total cycle length. Its use of a BFS tree to generate fundamental cycles follows the BF baseline in [Deo, Prabhu, and Krishnamoorthy (1982)](https://doi.org/10.1145/355984.355988). The root selection and basis-selection rule are implementation-specific heuristics.

`bfs-refined` applies local independent XOR exchanges to shorten the selected BFS basis.

`mcb` uses the MILP formulation in [Lee and Sun](https://arxiv.org/pdf/2510.03484) and supports SciPy/HiGHS and Gurobi.

## Motivation

The cycle basis determines the sparsity of the KVL constraints. For this unweighted objective, `mcb` provides the theoretical lower bound on total cycle-basis size. On SciGRID-DE, `bfs-refined` recovers 965 of the 1,060-edge reduction from `paton` to `mcb`: **91.0% of the theoretical reduction potential**. `mcb` takes 12.582 s with Gurobi and 23.666 s with SciPy/HiGHS; `bfs-refined` takes 0.311 s, compared with 0.304 s for `paton`. It captures most of the sparsity benefit without an additional solver dependency.

Benchmark details

SciGRID-DE has 364 independent cycles.

| Method | Basis time | Total cycle size | Maximum cycle length |
|---|---:|---:|---:|
| Paton | 0.304 s | 2,469 | 83 |
| BFS | 0.365 s | 2,020 | 26 |
| BFS-refined | 0.311 s | 1,504 | 19 |
| MCB / Gurobi | 12.582 s | 1,409 | 17 |
| MCB / SciPy-HiGHS | 23.666 s | 1,409 | 17 |

For the 24-snapshot SciGRID-DE generator-capacity-expansion model, the basis changes KVL coefficient sparsity:

| Method | KVL nonzeros | Share of model nonzeros |
|---|---:|---:|
| Paton | 59,256 | 9.02% |
| BFS | 48,480 | 7.50% |
| BFS-refined | 36,096 | 5.69% |
| MCB | 33,816 | 5.35% |

## Validation

- Added SciGRID-DE tests for `paton`, `bfs`, and `bfs-refined`.
- Added MCB unit tests, including parallel-edge cycles.
- Ran SciGRID-DE MCB integration tests with SciPy/HiGHS and Gurobi.
- Focused test suite: `15 passed, 2 skipped`.

Refs #1422, #1423


fneum:
Very cool, love a pragmatic fast approach known that has been known for more than 40 years!

I didn't have time for a deep look yet but:

a) given the `bfs-refined` yields most of the max benefit, I could imagine just not offering the `mcb` option to save some code.

b) If the `mcb` option should stay, it should use `linopy` for building the optimisation model (if possible) in a single formulation. Right now, you use different implementations based on the solver.

c) Are there already tests that check identical objective regardless of cycle basis method?

d) Did you observe any speed up in the optimisation?

WeiAi-Energy:
Thanks for your feedback @fneum! I tested the four methods on a customized 1250-bus system with 12 snapshots, solved using Gurobi with `Method=2` (barrier), `Crossover=0`, and `Threads=1`:

| | paton | bfs | bfs-refined | mcb |
|---|---|---|---|---|
| cycle basis | 0.32 s | 0.36 s | 0.34 s | 95.9 s |
| create_model | 3.39 s | 3.33 s | 3.61 s | 3.35 s |
| Gurobi runtime | 23.28 s | 25.07 s | 16.44 s | 15.00 s |
| objective | 239,731,885,372 | 239,731,775,750 | 239,731,879,317 | 239,731,885,365 |

The four methods yield same obj within 5e-7 relative, and yes, `bfs-refined` leeds to optimization speed up.

Would you prefer I keep only `bfs-refined`?

fneum:
Thanks for checking! Yes, I think given these results I wouldn't bother with `mcb` (good that you evaluated it though!). I would suggest to keep the incumbent method (standard `nx.cycle_basis()`) and `bfs-refined` (new default, since obviously advantageous). Could you also add the literature reference to https://docs.pypsa.org/latest/user-guide/optimization/power-flow/#kirchhoffs-voltage-law-kvl later? Cheers!

WeiAi-Energy:
Done as suggested!

coroa:
Very cool! Thank you for that addition

fneum (approved):
I took the liberty to add a few more tests and add setter/getter methods for the `n.cycle_basis_method` attribute in 4eee1f541bf4fd9a658eff2013924b2d6513e573.

@lkstrp, what do you think of `n.cycle_basis_method` as an attribute rather than being passed down as keyword arguments of `n.optimize()` and other functions? I thought it was quite nice to have the info stored on the network.

pypsa/network/power_flow.py
```diff
@@ -700,11 +705,21 @@ def find_tree(sub_network: SubNetwork, weight: str = "x_pu") -> None:
 def find_cycles(sub_network: SubNetwork, weight: str = "x_pu") -> None:
```
FabianHofmann, line 705: I would have expected that the method is a new argument to `find_cycles`

pypsa/networks.py
```diff
@@ -650,6 +654,35 @@ def meta(self, new: dict) -> None:
             raise TypeError(msg)
         self._meta = new
 
+    @property
```
FabianHofmann, line 657: I am not a fan of this. don't think the cycle method deserves a first class attribute on the network. if at all I would suggest to have a  read-only property on the subnetwork classes for reporting and bookkeeping purposes.

FabianHofmann: Controlling the method in the optimization should be done via an explicit argument to `.optimize`/`.create_model`. Reproducibility of the optimize function needs then to be tackled systematically via [a optimization config](https://github.com/PyPSA/PyPSA/issues/1622).

FabianHofmann (changes requested):
I don't think setting the cycle method via a property is a good choice. This would be a new logic to control network configs which I would not introduce. I suggest to explicitly set the method in relevant functions and have a bookkeeping/read-only property on the subnetwork class.

fneum:
Yes, I think passing through keyword arguments is what I would have done as well. But because @WeiAi-Energy suggested the attribute, I went through with it to see how it would look like. I am very open to revert to the alternative you suggest.

FabianHofmann:
> Yes, I think passing through keyword arguments is what I would have done as well. But because @WeiAi-Energy suggested the attribute, I went through with it to see how it would look like. I am very open to revert to the alternative you suggest.

yes, that would be nice!

fneum:
Ok, done @FabianHofmann -- you want to do a final check? https://github.com/PyPSA/PyPSA/pull/1898/commits/c34f0a7f3106c80826ac9cadc8dcfd4cba1d7c7b

WeiAi-Energy:
Oops, misclick on re-requesting review, please ignore!

FabianHofmann (approved):
I like that! thanks for tidying up @fneum ! The only thing I would like to raise is the new module network/cycle_basis.py whose content would actually fit in `graph.py` so perhaps worth considering moving the code there (also tagging our module master @lkstrp )

fneum:
> I like that! thanks for tidying up @fneum ! The only thing I would like to raise is the new module network/cycle_basis.py whose content would actually fit in `graph.py` so perhaps worth considering moving the code there (also tagging our module master @lkstrp )

Done, easy to revert if necessary.

lkstrp (approved):

PR #1911, opened 2026-09-08
Fix purchasable model, edge cases and tests (review of #1900)

FabianHofmann:
Builds on and targets the branch of #1900 (does not close an issue).

> [!NOTE]
> This description was AI-drafted from the review of #1900 and needs the author's confirmation.

Fixes from the review of #1900, targeted at that PR's branch so they land as a single reviewable diff on top of it.

**Must fix (crashes / wrong results)**

- Deleted the purchasable-committable maintenance branch that referenced an unbound `nominal_mod`, a never-created maintenance status variable, and a bilinear `A * w`. Added a consistency check rejecting `purchasable & committable & maintainable & ~modular` with a clear `ValueError`.
- Added a single `Components.active_purchasables` property (`purchasables ∩ extendables ∩ active_assets`) used in variables, constraints and the objective, so inactive and never-active multi-invest purchasables no longer raise `KeyError: 'Generator-p_nom'`.
- `check_purchasable_consistency` raises `ValueError` for `purchasable & ~extendable`; the earlier double `com_fix_i`/`com_pur_i` membership is gone.
- Dropped the modular branch that bounded a module count with a capacity-unit big-M (which capped capacity at the module size). The continuous bound `capacity <= cap_max * purchased` now applies to all purchasables; the modularity equality forces `n_mod = 0` when not bought.
- `cap_max` now uses an unscaled big-M (not scaled by `max_pu`), fixing a silent cap for low-`max_pu` assets and a crash for `StorageUnit`. `check_big_m_exceeded` extended to purchasables.
- Fixed the multi-investment scenarios case (previously an xfail) by splitting the lower-bound constraint into a plain bound and a scenario-broadcasting purchased bound, and switched the objective purchase term to `sum_dim` for correct multi-invest weighting.

**Cleanups**

- `n.statistics.capex()` now includes `purchased_opt * periodized_unit_cost` (reconciles with the objective).
- Renamed the computed property `unit_cost` to `periodized_unit_cost` (symmetry with `periodized_cost`).
- `check_cost_consistency` warns when both `unit_cost` and `unit_cost_overnight` are set.
- Removed dead code and copy-pasted comments; `available_*` skip via `startswith`; docstring and docs "current limitations" updated.

**Tests**

Converted the three strict xfails into genuine behaviour tests (non-extendable raises, `StorageUnit` solves, scenarios asserts objective). Added edge cases: inactive purchasable, mixed purchasable/non-purchasable with `p_nom_min`, small module size, unbounded `p_nom_max` with low `max_pu`, committable `Link` with negative `p_min_pu`, the maintainable-committable rejection, capex reconciliation, and a real multi-invest objective assertion. Parametrised the break-even and per-component tests.

## Checklist

- [ ] A note for the release notes `docs/release-notes.md` of the upcoming release is included. (covered by the #1900 feature note; these are fixes to unreleased code)

### DIFF
--- a/pypsa/statistics/expressions.py
+++ b/pypsa/statistics/expressions.py
@@ -753,6 +753,13 @@ class StatisticsAccessor(AbstractStatisticsAccessor):
                 attr_vals = comp.static[cost_attribute]
             piecewise_costs = comp.static.get(cost_attribute + "_piecewise_opt", 0)
             capex = capacity * (attr_vals + piecewise_costs)
+            if cost_attribute == "capital_cost" and "purchased_opt" in comp.static:
+                purchased = comp.static["purchased_opt"].fillna(0.0)
+                unit_cost = comp.periodized_unit_cost.to_series()
+                if isinstance(purchased.index, pd.MultiIndex):
+                    unit_cost = unit_cost.reorder_levels(purchased.index.names)
+                unit_cost = unit_cost.reindex(purchased.index).fillna(0.0)
+                capex = capex + purchased * unit_cost
             return capex
 
         df = self._aggregate_components(
--- a/pypsa/statistics/expressions.py
+++ b/pypsa/statistics/expressions.py
@@ -30,7 +30,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Collection, Sequence
 
     from pypsa import Network, NetworkCollection
-    from pypsa.components.components import PortsLike
+    from pypsa.components.components import Components, PortsLike
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


pypsa/statistics/expressions.py
```diff
@@ -753,6 +753,13 @@ def func(n: Network, c: str, port: str) -> pd.Series:
                 attr_vals = comp.static[cost_attribute]
             piecewise_costs = comp.static.get(cost_attribute + "_piecewise_opt", 0)
             capex = capacity * (attr_vals + piecewise_costs)
+            if cost_attribute == "capital_cost" and "purchased_opt" in comp.static:
```
brynpickering: I wasn't sure about this when implementing it but did forget to raise it in the PR description. Since `fom` can be requested separately, I wondered whether this should also extend to unit purchase costs as well.

brynpickering (approved):
Thanks for the cleanup. Happy to merge it in but have one question on the capex expression (it would need applying to the linopy model expression, too, if kept)

pypsa/statistics/expressions.py
FabianHofmann: Also not sure about this. Naturally I would say the purchase cost fall into the same bin as the capex but thinking about it, it would probably be cleaner to separate it out. should we do that?

FabianHofmann: okay, thought about it again. `periodized_cost` in pypsa includes `fom` but statistics capex does not. Therefore `fom` has its own statistics function. I would rather see the purchase cost as a natural component of the capex and expose the fine-grain control in a better way. 

how about we change the `cost_attribute` default to `None` which returns the composed capex including the purchase cost. When passing only `capital_cost` , only that will be used - same for `unit_cost`.

PR #1912, opened 2026-09-08
feat: add unit and constraint scaling (manual and auto tuned)

### LINKED ISSUE #309: Constraint matrix scaling
We would like to scale the constraint matrix which seems to speed up solving time. See [AnyMOD integration](https://github.com/leonardgoeke/AnyMOD.jl/blob/34dcb0445e4c75d8d128baadc8fdd592db1c7dfb/src/optModel/other.jl#L496-L543).

lkstrp:
Closes #309 
Closes #1773 (supersedes)
Closes #1763 (supersedes)

Adds `scaling` argument to `n.optimize` / `n.optimize.solve_model` to allow rescaling of the built linopy model before it goes to the solver. Nothing in the network data is touched. linopy maps solution, duals and objective back, so results come out in original units. We do that only on the linopy model, including custom constraints, without touching every constraint definition in PyPSA (@davide-f).

- `scaling=True` tunes one energy and one cost unit factor automatically. I did brute force this first with just the gurobi log feedback. This is now a small problem from the model's coefficient ranges which is solved and reproduces the exact same. It can also handle constraint rows (see below), which is not turned on per default since it doesn't help much/ can harm (see results). 
- A dict allows to manually pass scaling factors: `energy`, `cost` and per constraint `constraint_factors`. See docs. 
- `n.optimize.tune_scaling()` returns choice of the automatic scaler (so same as `scaling=true`). This is for inspection and to be able to be able to play around. It's exposed in the API. 
- This is an opt in feature. By default `scaling` is off. 
- User guide: [here](https://pypsa--1912.org.readthedocs.build/1912/user-guide/optimization/scaling/)

### Results
All runs can be found [here](https://pypsa--1912.org.readthedocs.build/1912/scaling-dashboard.html)

I benchmarked this with two PyPSA-Eur models on gurobi and highs with various solver settings. As discussed already a couple of times: Scaling can help a lot for crossover, a bit for simplex, and not really for barrier. For our default PyPSA Eur settings it doesnt help a lot (barrier, no cross over and high tolerance). 

I have all the results in this little dashboard. For all results below, there is a little button which selects the right runs to investigate the actual runs.

Results:
- All gurobi warnings on large bounds/ rhs are gone with scaling. 
- Unit scaling improves barrier+crossover work on h365 by a large margin. This is fully done by crossover gains, barrier iterations are unchanged, crossover runs with ~15x less iterations/ faster. 
    - But on h100 crossover already did run fine and, scaling didn't improve that at all. So it really depends on the model.
- Barrier alone doesn't really care. The barrier only configs run the same with or without scaling on both networks.
- Gurobi dual simplex gets a bit better. h365 has 1.25x less iterations, with some configs
- For HiGHS simplex scaling rescuses h365 times. Unscaled it times out, with scaling it can be solved. 
- With Gurobi ScaleFlag=0 (which turns off gurobi scaling), unscaled h100 barrier fails while the scaled version gets solved. This is currently the only failed gurobi run.
- Per constraint factors (so not just unit scaling) improve barrier but can hurt crossover. Here I am not quite sure yet. Gains for barrier are not much and crossover get's worse. In my understanding barrier doesn't care much here, since per iteration there is some equbrilation/ scaling done anyway. So our reshaped problem, doesn't help here much. Same goes for unit scaling. And crossover is doing a worse job, since unit scaling doesn't change how our problem is shaped. While per constraint scaling does, since this is add the end just a simple equribrilation and doesn't carry any opionated value (like unit scaling). And with the changed shape, it's harder for crossover the actually find the bounds.
- IPX and HiPO are nearly flat, which aligns with Barrier on gurobi. I also had problems with HiPO on h100. But I think that was some highspy stability thing. This probably needs a couple of other networks.

So results are mixed and I haven't properly understood them either. But this branch works quite fine already for further exploration.

Todos
- [ ] Remove dashboard from docs
- [ ] Blocked until linopy scaling is released


docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+## Automatic scaling
+
+``` py
+n.optimize(scaling=True)
+```
+
+`scaling=True` chooses one factor for energy quantities and one for cost
+quantities. A small integer program selects them from the coefficient ranges
+of the built model so that the spread of all magnitudes is minimised within
+the target window.
```
davide-f, line 46: What is "energy"? is it any carrier?

I'd like to see the model being solved or reference to it in the documentation

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+millions of tonnes next to emission factors below one tonne per MWh. The
+resulting constraint matrix, bounds, right-hand sides and objective have
+coefficient ranges that solvers handle poorly. Solver tolerances are absolute,
+so quantities of very different magnitude are checked against the same
+threshold, and a wide range leaves the smallest of them below the level the
+solver can distinguish from zero [^gurobi]. Solver vendors therefore recommend
+keeping matrix coefficients within six orders of magnitude, ideally in
+$[10^{-3}, 10^{6}]$, and objective coefficients, bounds and right-hand sides
+below about $10^{6}$ [^gurobi], or as close to order unity as the model allows
+[^highs].
```
davide-f, line 20: I'd suggest to summarize. I'm introducing a few revisions too.
```suggestion
Energy system models combine quantities that span many orders of magnitude.
Capacities in MW may sit next to capital costs in EUR/MW, and a CO2 budget of
millions of tonnes. The resulting constraint matrix, bounds, right-hand sides and
objective have coefficient ranges that solvers handle poorly. Ranges that are too
wide induce numerical errors as the resulting mathematical operations induce
values that may be within solver tolerances. This may result in quantities to be
mistakenly intepreted as zero [^gurobi]. Solver vendors therefore recommend
keeping matrix coefficients within six orders of magnitude, ideally in
$[10^{-3}, 10^{6}]$, and objective coefficients, bounds and right-hand sides
below about $10^{6}$ [^gurobi], or as close to order unity as the model allows
[^highs].
```

Solvers both have relative and absolute tolerances. I'd generalize slightly

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+
+Every solver applies its own internal scaling, but this operates on the matrix
+alone and cannot see the physical units behind a column or row [^highs].
+Rescaling the model in its physical units before export, for instance from MW
+to GW and from EUR to MEUR, compresses the ranges of coefficients, bounds and
+objective at once.
```
davide-f, line 26:
```suggestion
While every solver applies its own internal scaling to improve the solution, the
scaling is not calibrated to the intrinsic structure of the energy problems solved
with PyPSA. As such, PyPSA optionally supports rescaling the model in its physical
units before export; for instance from MW to GW and from EUR to MEUR,
to compress the ranges of coefficients, bounds and objective at once.
```

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+
+PyPSA can perform this rescaling on the built linopy model before it is handed
+to the solver. Factors are powers of two, so the rescaling is exact in
+floating-point arithmetic and introduces no rounding error [^highs]. Nothing in
+the network data is touched. linopy maps solution, duals and objective value
+back to the original units, so results are identical in form to an unscaled
+solve. The `scaling` argument of `n.optimize` accepts three kinds of values,
+described below. The default comes from `pypsa.options.params.optimize.scaling`
+(see [Options](../options.md)).
```
davide-f, line 35: I'd recommend to add an introduction of the scaling offered by linopy.
In particular, each row and constraint of the matrix has a designeted scaling, as well as the objective.
- The objective is multiplied the objective scaling
- Each variable has a variable scaling that multiplies all corresponding coefficients of the matrix and the objective function
- Each constraint has a constraint scaling that multiplies all coefficients and right-hand-side of the constraint

This helps the description of what automatic and manual scaling do

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+``` py
+n.optimize(scaling=True)
+```
+
+`scaling=True` chooses one factor for energy quantities and one for cost
+quantities. A small integer program selects them from the coefficient ranges
+of the built model so that the spread of all magnitudes is minimised within
+the target window.
+
+Each constraint row is divided by the factor of its variables. Matrix
```
davide-f, line 48: Can you clarify? Proposing a revision

```suggestion
Each coefficient of each constraint row is divided by the corresponding variable's scaling factor. Matrix
```

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+coefficients therefore keep their values and only bounds, right-hand sides and
```
davide-f, line 49: I don't understand how the scaling is done. From this I understand the matrix is the same and only the rhs/objective change.
Is it the intended reading?

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+the objective move. This is a pure change of units, equivalent to converting
+the inputs from MW to GW and from EUR to MEUR, which makes it safe as a
+default. Rows are not tuned individually, since in benchmarks on PyPSA-Eur
+networks per-row factors helped some solver algorithms and hurt others. They
```
davide-f, line 53: I'm partially surprised. All else being equal, I expected that row-scaling for co2 and biomass should improve the scaling by themself.

docs/user-guide/optimization/scaling.md
```diff
@@ -0,0 +1,112 @@
...
+
+[`n.optimize.tune_scaling()`][pypsa.optimization.OptimizationAccessor.tune_scaling]
+runs the tuner on the built model and returns its choice in the manual dict
+form. `scaling=True` is equivalent to passing that dict unchanged.
+
+``` py
+n.optimize.create_model()
+factors = n.optimize.tune_scaling()
+factors["constraint_factors"]["GlobalConstraint-co2_limit"] *= 4
+n.optimize(scaling=factors)
```
davide-f, line 89: How to specify custom scaling on a subset of constraints and per-variable (or group of) scaling?

For example, if I wanted to scale all variables related to AC or H2 assets, what is the recommended approach?

It would be interesting if the architecture could support with appropriate customization "bus"-level scaling, though it is of secondary priority.
The possible reason is that for large areas, some buses of the same carrier may have significantly lower demand than others and as such the rhs may be small and scaling can help. However, this is close to an edge case so it is of secondary importance if the corresponding architecture would be too heavy.

pypsa/optimization/scaling.py
```diff
@@ -0,0 +1,334 @@
...
+        msg = f"scaling key 'constraint_factors' must be a dict, got {spec.constraint_factors!r}"
+        raise TypeError(msg)
+    return ScalingFactors(
+        _positive("energy", spec.energy),
+        _positive("cost", spec.cost),
+        {k: _positive(k, v) for k, v in spec.constraint_factors.items()},
+    )
+
+
+def classify_columns(m: Model) -> dict[str, ColumnClass]:
```
davide-f, line 88: This function shall support also carriers, optionally, and energy generalizes "all" carriers.
For example, we can pick the carrier of the bus the variable relates to. This would be equivalent/similar to the logic in https://github.com/PyPSA/PyPSA/pull/1773.
I found it simpler than specifying the carrier of the component as otherwise it could become complicated

We need to ensure to provide interfaces to override this function and the classes

pypsa/optimization/scaling.py
```diff
@@ -0,0 +1,334 @@
...
+    units = _row_units(_group_ranges(m, classes), names)
+    unit_factor = {"energy": spec.energy, "cost": spec.cost, "none": 1.0}
+    rows = {
+        name: spec.constraint_factors.get(name, 1.0 / unit_factor[u])
+        for name, u in zip(names, units, strict=True)
+    }
+    return ScalingFactors(spec.energy, spec.cost, rows)
+
+
+def choose_factors(m: Model, constraint_factors: bool) -> ScalingFactors:
```
davide-f, line 249: Add documentation to the mathematical problem being solved.

pypsa/optimization/scaling.py
```diff
@@ -0,0 +1,334 @@
...
+    "cost": (1e-2, 1e6),
+    "bound": (1e-2, 1e6),
+    "rhs": (1e-2, 1e6),
+}
+WEIGHT: dict[str, float] = {"matrix": 2.0, "cost": 1.0, "bound": 1.0, "rhs": 1.0}
+VIOL_WEIGHT = 10.0
+EPS_ONE = 1e-3
+G_MAX = 40
+
+_DIMENSIONLESS_SUFFIXES = (
```
davide-f, line 34: Let's add a comment in the code section optimization to remind to update these entries when constraints are added/renamed.

pypsa/optimization/scaling.py
davide-f: For a large problem, e.g. 1e6 variables/constraints, this optimization problem can become expensive to compute. It would be great to offer two options (a) a "heuristic" such as predefined values with no optimization being executed and (b) "optimized" version with this model.
I see risks of having this scaling problem as expensive as solving the original problem. Have you tried large instances?

davide-f (commented):
Hi Lukas,

I like the proposal! Adding a few comments

PR #1916, opened 2026-09-15
Composite components from YAML definitions (proof of concept)

### LINKED ISSUE #1789: Reusable abstract "basket" of fundamental components
PyPSA has fundamental component that can be composed in various ways to represent single technologies in the world. A typical example is a battery being represented by one bus, two links & one store. In order to facilitate the creation of these "basket" of components a new data view is added to the network instance. This "basket" consists of a pre-defined set of components in a strict topological setup and additional information about respective constraints. In the example that would a "battery" basket that points to the base components and defines the capacity equality constraint of discharge and charge links.

## Naming 

The name of this feature is still to be decided on. "Basket" is preliminary placeholder. 

## API 
* the basket class runs under the network instance ie. `n.basket.<mybasket>`  
* the basket itself does not store any data but points to relevant fundamental containers (component, custom constraints)
* basket can be registered from a file ie. `n.basket.register(file or dict or BasketModel)`
* the basket exposes an add function `n.basket.<mybasket>.add()` which in the background assigns the components and the respective maths. the `add` function has a clear signature with exposed attribute from the definition file (see below) 
* there could be a `n.basket.<mybasket>.reconstruct()` basket
* a basket must define component-like data output fields to be processed by statistics module (`p`, `p0`, ...). the mapping of these output fields is defined through the expression mapping of "basket" variables to fundamental variables. definition though file  

## File based definition 
* a basket is fully defined through a "definition file" of some form (likely yaml). the definition files can be collected and distributed, made available to the community
* a definition file contains
  * the set of components
  * fixed attributes describing the topology of the components and non-exposed parameters
  * mapping of exposed attributes to fundamental attributes  
  * mapping of output basket quantities to expressions of fundamental variables

FabianHofmann:
> [!NOTE]
> AI assisted writing

Closes #1789 (proof of concept).

This PR is a prototype for composite components in PyPSA. A composite is a reusable recipe made of ordinary PyPSA components. It is described in a YAML file and can optionally carry a small piece of optimisation math written as a [math-spec](https://github.com/energy-models/math-spec) fragment. The math is layered onto the model with linopy's new `add_spec` method (PyPSA/linopy#954).

At this stage I would rather like to discuss the API of the feature. Take the following example of a battery which is composite defined in a yaml file, including a bus, a store and two links.

```python
battery = n.composites.register("battery.yaml")
n.add("battery", "bat1", bus="elec", capital_cost=10)   # or battery.add("bat1", ...)
n.add("battery", "bat2", bus="elec", availability=series)  # time series reach components and math
n.optimize()
battery.expressions["p_nom"]                        # read back per instance
n.statistics.optimal_capacity(groupby="composite")  # aggregate over members
```
In the example you see that a composite needs to be registered first. The function takes a yaml file or a pydantic model CompositeDefinition as argument. Composites can be added as single instances or batches with an add or `n.add` function which then adds all underlying base components to the network. 
Composites are optionally available in the statistics module if they come with a definition of capacity and dispatch.

# The definition

*open to see the file*
<details>

```
name: battery
description: Battery storage as a DC bus, a charger, a discharger and a store.

parameters:
  bus: null
  carrier: battery
  p_nom_extendable: true
  p_nom: 0.0
  max_hours: 4.0
  efficiency: 0.95
  capital_cost: 0.0
  marginal_cost: 0.0
  availability: 1.0

components:
  Bus:
    dc:
      carrier: $carrier
  Link:
    charger:
      bus0: $bus
      bus1: "@dc"
      efficiency: $efficiency
      p_nom: $p_nom
      p_nom_extendable: $p_nom_extendable
      capital_cost: $capital_cost
      carrier: $carrier
    discharger:
      bus0: "@dc"
      bus1: $bus
      efficiency: $efficiency
      p_nom: $p_nom
      p_nom_extendable: $p_nom_extendable
      marginal_cost: $marginal_cost
      carrier: $carrier
  Store:
    store:
      bus: "@dc"
      e_nom_extendable: $p_nom_extendable
      carrier: $carrier

math:
  variables:
    Link_p_nom:
      foreach: [name]
    Link_p:
      foreach: [snapshot, name]
    Store_e_nom:
      foreach: [name]
  expressions:
    p: sum(Link_p, by=discharger) - sum(Link_p, by=charger)
    p_nom: sum(Link_p_nom, by=discharger)
    e_nom: sum(Store_e_nom, by=store)
  constraints:
    symmetric_power:
      foreach: [battery]
      expression: sum(Link_p_nom, by=charger) == sum(Link_p_nom, by=discharger)
    energy_capacity:
      foreach: [battery]
      expression: sum(Store_e_nom, by=store) == max_hours * sum(Link_p_nom, by=discharger)
    discharge_availability:
      foreach: [snapshot, battery]
      expression: sum(Link_p, by=discharger) <= availability * sum(Link_p_nom, by=discharger)
```

</details>

First, a definition is plain data. The YAML file lists the member components, the parameters that are exposed to the user (a `null` default marks a parameter as required) and, optionally, the math fragment. The definition is validated once when it is registered.

Second, an instance is nothing more than its components. Calling `add` resolves the `$param` and `@member` references in the definition and then calls `n.add` once per member. Each member is tagged with two static columns, `composite` and `composite_type`. Exposed parameters that the math uses are stored on the primary member as `composite_param_<name>`: scalars as a static column, time series as a dynamic attribute. At `create_model` the accessor declares a parameter over `[<composite>]` or `[snapshot, <composite>]` depending on the data it finds. There is no hidden state, so `n.copy`, netCDF export and `n.remove` all work on instances without any special handling.

Third, the math only binds variables that already exist. When the model is created, the fragment is expanded into a full math-spec and added through `Model.add_spec`. It binds PyPSA's own variables, such as `Link-p_nom`, over an instance dimension, so no new PyPSA variables are introduced. This part requires linopy's v1 semantics and Python 3.12 or newer, and a guard raises a clear error otherwise.

Concretely, the PR adds the `pypsa/composites/` package with `CompositeDefinition` and the `n.composites` accessor, a `composite` grouper for statistics, a hook in `create_model`, an example definition in `examples/composites/battery.yaml`, an explanatory notebook at `docs/examples/composites-battery.ipynb`, 30 tests in `test/test_composites.py` (skipped when math-spec is not installed), and a `composites` dependency group that pins the linopy branch and math-spec.

## TODO

- [x] ~~A math fragment can only bind one component class.~~ Fragments now bind several classes. The accessor builds one shared `name` axis from the active assets of every bound class, which satisfies linopy's subset-in-order binding rule without a dimension alias. The battery example links `Store-e_nom` to `Link-p_nom` via `max_hours`. Component names must be unique across bound classes. Bound variables become reindexed views over the union axis, a memory cost on large networks that a linopy-side alias would avoid.
- [ ] Only extendable members can have their variables bound.
- [ ] Definitions live in memory. Instances persist as columns, but a network loaded from disk needs `register` again before it can be optimised.
- [ ] Time-varying exposed parameters are limited to single-period networks: linopy's wide-frame reader rejects a `MultiIndex` snapshot axis, so investment-period networks raise `NotImplementedError`.
- [ ] Required parameters (`null` default) are never declared to the math layer because the dtype comes from the default. A `dtype` key in the YAML parameters block would lift this.

One note for CI: the daily notebook executor installs only `--extra docs`, so the new notebook needs the `composites` group before it can run there.

## Checklist


FabianHofmann:
closing this as this requires a different approach and some strategic decisions. 

(background: I am seeing that the hybrid version of linopy where the linopy model in defined via python functions through `n.optimize()` and extends with layers of math on top, requires a lot of code and comes with rough edges. I don't feel comfortable introducing it)

PR #1917, opened 2026-09-15
feat: add dispatchable/committable loads

### LINKED ISSUE #1736: Active `Load` component
Price-responsive demand with operational characteristics is already expressible as a `Generator`
with negative dispatch. It works nicely but always causes confusion around users à la "should I touch change  the `sign` attribute or `p_min_pu`/`p_max_pu`?" and " why are they called generators then?". 

To have a clear separation of uses cases a new  active one-port component  `Consumer` is introduced. It mostly reuses the generator dispatch code path (with negative dispatch and cost contribution) and shares its full set of operational constraints.

coroa:
Closes #1736 .

Loads become dispatchable when any of their `p_set` (any scenario or snapshot) are NaN. Dispatchable loads are implemented as negative generators including uc and ramps, but without being extendable.

For differentiating between dispatchable and non-dispatchable loads, i added a new top-level attribute to the component code `dispatchable` which defaults to all assets, but are only the p_set is NaN somewhere ones for Loads. All the active_assets checks in constraints.py was changed to use to active_assets.intersection(dispatchable) instead of active to knead loads into the remaining code paths.

# Open questions/tasks

- [ ] max_hours for additional storability of dispatchable loads, ie. DSM?
- [x] Add documentation for active loads

## Checklist


pypsa/components/_types/loads.py
```diff
@@ -35,6 +42,70 @@ class Loads(Components):
...
+        active = self.active_assets
+        if active.empty:
+            return active
+
+        # A load is dispatchable if its static p_set is NaN, or (where a p_set
+        # time series is given) any snapshot value is NaN.
+        has_nan = self.static["p_set"].isnull()
+        dynamic = self.dynamic.get("p_set")
+        if dynamic is not None and not dynamic.empty:
+            has_nan.loc[dynamic.columns] = dynamic.isnull().any()
```
FabianHofmann:
```suggestion
        has_nan.loc[dynamic.columns] = self.dynamic["p_set"].isnull().any()
```

pypsa/components/_types/loads.py
```diff
@@ -35,6 +42,70 @@ class Loads(Components):
...
+
+    @property
+    def extendables(self) -> pd.Index:
+        """Loads are never extendable."""
+        return self.static.iloc[:0].index
+
+    @property
+    def fixed(self) -> pd.Index:
+        """Non-extendable loads with a dispatch variable (all dispatchable loads)."""
+        return self.dispatchable
```
FabianHofmann, line 83: There is a small inconsistency here with the `fixed` and `extendables` indexes compared to the other components. while for loads fixed and extendables are always a subset of active assets, this is not the case for other components. I don't think this is necessarily a problem. One could think about defining dispatchable independent of active assets and in the constraints formulation use the intersection of active assets and fixed/extendables (like done in other cases)

pypsa/components/descriptors.py
```diff
@@ -168,6 +168,23 @@ def active_assets(self) -> pd.Index:
...
+        set ``p_set``, which enters the nodal balance on the right-hand side)
+        override this to return only the dispatched subset.
+
+        Returns
+        -------
+        pd.Index
+            Single-level index of assets with a dispatch variable.
+
+        """
+        return self.active_assets
```
FabianHofmann: see above

FabianHofmann (changes requested):
I absolutely love this! this is very minimalistic and looks very stable. would have expected a larger diff. Just a small couple of requests

pypsa/components/_types/loads.py
coroa: extendables is always empty. fixed is always all of dispatchable.

coroa: ah, ok, you mean extendable for instance would be all components which are extendable not only the active ones.

FabianHofmann: yes

coroa: Ok, done

coroa:
For the failing `test_plots`, there is now a new load entry in the plots (for the active loads, even though there aren't any):

baseline
result

PR #1922, opened 2026-09-20
fix: name working replacements in deprecation warnings

Blizzeq:
Three deprecation warnings name a replacement that does not work. `sub_network.stores_i()` tells you to call `sub_network.components.stores.index.static`, which raises an `AttributeError`; the right path is `.static.index`, which is what the method's own docstring and all ten sibling accessors say. `sub_network.stores()` names `!!! deprecated.components.stores.static`, and the `aggregate_across_components` warning prints a raw `<!-- md:badge-version v1.0.0 -->` comment followed by a stray `.0`. The last two look like a docs-badge find and replace that reached into runtime strings.

The tests check the class of mistake rather than the three instances: no deprecation message carries documentation markup or an unclosed backtick, and every replacement suggested by a deprecated `SubNetwork` accessor resolves on the sub-network it points at. All three fail on master.

I ran the test files that touch `sub_network` plus the statistics and docs tests: 243 passed, 226 skipped without a solver. Ruff and ruff-format are clean on the changed files.


lkstrp (approved):
Thanks @Blizzeq !

